PR #1, opened 2023-10-06
initial commit of search feauture

zaiziclebby:
Added the following from GDS components:- 

- radio buttons 
- link (no href attribute) 
- text input


.flaskenv
```diff
@@ -6,7 +6,7 @@ FLASK_APP=govuk-frontend-flask.py
 FLASK_DEBUG=True
 FLASK_RUN_PORT=5000
 REDIS_URL=memory://
-SECRET_KEY=
+SECRET_KEY= njndfnk342
```
colinbowen: Could you remove this file, any env files should be kept for local development only. It might be good to add this to the gitignore.

app/templates/main/search.html
```diff
@@ -7,35 +7,55 @@
...
               <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">
-                <h1 class="govuk-fieldset__heading">
-                  Filter
-                </h1>
               </legend>
-              <div class="govuk-radios govuk-radios--small" data-module="govuk-radios">
+              
+
+              <div class="govuk-radios govuk-radios--small govuk-radios--inline" data-module="govuk-radios">
+                <p class="govuk-body govuk-link govuk-!-display-inline">
```
colinbowen: I think we should move this p block to after the radio buttons, there seems to be some css that we might need to add to get everything lined up. Looks good so far.

zaiziclebby: Have removed the advanced search link following call with Heather. No longer using feature.

PR #5, opened 2023-10-16
Record page

colinbowen:


anthonyhashemi (approved):

README.md
```diff
@@ -26,14 +26,21 @@ The app is provided intentionally bare, with just the essential parts that all s
...
+[Install poetry](https://python-poetry.org/docs/) and ensure you have the poetry.lock file.
+Check poetry has been installed using:
+```
+poetry --version
+```
+Then install the required dependencies using:
 ```
+poetry install
+```
+A shell can then be activated using:
```
anthonyhashemi, line 39: can also do

`poetry run **COMMAND_TO_RUN_IN_ENV`


outside the env

app/main/routes.py
```diff
@@ -93,7 +100,7 @@ def callback():
     session["session_state"] = access_token_response["session_state"]
 
     # send token to api gateway
-    api_gateway_url = "https://ljciqom6td.execute-api.eu-west-2.amazonaws.com/Dev"
+    api_gateway_url = os.getenv("API_GATEWAY_URL")
```
anthonyhashemi: do we even need this anymore. This is theurl for the authorizer api which we we are removing right?

app/tests/test_record_metadata.py
```diff
@@ -0,0 +1,17 @@
+import re
+from playwright.sync_api import Page, expect
+
+
+def test_page_title_and_header(page: Page):
```
anthonyhashemi, line 5: Nice

app/tests/test_record_metadata.py
```diff
@@ -0,0 +1,17 @@
...
+    page.goto("http://localhost:5000/record")
+    expect(page).to_have_title(
```
anthonyhashemi, line 7: could be worth installing pytest-bdd so that we can use gherking syntax and reuse building blocks.

example: https://github.com/cmoir/playwright_pytest_bdd_example/blob/master/tests/test_with_bdd.py

But not for now

app/main/routes.py
colinbowen: Good point I'll remove this now

app/main/routes.py
```diff
@@ -15,6 +17,9 @@
 
 from app.main import bp
 from app.main.forms import CookiesForm
+
+from app.data.data import consignment_response, consignment_files_response
```
colinbowen, line 21: This will be removed, was a temporary data store until opensearch integration was completed.

PR #13, opened 2023-10-25
Ayr 421/flask app search integration

anthonyhashemi:
Made a new branch for a WIP PR for the POC integration with opensearch I shared a couple days ago with the minimal changes needed for the POC search integration. The original branch had some extra changes that weren't needed for this and have also refactored some stuff here.

## Changes

- poc-search-view now makes a `multi-term` search to the configured open search instance given the free text given in the search box. We configure this in further iterations as we go.
- Have added `pytest-flask`
- Have added tests standard GET and PUT tests on the view poc-search-view (east to run in CI with pytest as we mock out the opensearch API)
- Have added unit tests for underlying business logic including aws param fetching
- Have added tests a playwright E2E test for the poc-search-view search journey (these depend on opensearch...need to think of what instance of the flask app we run this against, spin up localhost and point to the dev opensearch instance we've been using so far?, or point it directly to staging deployed application and only run these tests periodically (not on every push to branch)

## JIRA
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-421&text=search


colinbowen (approved):

app/__init__.py
```diff
@@ -39,20 +44,23 @@ def create_app(config_class=Config):
...
+
     # Set content security policy
     csp = {
         "default-src": "'self'",
         "script-src": ["'self'"],
     }
 
-    app.jinja_env.filters["null_to_dash"] = null_to_dash
+    # Disable https if app is run in testing mode
+    force_https = True if app.config["TESTING"] else False
```
anthonyhashemi: took a while to figure out what was causing redirects in the test flask client requests - turns out it is talisman. This turns it off in tests: https://github.com/GoogleCloudPlatform/flask-talisman/issues/50.

If we want to test that at all we can override on particular tests or more than likely test it in our e2e test suite.

PR #14, opened 2023-10-25
Move playwright tests to dedicated e2e tests folder

anthonyhashemi:
## Changes in this PR

- Organised the playwright tests and non e2e tests to be in separate locations so we can separate them when we run pytest. We can now have a run for non e2e tests (not requiring external apis just using the test flask dev server) and one for e2e tests against a real server.


app/main/routes.py
```diff
@@ -15,7 +15,7 @@
 from app.main import bp
 from app.main.forms import CookiesForm
 
-from app.data.data import consignment_response, consignment_files_response
+# from app.data.data import consignment_response, consignment_files_response
```
anthonyhashemi, line 18: this path doesnt exist

app/main/routes.py
```diff
@@ -153,13 +149,13 @@ def quick_access():
...
 
 
-@bp.route("/record", methods=["GET"])
-def record():
-    return render_template(
-        "record.html",
-        consignment=consignment_response,
-        consignment_files=consignment_files_response,
-    )
+# @bp.route("/record", methods=["GET"])
```
anthonyhashemi, line 152: cant import whats used in this view but I know @colinbowen is working on this view so Im sure will be fixed in an upcoming PR

app/tests/test_search.py
```diff
@@ -0,0 +1,15 @@
+from flask.testing import FlaskClient
+
+
+def test_poc_search_get(client: FlaskClient):
```
anthonyhashemi, line 4: flask pytest test - more of a unit test or integration but distinctly not an e2e test that relies upon other apis

config.py
```diff
@@ -2,15 +2,15 @@
 
 
 class Config(object):
-    CONTACT_EMAIL = os.environ.get("CONTACT_EMAIL")
-    CONTACT_PHONE = os.environ.get("CONTACT_PHONE")
-    DEPARTMENT_NAME = os.environ.get("DEPARTMENT_NAME")
-    DEPARTMENT_URL = os.environ.get("DEPARTMENT_URL")
+    CONTACT_EMAIL = os.environ.get("CONTACT_EMAIL", "")
```
anthonyhashemi, line 5: so the non e2e tests run without having to set these as we dont care about them in the current tests. maybe at some point we will override these in tests but not an issue for now

e2e_tests/test_homepage.py
```diff
@@ -3,7 +3,7 @@
 
 
 def test_has_title(page: Page):
-    page.goto("http://localhost:5000/")
+    page.goto("/")
```
anthonyhashemi, line 6: We can configure the base url in a config file, in the test files or in the cli commands

anthonyhashemi, line 6: need to decide how we will run playwright tests in ci. I think it makes sense to run them against the sandbox deployment. Im going to do a separate PR to add a deploy to sandbox workflow on merge to maim, after which we can run the playwright tests against it!

e2e_tests/test_record_metadata.py
```diff
@@ -0,0 +1,17 @@
+# import re
+# from playwright.sync_api import Page, expect
+
+
+# def test_page_title_and_header(page: Page):
```
anthonyhashemi, line 5: again once the record meatdata branch is merged in, this can be fixed

app/main/routes.py
colinbowen: This can be removed

colinbowen (approved):

PR #16, opened 2023-10-25
Isort install

colinbowen:
## Changes in this PR
- Add isort with a profile that supports black code formatting

## JIRA ticket
N/A

## Screenshots of UI changes
N/A

### After


LICENCE
```diff
@@ -0,0 +1,18 @@
+Copyright (C) 2023 Crown Copyright (The National Archives)
```
anthonyhashemi: Did this mean to be part of this PR? looks like we already have a couple of licenses, 1 of which looks identical to this one (TNA one).

Need to delete the land registry one.

I'll do that in a separatre PR. Not sure why git is allowing multiple files with same path

anthonyhashemi: see: https://github.com/nationalarchives/da-ayr-beta-webapp/pull/18

app/demos/routes.py
```diff
@@ -25,7 +25,9 @@
 @bp.route("/components/<string:component>", methods=["GET"])
 def component(component):
     try:
-        with open("govuk_components/{}/fixtures.json".format(component)) as json_file:
+        with open(
+            "govuk_components/{}/fixtures.json".format(component)
```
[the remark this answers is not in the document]
anthonyhashemi: @ben-cerium @colinbowen @MattoElGato @zaiziclebby CodeQL in practice!!! First time seeing it do anything. So that's exciting 😆

anthonyhashemi (approved):

LICENCE
colinbowen: Resolved via #18

PR #22, opened 2023-10-30
Make current poc routes protected

anthonyhashemi:
https://national-archives.atlassian.net/browse/AYR-460

When doing some testing for some other work I realised that our POC login flow was broken and we weren't protecting the poc pages that contains data/should be protected. We need to do this before we deploy to sandbox at minimum.

Note: the flash messages show fine when not redirected, but when redirected they don't show on the redirected page, only on the one after that so need to fix that.

## Changes in this PR
- Fix login flow (still using sessions for now)
- Make current poc routes protected


pgandhizaizi (approved):
reviewed , the decorator method is correct way to avoid duplication on code for each route protection , also refactoring of keycloak decode token is quite good too,
no issues found

config.py
```diff
@@ -14,4 +20,39 @@ class Config(object):
     SERVICE_URL = os.environ.get("SERVICE_URL", "")
     SESSION_COOKIE_HTTPONLY = True
     SESSION_COOKIE_SECURE = True
-    DEFAULT_AWS_PROFILE = os.getenv("DEFAULT_AWS_PROFILE")
+
+    @property
```
anthonyhashemi, line 24: we can probably further organise our configuration into a Production config which calls env vars from paramstore, another dev/local one which just reads from env vars etc. but for now this works and I didnt want to make this PR any bigger.

app/templates/main/index.html
```diff
@@ -5,6 +5,7 @@
 {% endblock %}
 
 {% block content %}
+{{ super() }}
```
anthonyhashemi, line 8: so that flashes messages are shown (as done in the base.html content block)

app/__init__.py
```diff
@@ -28,18 +30,19 @@ def null_to_dash(value):
     return value
 
 
+# use only for local development
+if os.environ.get("DEFAULT_AWS_PROFILE"):
```
anthonyhashemi, line 34: moved here so that we ensure that we have set the boto3 session to the right profile by the time we access any of the config properties which depend on aws.

PR #31, opened 2023-11-03
Ayr 390/footer component

colinbowen:
## Changes in this PR

- Footer HTML 
- CSS styling
- Footer tests written 

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-390

### Before
Screenshot 2023-11-08 at 15 40 45

### After
Screenshot 2023-11-08 at 15 40 17


e2e_tests/footer_tests.py
```diff
@@ -0,0 +1,60 @@
...
+    # Check whether the current URL matches the expected URL
+
+    expect(setup_page).to_have_url(expected_url)
+
+    # Check if the text of the link is contained within title of the page
+
+    expect(setup_page).to_have_title(expected_title)
+
+
+if __name__ == "__main__":
```
anthonyhashemi: Was this added because the test file isnt prefixed with test_? can we just rename the test file to `test_footer.py`

e2e_tests/footer_tests.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Wait for navigation and check URL is correct
+
+    setup_page.wait_for_url(expected_url, timeout=5000)
+
+    # Assertions
+
+    # Check whether the current URL matches the expected URL
+
+    expect(setup_page).to_have_url(expected_url)
```
anthonyhashemi: Could consider this assertion as redundant as we already wait for the url on line 46

e2e_tests/footer_tests.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture
+def setup_page(page: Page):
+    page.goto("/")
+    yield page
+
+
+@pytest.mark.parametrize(
+    "link_text, expected_url, expected_title", footer_links
+)
+def test_footer_links(setup_page, link_text, expected_url, expected_title):
```
anthonyhashemi: I think this e2e test is fine

but just rehighlighting the fact that when we think they are sufficient, we can also get away with tests that just test the html response of the flask views like:

```
def test_footer(client):
     assert f"<a class="govuk-footer__link" href="/cookies">Cookies policy</a>" in client.get("/").data
```

app/main/routes.py
```diff
@@ -211,6 +211,11 @@ def how_to_use():
     return render_template("how-to-use-this-service.html")
 
 
+@bp.route("/terms-of-use", methods=["GET"])
+def terms_of_use():
```
anthonyhashemi, line 215: Is this terms of use page defined yet or just a placeholder?

Just wondering if its worth testing anything in its content

e2e_tests/footer_tests.py
zaiziclebby: Ok will remove it @anthonyhashemi

app/main/routes.py
zaiziclebby: just placeholder for now @anthonyhashemi

e2e_tests/footer_tests.py
zaiziclebby: that's fine, will rename @anthonyhashemi

anthonyhashemi (approved):
looks good :)

PR #40, opened 2023-11-08
Add SCSS support

ahosgood:
Adds SCSS support for CSS files using NodeJS and [`sass`](https://www.npmjs.com/package/sass).

Requires application to build CSS before starting with `npm run build`. I need some help understanding where to add this command in the application's startup. From what I can tell, the `Dockerfile` isn't ever being used - is that correct?

Frontend developers can build and watch styles with `npm run dev`.

SCSS can be linted and fixed with `npm run lint`.


colinbowen:
Yes the Dockerfile isn't currently being used. We're planning on using Github Actions to deploy using zappa. I think the scss files could be built there once we have that setup.

colinbowen (approved):

PR #43, opened 2023-11-08
HTML/CSS files added, route created

zaiziclebby:
## Changes in this PR

-  Created sign-out.html file and built using GDS components
- Create route for sign-out page 
- Create sign-out.css file and linked to the base.html 
-  Added minimal custom css class to GDS button 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-442
### After
Screenshot 2023-11-08 at 21 25 41


app/main/routes.py
```diff
@@ -157,6 +157,11 @@ def start_page():
     return render_template("start-page.html")
 
 
+@bp.route("/sign-out", methods=["GET"])
```
anthonyhashemi: "Sign Out" seems misleading as this view doesn't sign a user out, could we change this to "signed out" as well as anywhere else we refer to sign out?

package-lock.json
```diff
@@ -14,6 +14,16 @@
         "stylelint-selector-bem-pattern": "^3.0.1"
       }
     },
+    "node_modules/@aashutoshrathi/word-wrap": {
```
anthonyhashemi: Do we need these changes in the package-lock.json? I dont see any changes to package.json?

zaiziclebby: I could just delete the lock file if that works @anthonyhashemi

zaiziclebby:
Have deleted the SCSS footer file. @colinbowen @anthonyhashemi

anthonyhashemi (approved):

package-lock.json
anthonyhashemi: just need to revert the changes in it, not delete it

app/templates/base.html
```diff
@@ -18,6 +18,7 @@
    <!-- custom css stylesheet -->
    <link rel="stylesheet" type="text/css" src="{{ url_for('static', filename='poc-search.css') }}" />
    <link rel="stylesheet" type="text/css" src="{{ url_for('static', filename='start-page.css') }}" />
+   <link rel="stylesheet" type="text/css" src="{{ url_for('static', filename='sign-out.css') }}" />
```
anthonyhashemi: needs updating to signed-out.css

PR #54, opened 2023-11-13
Start page tests

zaiziclebby:
## Changes in this PR

-  Created test_start-page.py file 
- Created test_start-page function which:-

1. Asserts the response status code is 200 (OK)
2. Checks if specific HTML content is present in the response data. 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-479
### After


zaiziclebby:
I added # noqa to ignore the 'line too long' Flake8 error on the button test in the test_start-page.py file. The test was otherwise not passing as needs to be formatted exactly as it's written in the HTML page @colinbowen

app/tests/test_start-page.py
```diff
@@ -0,0 +1,12 @@
+def test_start_page(client):
+    response = client.get("/start-page")
+
+    assert response.status_code == 200
+    assert b'<h2 class="govuk-heading-m">Before you start</h2>' in response.data
+    assert (
+        b'<h1 class="govuk-heading-l">Access your records</h1>' in response.data
+    )
+    assert (
```
anthonyhashemi, line 9:
```
    assert (
        b"""<a href="#" role="button" draggable="false"
        class="govuk-button govuk-button--start"
        data-module="govuk-button">Start now""" in response.data
    )
```
should hopefully work so that you dont need to ignore the flake8 complaint about long lines

anthonyhashemi: hmm it doesnt...

anthonyhashemi:
```
    assert (
        b'<a href="#" role="button" draggable="false" '
        b'class="govuk-button govuk-button--start" '
        b'data-module="govuk-button">Start now'
    ) in response.data

```

does work though

PR #58, opened 2023-11-14
Start page tests

zaiziclebby:
## Changes in this PR

- Start-page test file added
- Pre commit ran

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-479
### After


colinbowen:
Looks ok for now. Couple things to note for future tests:

Tests:
- Test that the links on the page (e.g., the "How to use this service" link) lead to the correct destinations and that they are functional.
- Test the start now button directs to the correct page
- PE Tests with CSS & JS enabled
- Browser compatibility tests (Chrome / firefox / mobile) etc.. will have to check how to do this with playwright as it currently uses chrome)
- Focus Management Testing: ensuring when users tab through the page the correct elements are focused in the correct order

colinbowen (approved):

PR #59, opened 2023-11-14
Ayr 445/filter search result user group

pgandhizaizi:
## Changes in this PR

added new files

poc-transferring-bodies.html -- added a new file to show list of unique transferring bodies based on the user access

modified existing files

poc-search.html -- modified logic to only show grid view of data if record count > 0
keycloak_login_required_decorator.py - added a functionality to extract keycloak user groups
search_logic.py - added a functionality to get unique list of transferring bodies from meta search
routes.py - added a new route /poc-transferring-bodies to show list of transferring bodies which the user has access to based on user group defined in keycloak
test_keycloak_login_required_decorator - added a new test cases
 
## JIRA ticket

AYR-445 - BackEnd: PoC Search results filtered by user's Keycloak groups

## Screenshots of UI changes

New html page added to show list of transferring bodies

image

### After


app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -37,30 +37,23 @@ def protected_route():
     @wraps(view_func)
     def decorated_view(*args, **kwargs):
         if current_app.config["TESTING"] and not current_app.config.get(
-            "FORCE_AUTHENTICATION_FOR_IN_TESTING"
+                "FORCE_AUTHENTICATION_FOR_IN_TESTING"
```
anthonyhashemi: did we have a change in our formatting configuration causing these indentation diff?

app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -80,3 +73,46 @@ def _check_if_user_has_access_to_ayr(keycloak_ayr_user_group, decoded_token):
...
+        realm_name=current_app.config["KEYCLOAK_REALM_NAME"],
+        client_secret_key=current_app.config["KEYCLOAK_CLIENT_SECRET"],
+    )
+
+    decoded_token = keycloak_openid.introspect(access_token)
+
+    return decoded_token
+
+
+def get_user_transferring_body_groups():
```
anthonyhashemi: Can we move `get_user_transferring_body_groups` and any share user group code to a separate `keycloak_user_groups.py` file and just leave the decorator in this keycloak_login_required_decorator.py file? 

Or at least rename this file

app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -80,3 +73,46 @@ def _check_if_user_has_access_to_ayr(keycloak_ayr_user_group, decoded_token):
...
+    function that returns list of transferring body group names based on assignment of user group in keycloak for a user
+
+    User has user group assigned in keycloak, keycloak user groups are linked to transferring bodies
+    Based on active access token of user, user groups are extracted and list of user groups details retrieved
+
+    Args:
+        N/A
+
+    Returns:
+        function: return list of transferring bodies
```
anthonyhashemi:
```suggestion
        List[str]: return list of transferring bodies
```

app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -80,3 +73,46 @@ def _check_if_user_has_access_to_ayr(keycloak_ayr_user_group, decoded_token):
...
+
+
+def get_user_transferring_body_groups():
+    """
+    function that returns list of transferring body group names based on assignment of user group in keycloak for a user
+
+    User has user group assigned in keycloak, keycloak user groups are linked to transferring bodies
+    Based on active access token of user, user groups are extracted and list of user groups details retrieved
+
+    Args:
```
anthonyhashemi: I think we can omit this Args section in the description if there are no arguments to the function

app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -80,3 +73,46 @@ def _check_if_user_has_access_to_ayr(keycloak_ayr_user_group, decoded_token):
...
+        N/A
+
+    Returns:
+        function: return list of transferring bodies
+    """
+
+    access_token = session.get("access_token")
+    user_group_list = []
+    if access_token:
```
anthonyhashemi: Could we replace the check that uses `found_index = group.find("transferring_body")` 
 with `group.starts_with("transferring_body/") to ensure we are looking for `transferring_body/` at the start of the group?


Either way, it could be worth inverting the condition blocks here so that we aren't nesting so much if you agree that its easier to comprehend - might be a personal choice though

e.g.

```
   if not access_token:
        return []
   decoded_token = _decode_token(access_token)
   if not decoded_token["active"]:
       return []

   groups = decoded_token["groups"]
   
   users_transferring_bodies = []   
   for group in groups:
        if not group.starts_with("transferring_body/"):
           continue
        split_group_string = group.split("transferring_body/")
        if len(split_group_string) == 1:
           continue
        transferring_body = (split_group_string[2]).strip()
        users_transferring_bodies.append(transferring_body)
    return users_transferring_bodies
```

app/main/routes.py
```diff
@@ -103,19 +104,47 @@ def poc_search():
...
         )
-        results = open_search_response["hits"]["hits"]
-        session["search_results"] = results
-
-    num_records_found = len(results)
+        search_results = open_search_response["hits"]["hits"]
+        session["search_results"] = search_results
 
+    num_records_found = len(search_results)
+    print(num_records_found)
```
anthonyhashemi: can we remove this print?

app/main/search/search_logic.py
```diff
@@ -46,3 +46,31 @@ def generate_open_search_client_and_make_poc_search(query: str, index) -> Any:
...
+        "size": 0,
+        "aggs": {
+            "unique_values": {
+                "terms": {
+                    "field": field_name,
+                }
+            }
+        }
+    }
+    # print(open_search_query)
```
anthonyhashemi: can we remove this?

app/main/routes.py
```diff
@@ -103,19 +104,47 @@ def poc_search():
...
     )
 
 
+@bp.route("/poc-transferring-bodies", methods=["POST", "GET"])
+@access_token_login_required
+def poc_transferring_bodies():
+    form = SearchForm()
+    field_name = "Source_Organization.keyword"
+
+    # check if user has transferring body group access or not
```
anthonyhashemi: Could we do without the comments in this function? This variable name and function call here seems descriptive enough to not need a descriptive comment which would need maintaining. if anything changes.

and `# get unique transferring bodies` can be replaced with a private function if we want to make that bit of the logic clear:

```
def get_unique_transferring_bodies():
    user_transferring_body_groups = get_user_transferring_body_groups()
    unique_response = search_logic.generate_open_search_client_and_unique_values(
        current_app.config["AWS_OPEN_SEARCH_INDEX"], field_name)
    unique_values = unique_response["aggregations"]["unique_values"]["buckets"]
    unique_transferring_bodies = []
    for value in unique_values:
        group_name = value["key"]
        if len(user_transferring_body_groups) > 0:
            for user_group in user_transferring_body_groups:
                if user_group.strip().replace(' ', '').lower() == group_name.strip().replace(' ', '').lower():
                    unique_transferring_bodies.append(group_name)
    return unique_transferring_bodies
```


like the other comment, could also do some condition inversion if we think it makes the for loop easier to read

app/main/authorize/keycloak_login_required_decorator.py
anthonyhashemi: so it seems the pre-commit github action check is failing due to formatting issues.

If you haven't already, could you  run `pre-commit install` in your poetry environment, to ensure that every time you commit it runs the pre-commit hooks.

To fix all the files see which ones are causing issues in the gh action, and then manually run `pre-commit run --files LIST_OF_FILES_TO_RUN_ON`, and then commit them (the checks should run against all staged files but pass now that the first run fixed them)

app/main/authorize/keycloak_login_required_decorator.py
pgandhizaizi: reset file to original state : keycloak_login_required_decorator.py

app/main/authorize/keycloak_login_required_decorator.py
pgandhizaizi: removed Args from the comment details

app/main/authorize/keycloak_login_required_decorator.py
pgandhizaizi: created a new file : kecloak_manager.py and amended logic for get_user_transferring_body_groups() function

app/main/routes.py
pgandhizaizi: removed print comment

app/main/search/search_logic.py
pgandhizaizi: removed print comment

app/main/routes.py
pgandhizaizi: as discussed over the call, the function logic is temporary and will move once full browse view functionality done

pgandhizaizi:
no longer a valid branch and changes, latest changes are reflected in another branch AYR-445-user-group-filter-transferring-body-view

PR #64, opened 2023-11-16
Update SCSS files with custom classes

zaiziclebby:
## Changes in this PR

- Deleted CSS files for POC search, start-page and signed in page. 
- Amended classes in above SCSS files to be unique to those pages

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-475
### After

### DIFF
--- a/app/static/src/scss/poc-search.scss
+++ b/app/static/src/scss/poc-search.scss
@@ -1,16 +1,17 @@
 /* CSS for Proof of concept Search Page */
 
-.govuk-form-group {
+.govuk-form-group--poc-search {
+
   width: 53rem;
 }
 
-.govuk-button {
+.govuk-button--poc-search {
   margin-top: -4.3rem;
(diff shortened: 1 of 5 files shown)


app/static/src/scss/poc-search.scss
```diff
@@ -1,16 +1,17 @@
...
 
-.govuk-form-group {
+.govuk-form-group--poc-search {
+
   width: 53rem;
 }
 
-.govuk-button {
+.govuk-button--poc-search {
   margin-top: -4.3rem;
```
ahosgood, line 9: Just as a warning - negative margins _can_ cause layout issues. I wonder if there is an easier way of positioning these elements?

ahosgood: (We agreed to have a separate discussion about this)

ahosgood (approved):

anthonyhashemi (approved):

PR #65, opened 2023-11-16
Add sign out button

anthonyhashemi:
## Changes in this PR

- Main changes: 
     - Added logout route which logs the user out from keycloak and clears the user session when an authenticated user access the logout route
     - Added "Sign out" button to banner on views that require authentication (using context processor which hooks into flask's g object https://flask.palletsprojects.com/en/2.3.x/appcontext/#storing-data), which redirects to the logout view
     - Removed temporary flask flashed message about being logged in, now that we have the sign out button.

- Bonus changes: 
     - Improved testing of routes protected by the login required decorator
     - Removed border from banner as per new designs. (havent done the text changes but if its just a text change, Im happy to if we are certain without the new text in the designs @Terry-Price )

## JIRA ticket

https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?assignee=63b4326678aabbefa9d4400d&selectedIssue=AYR-478&text=ign%20out

### Before
Screenshot 2023-11-16 at 15 29 34

### After

Screenshot 2023-11-16 at 15 21 12

Screenshot 2023-11-16 at 15 20 55
Screenshot 2023-11-16 at 15 21 04


app/main/authorize/keycloak_login_required_decorator.py
```diff
@@ -36,39 +36,48 @@ def protected_route():
...
+            return view_func(*args, **kwargs)
+        finally:
+            g.access_token_login_required = (
+                False  # Clear attribute after view execution
             )
-            return redirect(url_for("main.index"))
 
-        flash("TNA User is logged in and has access to AYR.")
-        return view_func(*args, **kwargs)
+    decorated_view.access_token_login_required = True
```
anthonyhashemi: used to mark whether the view is decorated or not so can be introspected in tests

app/main/routes.py
```diff
@@ -28,6 +28,21 @@ def index():
     return render_template("index.html")
 
 
+@bp.route("/logout", methods=["GET"])
+@access_token_login_required
+def logout():
```
anthonyhashemi: called this view logout even though the naming of the view we are redirected to after is signed_out, since our login view is called login but I guess I should probably change both logout and login to be sign_out and sign_in?

app/static/src/scss/base.scss
```diff
@@ -0,0 +1,13 @@
+.banner__container {
```
anthonyhashemi, line 1: are these names following the BEM naming convention?

app/static/src/scss/base.scss
```diff
@@ -0,0 +1,13 @@
...
+    display: flex;
+    justify-content: space-between;
+    border-bottom: none;
+}
+
+.sign-out {
+    font-size: 1.25rem;
```
anthonyhashemi: should we be using rem or gds font sizes? similar discussion here: https://designnotes.blog.gov.uk/2018/02/19/developing-new-typography-and-spacing-for-gov-uk-frontend/#comment-66108

app/main/routes.py
colinbowen: Sounds good, we should go with sign in & sign out. The phrasing is slightly more formal and welcoming than "login."

app/static/src/scss/base.scss
colinbowen: I believe this is ok. BEM allows for a block__element with no modifier

app/main/routes.py
pgandhizaizi: I agree with sign in and sign out are better phrases

app/static/src/scss/base.scss
ahosgood: In this instance, I feel we should be using the same units that GDS does as we are extending their styles and want to avoid issues with conflicting units and sizes when users resize text in their browsers.

app/static/src/scss/base.scss
ahosgood: This format is fine 👍🏼

pgandhizaizi (approved):
Looks good to me, approved

PR #70, opened 2023-11-22
Browse journey

colinbowen:
## Changes in this PR
- Initial Browse html, css
- Browse Playwright tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-397

## Screenshots of UI changes
Screenshot 2023-11-22 at 13 13 01

### After


e2e_tests/test_browse.py
```diff
@@ -0,0 +1,92 @@
+import time
+
+from playwright.sync_api import Page
+
+
+def test_has_title(page: Page):
+    time.sleep(1)
```
anthonyhashemi: can get away without sleep in the tests themselves. I think we can just run the playwright with `-slowmo=200` for example to pause 200ms between calls

e2e_tests/test_browse.py
```diff
@@ -0,0 +1,92 @@
...
+        ".filters-form__transferring-body--select", "Arts Council England"
+    )
+    time.sleep(1)
+
+    records = page.query_selector_all(".govuk-summary-list__value a")
+    for record in records:
+        assert record
+        # assert record.text() == "Arts Council England"
+
+
```
anthonyhashemi, line 45: I guess this PR isnt ready to review yet until Pavans work is merged in?

colinbowen: Yes, this will be updated with the ui revisions & test updates to confirm actual records that should be visible to our test users.

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+{% block content %}
+<div class="govuk-grid-row browse__page">
+  {{ super() }}
+
+  <div class="search__container govuk-grid-column-full">
+    <div class="search__container__content">
+      <p class="govuk-body search__heading">Search for digital records</p>
+      <form method="POST">
+        {{ form.csrf_token }}
+        <div class="govuk-form-group search__form">
```
zaiziclebby: I believe where we use the GDS components we need to keep those intact and build on it by repeating the class and using the element/modifier at the end of the GDS class. For example  "govuk-form-group govuk-form-group__search--form"

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+          <label for="searchInput"></label>
+          <input class="govuk-input govuk-!-width-three-quarters" id="searchInput" name="query" type="text">
+          <button class="govuk-button search__button" data-module="govuk-button" type="submit">Search</button>
```
zaiziclebby: Here you could do "govuk-button  govuk-button__search--button" that ensures you use the GDS class and keep the element/modifer on the custom class.

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+        </div>
+        <p class="govuk-body-s">Search using a record metadata term, for example – transferring body, series,
+          consignment
+          ref etc.</p>
+      </form>
+    </div>
+  </div>
+  {% if results %}
+  <h1 class="govuk-heading-l browse_records-found_text">Records found {{ num_records_found }}</h1>
```
zaiziclebby: Here you could do "govuk-heading-l  govuk-heading-l__browse_records-found_text" you should be able to use underscores, I usually just use a dash but it shouldn't make a difference.

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+            <option value="updated">Arts Council England</option>
+            <option value="views">Food Standards Agency</option>
+            <option value="comments">Foreign Office</option>
+          </select>
+        </div>
+      </div>
+
+      <div class="browse-series-container">
+        <p class="govuk-body browse-body">Series</p>
+        <div class="govuk-form-group browse-filters__form-group filters-form__series-group">
```
zaiziclebby: This class is quite long do you need all the added classes? Could you do something simpler like "govuk-form-group govuk-form-group__browse-filters"

app/static/src/scss/browse.scss
```diff
@@ -0,0 +1,248 @@
...
+
+.browse-sort-container {
+  margin-top: 3rem;
+}
+
+.sort-container {
+  align-items: center;
+}
+
+.sort-container__form {
```
zaiziclebby: Could nest the sort-container classes like this:- 

.sort-container {

&__form {
  margin-bottom: 0.5rem;
}

&__select {
  margin-right: 0.25rem;
}
}

app/static/src/scss/browse.scss
```diff
@@ -0,0 +1,248 @@
...
+  font-size: 1.5rem;
+}
+
+.browse-filter-container {
+  background-color: #f3f2f1;
+  padding: 0.5rem;
+  min-width: 100%;
+  margin-bottom: 0.5rem;
+}
+
```
zaiziclebby: Could nest the browse-filter class like this- 


.browse-filter {

&__header {
  display: flex;
}

&__icon {
  justify-content: space-between;
  width: 2rem;
  margin-left: auto;
}
}

app/static/src/scss/browse.scss
```diff
@@ -0,0 +1,248 @@
...
+.browse-date-container {
+  background-color: #f3f2f1;
+  padding: 0.5rem;
+  min-width: 100%;
+}
+
```
zaiziclebby: You could nest the filters-form like this:- 

.filters-form {

&__transferring-body-group {
  display: flex !important;
  align-items: baseline;
}

&__series-group {
  display: flex;
}

&__consignment-ref-group {
  display: flex;
}

&__date-from-group {
  display: flex;
  justify-content: space-between;
}

&__date-to-group {
  display: flex;
  justify-content: space-between;
}

&__transferring-body--select {
  width: 100%;
  min-width: auto;
  font-size: 1rem;
}

&__series--input {
  flex-grow: 1;
  width: 100%;
}

&__consignment-ref--input {
  flex-grow: 1;
  width: 100%;
}

&__buttons {
  display: flex;
  justify-content: space-between;
  align-items: center;
  width: auto;
  margin-top: 1rem;
}

&__apply--button {
  align-self: flex-start;
  margin-left: 1rem;
  font-size: 1rem;
}

&__clear--button {
  margin-right: 1rem;
  font-size: 1rem;
}

&__transferring-body--button {
  display: block;
}

&__series--button {
  align-self: flex-end;
  justify-content: flex-end !important;
  margin-right: 0.5rem !important;
  margin-left: auto;
}

&__consignment-ref--button {
  align-self: flex-end;
  justify-content: flex-end !important;
  margin-right: 0.5rem !important;
}

&__date--button {
  align-self: flex-end;
  margin-right: 0.5rem !important;
}
}

app/static/src/scss/browse.scss
```diff
@@ -0,0 +1,248 @@
...
+  font-size: 1.5rem !important;
+}
+
+
+.browse-pagination--centered {
+  justify-content: center;
+}
+
+
+.browse-grid--two-thirds {
```
zaiziclebby: You could nest the browse-grid block like this:-

.browse-grid {

      &--two-thirds {
      padding: 0;
     }

       &__list {
      font-size: 1rem !important;
       }

        &__key {
         width: auto;
        }

          &__key__transferring-body {
          width: 30%;
          }
}

zaiziclebby (changes requested):
.

app/static/src/scss/browse.scss
colinbowen: Updated to recommended class

app/static/src/scss/browse.scss
colinbowen: Also updated

app/static/src/scss/browse.scss
colinbowen: Updated to use new browse-grid class

app/static/src/css/browse.css
```diff
@@ -0,0 +1,236 @@
+.govuk-form-group__search-form {
```
ahosgood: Is this file meant to be here? It looks like the compiled version of `app/static/src/scss/browse.scss`.

I would have though it'd been ignored by https://github.com/nationalarchives/da-ayr-beta-webapp/blob/main/.gitignore#L9

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+      <div class="browse-filter-container">
+        <div class="browse-filter__header">
+          <p class="govuk-body-l browse__body filters-form__title">Filter within browse</p>
+          <img src="{{ url_for('static', filename='image/filter-icon.svg') }}" class="browse-filter__icon"
+            alt="filter-icon">
+        </div>
+        <p class="govuk-body browse__body">Transferring body</p>
+        <div class="govuk-form-group filters-form__group filters-form__transferring-body-group">
+          <select class="govuk-select govuk-select__filters-form-transferring-body-select" id="sort" name="sort">
+            <option value="published" selected>Choose one...</option>
```
ahosgood: Would this be better as "All" or "Any" as a default option?

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+
+      <div class="filters-form__series__container">
+        <p class="govuk-body browse__body">Series</p>
+        <div class="govuk-form-group filters-form__group filters-form__series-group">
+          <input class="govuk-input filters-form__series--input" id="width-5" name="width10" type="text">
+        </div>
+      </div>
+
+      <div class="filters-form__consignment-ref__container">
+        <p class="govuk-body browse__body">Consignment ref</p>
```
ahosgood:
```suggestion
        <p class="govuk-body browse__body">Consignment <abbr title="reference">ref</abbr></p>
```

ahosgood: (if "ref" is short for "reference")

app/static/src/css/browse.css
colinbowen: I've removed this now, all css are in the gitignore 👍

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+        <div class="govuk-form-group filters-form__group filters-form__consignment-ref-group">
+          <input class="govuk-input filters-form__consignment-ref--input" id="width-5" name="width10" type="text">
+        </div>
+      </div>
+
+      <div class="filters-form__date__container">
+        <div class="govuk-form-group date-form filters-form__group filters-form__date-from-group">
+          <p class="govuk-body date-form browse__body filters-form__body-date">Date from</p>
+          <fieldset class="govuk-fieldset" role="group" aria-describedby="date-from">
```
ahosgood, line 91: `aria-describedby="date-from"` means there should be an element with the ID `date-from` which I don't think we have.

GOV.UK have added a hint with that ID: https://design-system.service.gov.uk/components/date-input/. We can either add a hint or we can remove the `aria-describedby`.

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+            <div class="govuk-date-input" id="date-to">
+              <div class="govuk-date-input__item">
+                <div class="govuk-form-group">
+                  <input class="govuk-input govuk-date-input__input govuk-input--width-2" id="date-from-day"
+                    name="date-from-day" type="text" inputmode="numeric">
+                  <label class="govuk-label govuk-date-input__label filters-form__date__label" for="date-from-day">
+                    (DD)
```
ahosgood, line 98: Testing might give us a better indication of whether `DD`, `MM` and `YYYY` are easier to understand than `Day`, `Month` and `Year`.

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+            </div>
+          </fieldset>
+        </div>
+
+
+        <div class="filters-form__buttons">
+          <button type="button" class="govuk-button govuk-button__filters-form-apply-button" data-module="govuk-button">
+            Apply filters
+          </button>
+          <p class="govuk-body-m filters-form__clear--button"><a href="#">Clear all filters</a></p>
```
ahosgood: Is `<a href="#">` right? Would a `<button type="reset">` work? Just thinking that this is a link that doesn't take you anywhere and would jump you to the top of the page.

app/static/src/scss/browse.scss
```diff
@@ -0,0 +1,262 @@
...
+  &__icon {
+    justify-content: space-between;
+    width: 2rem;
+    margin-left: auto;
+  }
+}
+
+.filters-form {
+  float: right;
+  font-size: 1rem !important;
```
ahosgood: Are the `!important` declarations necessary?

app/templates/base.html
```diff
@@ -21,97 +21,101 @@
...
-                'id': "rejected-hide"
-              },
-              'text': "Hide this message"
-            }
-          ]
-        }
-      ]
-    }) }}
-  {% endif %}
+{% if "cookies_policy" not in request.cookies %}
```
ahosgood: This file has become arguably less readable now the indentation has been removed. Was this an autoformatter?

app/templates/main/browse.html
```diff
@@ -0,0 +1,270 @@
...
+
+      <div class="browse-filter-container">
+        <div class="browse-filter__header">
+          <p class="govuk-body-l browse__body filters-form__title">Filter within browse</p>
+          <img src="{{ url_for('static', filename='image/filter-icon.svg') }}" class="browse-filter__icon"
+            alt="filter-icon">
+        </div>
+        <p class="govuk-body browse__body">Transferring body</p>
+        <div class="govuk-form-group filters-form__group filters-form__transferring-body-group">
+          <select class="govuk-select govuk-select__filters-form-transferring-body-select" id="sort" name="sort">
```
ahosgood, line 65: Do these filters do anything without JavaScript enabled? They don't appear to be in a `<form>` element.

app/templates/main/browse.html
colinbowen: Added abbreviation element for consignment reference.

app/templates/main/browse.html
colinbowen: We've still to implement this select element on the backend so it does have some default data in there

app/templates/main/browse.html
colinbowen: Will be sure to let Terry know it might be good to get feedback on this

app/templates/base.html
colinbowen: Reverted formatting, looks like VS Code is causing an issue here. Should be cleaner to view now.

app/templates/main/browse.html
colinbowen: The filters will be calculated on the backend, we don't have any JS features yet so will be sure to put this inside of a form.

app/static/src/scss/browse.scss
colinbowen: Removed all usage of important 👍

zaiziclebby (dismissed):

app/templates/main/browse.html
colinbowen: Added an id for the date from field, we're currently using the DD MM YYYY as hints, I'll check with @Terry-Price to see if hints or labels are preferred. GDS also has some error messaging around this which should be implemented.

The filter functionality is to be implemented so this will be a placeholder for now.

app/templates/main/browse.html
colinbowen: Noted in ticket: https://national-archives.atlassian.net/browse/AYR-560

app/templates/main/browse.html
colinbowen: Changed to a button, this will have further work during the filters development

pgandhizaizi:
1 ) in routes.py , browse() function, some code does not make sense and will confuse with logic
e.g. search_type = "browse' set on line 105
then again when set on line 121 for search - I would put it as search_type="search" 
and line 124 checks search_type="browse" and query:  it should just check search_type

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
 def accessibility():
     return render_template("accessibility.html")
 
 
+@bp.route("/browse", methods=["POST", "GET"])
+@access_token_sign_in_required
+def browse():
+    transferring_body_id = request.args.get("transferring_body", None)
+    series_id = request.args.get("series", None)
+    consignment_id = request.args.get("consignment_id", None)
```
anthonyhashemi, line 100: can we make this consistent with the other 2 args either drop _id on this one or add _id to those

e2e_tests/test_browse.py
```diff
@@ -0,0 +1,69 @@
...
+    page.select_option(
+        ".filters-form__transferring-body--select", "Arts Council England"
+    )
+
+    records = page.query_selector_all(".govuk-summary-list__value a")
+    for record in records:
+        assert record
+
+
+def test_pagination(page: Page):
```
anthonyhashemi: you can use authenticated_page so you dont have to write the login code each time.

e2e_tests/test_browse.py
```diff
@@ -0,0 +1,69 @@
...
+def test_sort_dropdown(page: Page):
+    page.goto("/browse")
+    page.get_by_label("Email address").click()
+    page.get_by_label("Email address").fill(username)
+    page.get_by_label("Password").click()
+    page.get_by_label("Password").fill(password)
+    page.get_by_role("button", name="Sign in").click()
+    page.get_by_label("", exact=True).click()
+    page.get_by_role("button", name="Search").click()
+    page.select_option("#sort", "body-a")
```
anthonyhashemi: do we want to make any assertions on this test? Otherwise were just showing that these can be selected but nothing about the effect of selecting

e2e_tests/test_browse.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_page_loading(page: Page):
+    page.goto("/browse")
+    page.get_by_label("Email address").click()
+    page.get_by_label("Email address").fill(username)
+    page.get_by_label("Password").click()
+    page.get_by_label("Password").fill(password)
+    page.get_by_role("button", name="Sign in").click()
+    page.wait_for_selector(".govuk-heading-l")
```
anthonyhashemi: what are we testing here?

e2e_tests/test_browse.py
```diff
@@ -0,0 +1,69 @@
...
+    page.get_by_label("Password").click()
+    page.get_by_label("Password").fill(password)
+    page.get_by_role("button", name="Sign in").click()
+    page.select_option(
+        ".filters-form__transferring-body--select", "Arts Council England"
+    )
+
+    records = page.query_selector_all(".govuk-summary-list__value a")
+    for record in records:
+        assert record
```
anthonyhashemi, line 43: what is the purpose of this assertion?

app/tests/test_browse.py
```diff
@@ -0,0 +1,88 @@
...
+
+    response = client.get(f"/browse?series={series_id}")
+
+    assert response.status_code == 200
+    assert b"Search for digital records" in response.data
+    assert b"Records found 1" in response.data
+    assert b"test body1" in response.data
+
+
+def test_browse_consignment(client: FlaskClient):
```
anthonyhashemi, line 55: the consignment page is not currently working properly, but I guess its fine, we know we need to update the page and then we will have to update this test when we do that

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
+
+    # Apply filters to browse_data or call fuzzy_search based on search_type
+    if search_type == "browse" and query:
+        search_results = fuzzy_search(query)
+    elif filters:
+        search_results = browse_data(
+            **{k: v for d in filters for k, v in d.items()}
+        )
+    else:
+        search_results = browse_data()
```
anthonyhashemi: browse_results
```suggestion
        browse_results = browse_data()
```

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
+        search_type = "consignment"
+    if query:
+        filters.append({"query": query})
+        search_type = "browse"
+
+    # Apply filters to browse_data or call fuzzy_search based on search_type
+    if search_type == "browse" and query:
+        search_results = fuzzy_search(query)
+    elif filters:
+        search_results = browse_data(
```
anthonyhashemi, line 127:
```suggestion
        browse_results = browse_data(
```

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
+    if consignment_id:
+        filters.append({"consignment_id": consignment_id})
+        search_type = "consignment"
+    if query:
+        filters.append({"query": query})
+        search_type = "browse"
+
+    # Apply filters to browse_data or call fuzzy_search based on search_type
+    if search_type == "browse" and query:
+        search_results = fuzzy_search(query)
```
anthonyhashemi, line 125:
```suggestion
        browse_results = fuzzy_search(query)
```

app/main/routes.py
```suggestion
        browse_results = browse_data(
```

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
+        search_type = "series"
+    if consignment_id:
+        filters.append({"consignment_id": consignment_id})
+        search_type = "consignment"
+    if query:
+        filters.append({"query": query})
+        search_type = "browse"
+
+    # Apply filters to browse_data or call fuzzy_search based on search_type
+    if search_type == "browse" and query:
```
anthonyhashemi, line 124: like Pavan mentioned, maybe renaming this to 
`browse_type` will be clearer

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
...
+    form = SearchForm()
+    search_results = []
+    search_type = "browse"
+
+    # Create a list to accumulate filter conditions
+    filters = []
+
+    if transferring_body_id:
+        filters.append({"transferring_body_id": transferring_body_id})
+        search_type = "transferring_body"
```
anthonyhashemi: like Pavan mentioned, maybe renaming this to 
`browse_type` will be clearer

app/main/routes.py
anthonyhashemi: like we discussed with Terry, doing a search on the browse view should lead to a search results page which we havent yet created but should be something like the poc-search view. 
Maybe we can just point to that for now in this browse template, that way we can remove the need for confusing this view with search logic?

app/main/routes.py
```diff
@@ -84,14 +84,64 @@ def callback():
     session["token_scope"] = access_token_response["scope"]
     session["session_state"] = access_token_response["session_state"]
 
-    return redirect(url_for("main.poc_search"))
+    return redirect(url_for("main.browse"))
```
anthonyhashemi, line 87: I configured it so that only our non e2e tests contribute to our code coverage as our e2e tests should be supplementary to our normal tests

but its not a big deal that this line is not covered in the unit tests.

app/templates/main/browse.html
```diff
@@ -0,0 +1,433 @@
...
+          </dd>
+        </div>
+        {% endfor %}
+      </dl>
+      {% endif %}
+      {% endif %}
+      {% endif %}
+
+
+      {% if search_type == "series" %}
```
anthonyhashemi: not going to suggest you do it now, but I think it could be worth us splitting this out into sub templates since this file is huge.

I may do that when doing the consignment browse view

anthonyhashemi (approved):
Looks good, like we discussed we still need to iron out some bits, in particular sorting and consignment browse, but will be sorted in separate tickets.

e2e_tests/test_browse.py
colinbowen: Once we have sort functionality, this test will be changed also.

e2e_tests/test_browse.py
colinbowen: Updated. 👍

app/main/routes.py
colinbowen: Added _id to keep consistent with the rest of the code.

e2e_tests/test_browse.py
colinbowen: Hoping to test that we have the expected records/ transferring bodies showing for each page. Once we have completed the filter work this test should be updated.

app/templates/main/browse.html
colinbowen: Yeah it is getting bigger with a bunch of conditional logic for each browse type. Will make a note of this.

app/main/routes.py
colinbowen: Good shout, removed integrating the search logic into browse for now.

app/main/routes.py
colinbowen: Updated, should be much clearer now removing search logic.

e2e_tests/test_browse.py
anthonyhashemi: ok, worth noting that in the ticket created for filtering

e2e_tests/test_browse.py
anthonyhashemi: worth noting in sort ticket

app/main/routes.py
```diff
@@ -84,14 +84,57 @@ def callback():
...
+        browse_type = "transferring_body"
+    if series_id:
+        filters.append({"series_id": series_id})
+        browse_type = "series"
+    if consignment_id:
+        filters.append({"consignment_id": consignment_id})
+        browse_type = "consignment"
+
+    # Check browse type
+    if browse_type == "browse":
```
anthonyhashemi: I think this filter can be simplified from 
`browse_data(**{k: v for d in filters for k, v in d.items()})` to `browse_data(**filters)`

if we adjust the filters object to be a dict rather than list of dict, i.e

```
filters = {}
if transferring_body_id:
    filters["transferring_body_id] = transferring_body_id
....
```

And further to that I think we could get away with just the following 1 line rather than needing the if else block:

```
browse_results = browse_data(**filters)
```

but not a big deal

pgandhizaizi: I think, filter list /dictionary won't be required when we integrate as being passed as individual parameters to browse_data query

e2e_tests/test_browse.py
colinbowen: This was to test that the page loads, it is a bit of a redundant test as it should be covered by other tests in the suite.

e2e_tests/test_browse.py
colinbowen: Added: https://national-archives.atlassian.net/browse/AYR-560

e2e_tests/test_browse.py
colinbowen: I've removed this test for now

e2e_tests/test_browse.py
colinbowen: Added: https://national-archives.atlassian.net/browse/AYR-564

app/main/routes.py
colinbowen: This will be changed to account for single queries which should be good for browse. When it comes to filter functionality we may want to consider compounding filters.

PR #71, opened 2023-11-22
Ayr 446/flask app in lambda

anthonyhashemi:
## Changes in this PR

- Added zappa as poetry dependency
- Added `zappa_settings.json.template` and `aws_attach_policy.json`
- Add Deployment section in `README.md` so devs can run manual deployments to a Lambda-APIGateway set up on AWS, given a AWS profile of their choice, from their local machines.

## JIRA ticket

https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-446&text=poetry


zappa_settings.json.template
```diff
@@ -0,0 +1,16 @@
+{
+    "sandbox": {
+        "app_function": "main_app.app",
+        "memory_size": 512,
+        "keep_warm": false,
+        "aws_region": "eu-west-2",
+        "project_name": "da-ayr-flask-la",
```
colinbowen: Is there a reason for the project name to end in "la"?

colinbowen: If this is for lambda-API, this seems ok to me.

colinbowen:
PR looks clear to me with a great update to the README.

ben-cerium:
This ok to get you started.  The IAM policy we can slim down in the following days.   
Note that the IP address restrictions have been handcranked on this currently live resource so we need make sure that stays in place.  We have that policy in code but don't want to put that in a public repo so we need to chat about that also

ben-cerium:
Might want to consider adding a lambda timeout.  Think it's 30 seconds by default which is probably too high e.g.
   ` "timeout_seconds": 7,` happy if this this is put in later

zaiziclebby (commented):
Seems ok to me

zappa_settings.json.template
anthonyhashemi: it was what zappa created when I first ran zappa init. I think it shortened that from flask-lambda-app or something. So it is worth changing tbf.

pgandhizaizi (approved):
very detailed explanation in readme.md  -looks good
Please check for default lambda timeout and may be reduce to a level where it has to set minimum value

anthonyhashemi:
> This ok to get you started. The IAM policy we can slim down in the following days. Note that the IP address restrictions have been handcranked on this currently live resource so we need make sure that stays in place. We have that policy in code but don't want to put that in a public repo so we need to chat about that also

So from when I have run this, it doesnt seem to overwright any of the ip stuff you have set manually. I believe that is because I am not attaching a policy for the API Gateway but just for the lambda. If we were to do that we would need to make sure the ips wereset appropriately and were filled in by env vars in github actions so they werent commited to the repo, but I think for now we can get away without it. Worth tidying it up in a follow up PR dedicated to the API Gateway policy if you agree?

And yes you are right, most of that policy is uneeded, we currently just need the Paramter store access so I will remove everything apart from that. And then once we sort the rds stuff I will add that in as needed.

anthonyhashemi:
> Might want to consider adding a lambda timeout. Think it's 30 seconds by default which is probably too high e.g. ` "timeout_seconds": 7,` happy if this this is put in later

good shout, can do this here

anthonyhashemi:
> > This ok to get you started. The IAM policy we can slim down in the following days. Note that the IP address restrictions have been handcranked on this currently live resource so we need make sure that stays in place. We have that policy in code but don't want to put that in a public repo so we need to chat about that also
> 
> So from when I have run this, it doesnt seem to overwright any of the ip stuff you have set manually. I believe that is because I am not attaching a policy for the API Gateway but just for the lambda. If we were to do that we would need to make sure the ips wereset appropriately and were filled in by env vars in github actions so they werent commited to the repo, but I think for now we can get away without it. Worth tidying it up in a follow up PR dedicated to the API Gateway policy if you agree?
> 
> And yes you are right, most of that policy is uneeded, we currently just need the Paramter store access so I will remove everything apart from that. And then once we sort the rds stuff I will add that in as needed.

have added apigateway policy with ip whitelisting being set by env vars

zappa_settings.json.template
anthonyhashemi: have updated name, undeployed project with previous name and deployed again with new one

PR #77, opened 2023-11-28
Ayr 489 swap open search to postgres db

pgandhizaizi:
## Changes in this PR

1) modified create_app function in init py to initialise db object 
2) modified conftest.py to include mock database settings
3) created models.py to generate database table object mapping
4) created queries.py to generate query to fetch data from postgresSQL database
5) created mockdatabase.py to create mock data for test database

## JIRA ticket

AYR-489
 
## Screenshots of UI changes
Before

https://zaizi.slack.com/archives/D0619KE96CA/p1701164620733569 

After

image

### After


app/templates/main/poc-search.html
anthonyhashemi, line 1: Did you end up confirming with @Terry-Price whether we wanted to change the poc search view?
Seeing as this ticket was just for swapping out the db seems like we could give this a miss until we have requirements for how the search view should evolve?

@colinbowen @zaiziclebby Since terry is away, do you know of any new search page designs?

configs/base_config.py
```diff
@@ -15,20 +15,32 @@ def AWS_REGION(self):
...
+        return self._get_config_value("AWS_RDS_DB_PASSWORD")
+
+    @property
+    def AWS_RDS_DB_PORT(self):
+        return self._get_config_value("AWS_RDS_DB_PORT")
+
+    @property
+    def SQLALCHEMY_DATABASE_URI(self):
+        return (
+            f'postgresql+psycopg2://{self._get_config_value("AWS_RDS_DB_USERNAME")}:'
```
anthonyhashemi: You dont need to get the value again, you can just access the properties:

```
f'postgresql+psycopg2://{self.AWS_RDS_DB_USERNAME}:'
f'{self:AWS_RDS_DB_PASSWORD}:{self.AWS_RDS_DB_HOST}:'
f'{self.AWS_RDS_DB_PORT}/{self.AWS_RDS_DB_NAME}`
```


Alternatively, if we don't use any of the individual values throughout the codebase and only ever access `SQLALCHEMY_DATABASE_URI`, we _could_ get rid of the properties such as `AWS_RDS_DB_HOST` etc and just keep `SQLALCHEMY_DATABASE_URI` making the explicit calls to `_get_config_value`.

app/tests/mock_database.py
```diff
@@ -0,0 +1,141 @@
+import uuid
+
+from app.main.db.models import Body, Consignment, File, FileMetadata, Series, db
+
+
+def generate_mock_data():
```
anthonyhashemi: Could rename this function to something more specific like `create_two_files_with_test_string_in_fields` so that a reader looking at the test can quickly understand why the query without "test" in it, would result in 0 results, but one with "test" in it would result in 2!

configs/base_config.py
pgandhizaizi: amended to one SQLALCHEMY_DATABASE_URI to retrieve from parameter store or environment variable

app/tests/mock_database.py
```diff
@@ -0,0 +1,141 @@
...
+    # first record
```
anthonyhashemi, line 7: Also, not for this PR, but we should definitely start using something like https://factoryboy.readthedocs.io/en/stable/orms.html#sqlalchemy to create test db objects. (I have used it before for Django and its great)

This will allow us to abstract away object creation to the factories and reduce the amount of code we need to write due to the nested nature of the factory creation and default values.

app/templates/main/poc-search.html
pgandhizaizi: Terry did mentioned that both browse and search view should be same, but it not confirm which fields will be available as this will change based on survey with users, so the current version is not he final one

app/main/db/queries.py
```diff
@@ -0,0 +1,72 @@
...
+
+        for r in result:
+            record = {
+                "TransferringBody": r.TransferringBody,
+                "Series": r.Series,
+                "ConsignmentReference": r.ConsignmentReference,
+                "FileName": r.FileName,
+            }
+            results.append(record)
+    except Exception as e:
```
anthonyhashemi: Best not to catch general exceptions, can we catch specific Exceptions around the tightest block of code we expect that exception?

If the purpose of this check is that `db.session.execute(query)` errors, let's catch the error we are expecting to get.

app/tests/mock_database.py
pgandhizaizi: the function is creating mock data,  when called in test_search_view_with_no_results test, we are entering query search value which is specifically should not return a value, so I don't think we should change function name here , unless @colinbowen if you have any different opinion

app/main/db/queries.py
pgandhizaizi: modified exception to exc.SQLAlchemyError specific exception

app/main/db/models.py
```diff
@@ -0,0 +1,72 @@
+from flask_sqlalchemy import SQLAlchemy
+from sqlalchemy import Boolean, DateTime, ForeignKey, String, Text
+from sqlalchemy.dialects.postgresql import UUID
+
+db = SQLAlchemy()
+
+
```
anthonyhashemi, line 7: @pashworth-tna worth you taking a look at these models, but we can discuss them further in the meeting on Thursday about db, schema, db maintenance. I think this file will be a great place to focus on in that discussion.


Some quick thoughts on database migrations (i.e when the schema changes or a table is created etc.) in flask apps using sqlalchemy:


There is:
- [Alembic](https://alembic.sqlalchemy.org/en/latest/) which is the most popular SQLAlchemy db migration package for python
- [https://flask-migrate.readthedocs.io/en/latest/#example](flask-migrate) which uses alembic under the hood but just configures some boilerplatey flask connectivity for us which looks well used

On the other hand there is also
- pyway which is the python package for flyway.


This blog post on alembic is really good but in particular this section was interesting: https://thinhdanggroup.github.io/alembic-python/#alembic-vs-flyway

```
Alembic vs Flyway
Flyway is a database migration tool that works with SQL scripts. It is designed to be database-agnostic and supports a wide range of databases.

Unlike Alembic, which allows you to define migrations using Python code, Flyway uses SQL-based migration scripts that are organized in a specific directory structure. This makes Flyway a good choice if you prefer a SQL-based solution that works with a variety of databases.

However, if you prefer a Python-based solution with more flexibility, Alembic might be a better choice.
```


If we decide that the database is not the responsibility of the webapp, then we need to figure out a good way to update our ORM tables.  Manually updating the models code every time the external schema changes seems very onerous.

Question: Where is the schema currently defined? How does Talend currently interact with that?


If we decide that the database is the ayr webapp's responsibility, which I think is currently the consensus , then I think the source of truth of the table schema should be our Model definitions as defined here in our SQLAlchemy ORM models, and we should use Alembic (either purely or through flask-migrate) to handle any migrations. 

But then we need to figure out a way to expose this schema to the system that writes to the databse (Talend). I think we can do this by exposing an API that the Talend or whatever microservice surrounding talend can send the output of talend to, rather than Talend speaking directly to the db. That way Talned has a clear API schema to write to, and which will be updated on new releases of the webapp.

Anyway, I will stop rambling and look forward to discussing on Thursday as there's a lot to get our heads around!

app/__init__.py
```diff
@@ -25,7 +27,7 @@ def null_to_dash(value):
     return value
 
 
-def create_app(config_class):
+def create_app(config_class, database_uri=None):
```
anthonyhashemi, line 30: @colinbowen we know this seems a bit hacky but it was the easiest way to be able to manually set the db url in the Test App crated by our app fixture since we spin up a test postgres db using testing.postgres, and so the url is dynamic and cant be defined in the Config before runtime.

app/tests/mock_database.py
anthonyhashemi: It's not creating mocks, it is creating actual database objects, but we only use it in our tests and so is only ever used with the test db we spin up.

And my rename suggestion was because a user can't easily verify that:
- `form_data = {"query": "junk"}` should return 0 objects,
-  `form_data = {"query": "test"}` should return 2 objects,

without having to go and read the function `generate_mock_data`, whereas if this was named `create_two_files_with_test_string_in_fields` which is a much more explicit description of the function, enough for a user not to have to go and read the whole function to understand if the test is doing the right thing or not

app/templates/main/poc-search.html
colinbowen: The latest search view can be found on Axure (Browse Iteration 1). Which is a revised version of what was previously on the POC search.

anthonyhashemi: @colinbowen is that not a browse view?

app/main/db/queries.py
anthonyhashemi: not sure if you have pushed that change, or is github playing up

app/main/db/models.py
colinbowen: I've used flask migrate before which yes does use Alembic under the hood but abstracts away some of the lower level stuff and should be easier to work with. 

Ideally there should be one source of truth that is then replicated elsewhere which sounds like we definitely do need an API to handle this.

anthonyhashemi: yeah, if we decide the db is defined outside the webapp then I think whatever that is it needs an api around it for us to interact with.

Down the line, we will be making calls to external apis anyway that once we are interacting with axiel collections etc. rather than reading from a db in the webapp itself but since we already have a flask app we get an api for free rather than needing to create a new microservice essentially which might not be needed down the road?

app/main/db/queries.py
```diff
@@ -0,0 +1,72 @@
...
+
+        for r in result:
+            record = {
+                "TransferringBody": r.TransferringBody,
+                "Series": r.Series,
+                "ConsignmentReference": r.ConsignmentReference,
+                "FileName": r.FileName,
+            }
+            results.append(record)
+    except exc.SQLAlchemyError as e:
```
anthonyhashemi, line 70: which calls may raise SQLAlchemyError, just `db.session.execute(query)`? If so can we adjust this to 

```

    try:
        result = db.session.execute(query)

app/templates/main/poc-search.html
colinbowen: Yeah the search & browse view are combined now. The search box will always be visible on all browse pages.

anthonyhashemi (approved):
Looks good, left a couple maintainability comments but good to merge 👍

app/__init__.py
colinbowen: Sure, something to just be aware of so thanks for the heads up

colinbowen (approved):

app/main/db/queries.py
pgandhizaizi: updated code base to limit except to SQLAlchemyError

app/main/db/queries.py
anthonyhashemi: I see it now

app/templates/main/poc-search.html
anthonyhashemi: I meant what will the results table look like for search? No point changing the poc search if the search results wont look like the changes made here

app/tests/mock_database.py
anthonyhashemi: created https://national-archives.atlassian.net/browse/AYR-512 to add it in if not done otherwise

app/tests/mock_database.py
pgandhizaizi: renamed the function create_two_test_records

app/main/db/models.py
pgandhizaizi: I will review it , based on the discussion the converstation has been marked as resolved

app/templates/main/poc-search.html
pgandhizaizi: as discussed over the call, we will keep current changes on poc search as they are closed to browse view

PR #78, opened 2023-11-28
browse view functionality 

pgandhizaizi:
## Changes in this PR

1) created new browse view query in queries.py
2) created a new view poc-browse.html to show result of browse query
3) added a new route for poc-browse to show poc-browse.html page
4) created test_browse.py to perform testing

## JIRA ticket
AYR - 455

## Screenshots of UI changes
image

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -72,20 +72,20 @@ def fuzzy_search(query_string):
     return results
 
 
-def browse_view():
+def get_file_data_grouped_by_transferring_body_and_series():
     results = []
 
     query = (
         db.select(
-            Body.Name.label("TransferringBody"),
-            Series.Name.label("Series"),
+            Body.Name.label("transferring_body"),
+            Series.Name.label("series"),
             func.max(Consignment.TransferCompleteDatetime).label(
-                "Last_Record_Transferred"
+                "last_record_transferred"
             ),
             func.count(func.distinct(Consignment.ConsignmentReference)).label(
-                "Consignment_in_Series"
+                "consignment_in_series"
             ),
-            func.count(func.distinct(File.FileId)).label("Records_Held"),
+            func.count(func.distinct(File.FileId)).label("records_held"),
         )
         .join(Consignment, Consignment.ConsignmentId == File.ConsignmentId)
         .join(Body, Body.BodyId == Consignment.BodyId)
@@ -93,19 +93,20 @@ def browse_view():
         .group_by(Body.BodyId, Series.SeriesId)
         .order_by(Body.Name, Series.Name)
     )
-
+    query_results = None
     try:
         query_results = db.session.execute(query)
-
-        for r in query_results:
-            record = {
-                "TransferringBody": r.TransferringBody,
-                "Series": r.Series,
-                "Consignment_in_Series": r.Consignment_in_Series,
-                "Last_Record_Transferred": r.Last_Record_Transferred,
-                "Records_Held": r.Records_Held,
-            }
-            results.append(record)
     except exc.SQLAlchemyError as e:
         print("Failed to return results from database with error : " + str(e))
+
+    if query_results is not None:
+        for r in query_results:
+            record = {
+                "transferring_body": r.transferring_body,
+                "series": r.series,
+                "consignment_in_series": r.consignment_in_series,
+                "last_record_transferred": r.last_record_transferred,
+                "records_held": r.records_held,
+            }
+            results.append(record)
     return results
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
     except exc.SQLAlchemyError as e:
         print("Failed to return results from database with error : " + str(e))
     return results
+
+
+def browse_view():
+    results = []
+
+    query = (
+        db.select(
+            Body.Name.label("TransferringBody"),
+            Series.Name.label("Series"),
+            func.max(Consignment.TransferCompleteDatetime).label(
+                "Last_Record_Transferred"
+            ),
+            func.count(func.distinct(Consignment.ConsignmentReference)).label(
+                "Consignment_in_Series"
+            ),
+            func.count(func.distinct(File.FileId)).label("Records_Held"),
+        )
+        .join(Consignment, Consignment.ConsignmentId == File.ConsignmentId)
+        .join(Body, Body.BodyId == Consignment.BodyId)
+        .join(Series, Series.SeriesId == Consignment.SeriesId)
+        .group_by(Body.BodyId, Series.SeriesId)
+        .order_by(Body.Name, Series.Name)
+    )
+
+    try:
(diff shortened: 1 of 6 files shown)


app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
     except exc.SQLAlchemyError as e:
         print("Failed to return results from database with error : " + str(e))
     return results
+
+
+def browse_view():
```
anthonyhashemi: can we rename this to?
```suggestion
def get_file_data_grouped_by_transferring_body_and_series():
```

app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
...
+
+    query = (
+        db.select(
+            Body.Name.label("TransferringBody"),
+            Series.Name.label("Series"),
+            func.max(Consignment.TransferCompleteDatetime).label(
+                "Last_Record_Transferred"
+            ),
+            func.count(func.distinct(Consignment.ConsignmentReference)).label(
+                "Consignment_in_Series"
```
anthonyhashemi: Could we make all these labels snake_case?

e.g.

```suggestion
                "consignment_in_series"
```

app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
...
+        .join(Body, Body.BodyId == Consignment.BodyId)
+        .join(Series, Series.SeriesId == Consignment.SeriesId)
+        .group_by(Body.BodyId, Series.SeriesId)
+        .order_by(Body.Name, Series.Name)
+    )
+
+    try:
+        query_results = db.session.execute(query)
+
+        for r in query_results:
```
anthonyhashemi, line 103: can we not just access the results object in the view rather than creating this mapping?

app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
...
+            record = {
+                "TransferringBody": r.TransferringBody,
+                "Series": r.Series,
+                "Consignment_in_Series": r.Consignment_in_Series,
+                "Last_Record_Transferred": r.Last_Record_Transferred,
+                "Records_Held": r.Records_Held,
+            }
+            results.append(record)
+    except exc.SQLAlchemyError as e:
```
anthonyhashemi: If we would get any SQLAlchemyError this will happen on line 98, therefore we can focus the try except on that line alone to be more precise.

app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
...
+        print("Failed to return results from database with error : " + str(e))
```
anthonyhashemi: Thinking as an end user, if the application or database is broken for one reason or another, would I want to see:
- 0 results
- internal server error page?


0 results with no indicattion that there is a problem with the app would be concerning to me. But maybe thats a personal preference.

Could be a question for UX.

app/tests/mock_database.py
```diff
@@ -139,3 +139,5 @@ def create_two_test_records():
     )
     db.session.add(new_file_metadata)
     db.session.commit()
+
+    # third record
```
anthonyhashemi: can we delete this comment?

app/tests/test_browse.py
```diff
@@ -0,0 +1,97 @@
...
+            result.text for result in row.find_all("td")
+        ] == expected_results_table[row_index + 1]
+
+
+@patch("app.main.db.queries.db")
+def test_fuzzy_search_exception_raised(db, capsys):
+    """
+    Given a fuzzy search function
+    When a call made to fuzzy search , when database execution failed with error
+    Then list should be empty and should raise an exception
```
anthonyhashemi: It logs a message
```suggestion
    Then list should be empty and should log a message detailing the exception
```

app/main/routes.py
```diff
@@ -127,6 +127,24 @@ def poc_search():
...
+@bp.route("/poc-browse", methods=["POST", "GET"])
+@access_token_sign_in_required
+def poc_browse():
+    form = SearchForm()
+
+    search_results = browse_view()
+    session["search_results"] = search_results
+
+    num_records_found = len(search_results)
+    # print(num_records_found)
```
anthonyhashemi: can remove comment but I guess this poc browse will be deleted when Colin integrates so not big deal

app/templates/main/poc-browse.html
```diff
@@ -0,0 +1,66 @@
+{% extends "base.html" %}
```
anthonyhashemi, line 1: Havent reviewed this code as Colin's real browse page to supersede this

app/main/db/queries.py
```diff
@@ -70,3 +70,42 @@ def fuzzy_search(query_string):
     except exc.SQLAlchemyError as e:
         print("Failed to return results from database with error : " + str(e))
     return results
+
+
+def browse_view():
+    results = []
+
+    query = (
```
anthonyhashemi, line 78: Query nice and clear and makes sense to me!

anthonyhashemi (approved):
Nice work. Have left a few comments but mainly around integration bits which could be covered by the integration PR between you and @colinbowen

app/main/db/queries.py
pgandhizaizi: function renamed as requested

app/main/db/queries.py
pgandhizaizi: variables are set with snake_case a requested

app/main/db/queries.py
pgandhizaizi: modified code to reflect only line 98 in try catch block

app/main/db/queries.py
pgandhizaizi: This is a wider discussion to get confirmation on what needs to be responded to frontend when error occurs

app/main/db/queries.py
pgandhizaizi: query result from database is by default not serializable , Object of type ChunkedIteratorResult is not JSON serializable , and not easier to convert it to json object, due to that used above code to send data to frontend

app/tests/mock_database.py
pgandhizaizi: comment removed, used for testing signed commit

app/tests/test_browse.py
pgandhizaizi: modified test_browse comment to reflect the changes as requested

app/main/routes.py
pgandhizaizi: removed print statment

app/templates/main/poc-browse.html
pgandhizaizi: the view will be superseeded with frontend browse view, so no need to review it

app/main/db/queries.py
pgandhizaizi: thank you for good word, nothing has been highlighted by Colin so mark the conversation as resolved

PR #80, opened 2023-11-29
series view changes

pgandhizaizi:
…t testing, renamed file to test_access_token_sign_in_required.py

1. queries.py --> added a new query to get search result based on series filter
2. routes.py --> added a new route to display search results for series filter
3. poc-browse-series.html --> added a new view to display search results
4. test_access_token_sign_in_required.py --> renamed from 'test_keycloak_login_required_decorator.py ', modified protected route list
5. test_browse_series.py --> added a new file to perform unit testing on series filter

## JIRA ticket

AYR - 516
## Screenshots of UI changes

image


app/templates/main/poc-browse-series.html
```diff
@@ -0,0 +1,69 @@
+{% extends "base.html" %}
```
anthonyhashemi, line 1: looks very similar to the normal (consignment) browse view, I think it would be good to create an intermediary base template if possible to reduce duplication.

Again this is just your poc view of it but for the proper view I mean.

@colinbowen what do you think?

app/main/routes.py
```diff
@@ -127,6 +127,30 @@ def poc_search():
     )
 
 
+@bp.route("/poc-browse-series", methods=["POST", "GET"])
+@access_token_sign_in_required
+def poc_browse_series():
+    form = SearchForm()
+    search_results = []
+    series = request.form.get("series", "").lower()
```
anthonyhashemi, line 157: If there is a separate route for series browse then we dont need the "series" query param - would only need it if we share one browse view.

@colinbowen has @Terry-Price spoken to you about how the consignmen vs series browse pages will differ in that sense?

app/main/db/queries.py
```diff
@@ -70,3 +70,50 @@ def fuzzy_search(query_string):
...
+                "Records_Held": r.Records_Held,
+                "ConsignmentReference": r.Consignment_Reference,
+            }
+            results.append(record)
+    except exc.SQLAlchemyError as e:
+        print("Failed to return results from database with error : " + str(e))
+    return results
+
+
+def get_full_list_of_series():
```
anthonyhashemi: this isnt being used anywhere it seems?

app/main/db/queries.py
```diff
@@ -70,3 +70,50 @@ def fuzzy_search(query_string):
     except exc.SQLAlchemyError as e:
         print("Failed to return results from database with error : " + str(e))
     return results
+
+
+def browse_view_series(series):
```
anthonyhashemi: is this the same as the function in the normal browse PR?

app/templates/main/poc-browse-series.html
pgandhizaizi: poc-browse-series view will be redundant and super seeded by front end browse designed by Colin

app/main/routes.py
pgandhizaizi: poc_browse_series will be redundant and super seeded by front end browse route designed by Colin

app/main/db/queries.py
pgandhizaizi: all browse view queries will be combined into one or two queries and parameters will be passed as per requirement i.e. transferring body, series , consignment reference and so on.

app/main/db/queries.py
pgandhizaizi: get_full_list_of_series() will be removed in ticker AYR - 457 which I am currently working on

pgandhizaizi:
superseded by AYR-457 ticket, no more need to pull the request

PR #82, opened 2023-12-04
Add pe support to playwright

colinbowen:
## Changes in this PR
- Modifies the playwright browser context to allow for tests to run with / without JavaScript. 
- Adds block_resource_requests which can be imported into other tests to block CSS from loading.

## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-469

## Screenshots of UI changes
N/A

### After


conftest.py
pgandhizaizi, line 1: do we need to enable java script for testing?

conftest.py
```diff
@@ -0,0 +1,24 @@
...
+    Parameters:
+        browser_context_args (dict): The default browser context arguments.
+
+    Returns:
+        dict: Updated browser context arguments with customized settings.
+    """
+    return {
+        **browser_context_args,
+        "ignore_https_errors": True,
+        "java_script_enabled": True,
```
anthonyhashemi: So when writing tests by default when using the browser fixture it will have js enabled, and if we want to have it run without js we need to pass in this kwarg somewhere?

Could you add some documentation in the readme around how we might write a test that needs js off?

Also, if the baseline is that the page should work without js and css, maybe we should have that as the default?

Then, only when we have something interesting that we specifically want to test that relies on js we can turn it on for that specific tests?

colinbowen: Yeah that is a better process of running the tests. I'll add in a test and example on how to run it inside this pr once I've made this change.

README.md
```diff
@@ -202,6 +202,12 @@ To enable this flexibility we suggest any Playwright tests added to the repo use
 
 In addition, we recommend that any tests that have dependencies on data, do not make assumptions about any particular database or instance involved, and instead do the test data set up and teardown as part of the test suite.
 
+### E2E Tests (Progressive Enhancement Support)
+
+E2E Tests by default run without JavaScript. To also disable CSS within tests the decorator `block_css_decorator` should be used which will block all css stylesheets from being loaded on the page.
```
anthonyhashemi: we could make the block_css_decorator an autouse fixture so all tests block css by default?

But maybe just mentioning that using the page fixture will block css and we recommend using that is sufficient for now

anthonyhashemi (approved):
Nice one!

e2e_tests/test_homepage.py
```diff
@@ -8,3 +8,10 @@ def test_has_title(page: Page):
 
     # Expect a title "to contain" a substring.
     expect(page).to_have_title(re.compile("AYR - Access Your Records – GOV.UK"))
+
+
+def test_css_has_title(page: Page):
```
anthonyhashemi, line 13: is this just an example of activtaing css in tests, theres no css here we particularly want to /are  testing right?

conftest.py
pgandhizaizi: I agree with Anthony to add flexibility to enable or disable javascript for test.  looks good to me

pgandhizaizi (commented):
looks good to me

PR #85, opened 2023-12-05
AYR-457 updated queries to send Ids to frontend, amended unit testing

pgandhizaizi:
## Changes in this PR
access_token_sign_in_required.py --> updated code to make a call to decode_token function from keycloak_manager.py
keycloak_manager.py --> added a new file
queries.py --> amended query logic to return Ids and combine multiple browse queries
poc-search.html --> amended to refer snake case variables
conftest.py --> amended client fixture to scope on function level and amend code to drop and re-create database objects
test_access_token_sign_in_required.py --> amended to use patch from keycloak_manager for keycloak introspect
test_queries.py --> added new file to perform unit testing on queries

### After


pgandhizaizi:
added a new file test_keycloak_manager.py with all test cases

app/tests/test_queries.py
```diff
@@ -0,0 +1,211 @@
...
+from app.tests.mock_database import create_two_test_records
+
+
+def test_fuzzy_search_no_results(client: FlaskClient):
+    """
+    Given a user with a search query
+    When they make a request, and no results are found
+    Then they should see no records found.
+    """
+    # create_two_test_records()
```
anthonyhashemi: can remove this comment

pgandhizaizi: done, just checked in the changes

app/main/db/queries.py
```diff
@@ -54,69 +59,130 @@ def fuzzy_search(query_string):
...
+                for user_group in user_transferring_body_groups:
+                    if (
+                        user_group.strip().replace(" ", "").lower()
+                        == group_name.strip().replace(" ", "").lower()
+                    ):
+                        unique_transferring_bodies.append(group_name)
+    return unique_transferring_bodies
+
+
+def get_file_meta_data(file_id):
```
anthonyhashemi:
```suggestion
def get_file_metadata(file_id):
```

app/main/db/queries.py
```diff
@@ -54,69 +59,130 @@ def fuzzy_search(query_string):
...
+    )
+    query_results = None
+    try:
+        query_results = db.session.execute(query)
+    except exc.SQLAlchemyError as e:
+        print("Failed to return results from database with error : " + str(e))
+
+    if query_results is not None:
+        for r in query_results:
+            record = {"PropertyName": r.PropertyName, "PropertyValue": r.Value}
```
anthonyhashemi:
```suggestion
            record = {"property_name": r.PropertyName, "property_value": r.Value}
```

app/main/db/queries.py
```diff
@@ -54,69 +59,130 @@ def fuzzy_search(query_string):
...
+            Body.Name.label("TransferringBody"),
+            Series.SeriesId.label("SeriesId"),
+            Series.Name.label("Series"),
             func.max(Consignment.TransferCompleteDatetime).label(
-                "last_record_transferred"
+                "Last_Record_Transferred"
             ),
             func.count(func.distinct(Consignment.ConsignmentReference)).label(
-                "consignment_in_series"
+                "Consignment_in_Series"
```
anthonyhashemi: can we make all the labels throughout this PR snake case?

Or at least can we make them consistently CamelCase if its going to be weird to have model fields as CamelCase and labels as snake case?

```suggestion
                "consignment_in_series"
```

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+            "/transferring_body_user/test body1",
+            "/transferring_body_user/test body2",
+            "/ayr_user/bar",
+        ],
+    }
+    results = get_user_accessible_transferring_bodies(
+        mock_decode_keycloak_access_token
+    )
+    assert len(results) == 2
+    assert results[0] == "test body1"
```
anthonyhashemi: can we assert both results. e.g. ["test body1", "test body2"]

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+def test_get_file_meta_data_with_results(client: FlaskClient):
+    """
+    Given a user with a search query which should return n results
+    When they make a request on the browse series page
+    Then a table is populated with the n results with metadata fields.
+    """
+    create_two_test_records()
+    files = File.query.all()
+
+    if len(files) > 0:
```
anthonyhashemi: Not a good idea to have conditional logic in tests and we shouldnt need to have this conditional statement here. files should always be greater than length 0, so I think we can just remove it.

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+def test_browse_data_with_consignment_reference_filter(client: FlaskClient):
+    """
+    Given a user accessing the browse view query
+    When they make a POST request with a consignment reference filter
+    Then it should return records matched to consignment reference filter.
+    """
+    create_two_test_records()
+    consignments = Consignment.query.all()
+
+    if len(consignments) > 0:
```
anthonyhashemi: same here

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+def test_browse_data_with_series_filter(client: FlaskClient):
+    """
+    Given a user accessing the browse view query
+    When they make a POST request with a series filter
+    Then it should return records matched to series filter.
+    """
+    create_two_test_records()
+    series = Series.query.all()
+
+    if len(series) > 0:
```
anthonyhashemi: and here

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+def test_browse_data_with_transferring_body_filter(client: FlaskClient):
+    """
+    Given a user accessing the browse view query
+    When they make a POST request with a transferring body filter
+    Then it should return records matched to transferring body filter.
+    """
+    create_two_test_records()
+    bodies = Body.query.all()
+
+    if len(bodies) > 0:
```
anthonyhashemi: and here

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+    When they make a request on the browse series page
+    Then a table is populated with the n results with metadata fields.
+    """
+    create_two_test_records()
+    files = File.query.all()
+
+    if len(files) > 0:
+        file_id = files[0].FileId
+        search_results = get_file_meta_data(file_id=file_id)
+        assert len(search_results) == 3
```
anthonyhashemi: Since we only need one File to test this query, I think we can simplify the setup by just creating one File with some related file metadata objects rather than creating 2 records and then grabbing one of them from the db.

Again, once we start using FactoryBoy we can make test setup simpler. Could be worth incorporating it here but for now, to get this merged in quicker can just copy one part of that create two test records function.

Either way, I think it would be worth making the docstring more explciit so its easier to understand whats happening E.g.:

```
    """
    Given a File object with 3 related metadata objects
    When get_file_meta_data is called with the file's id
    Then 3 FileMetadata results are returned
    """
```

app/main/db/queries.py
pgandhizaizi: function name changed to get_file_metadata as requested

app/main/db/queries.py
pgandhizaizi: changed to snake case as requested

app/tests/test_queries.py
pgandhizaizi: condition has been removed from test case as requested

app/tests/test_queries.py
pgandhizaizi: condition has been removed from test case as requested

app/tests/test_queries.py
pgandhizaizi: condition has been removed from test case as requested

app/tests/test_queries.py
pgandhizaizi: condition has been removed from test case as requested

app/tests/test_queries.py
pgandhizaizi: amended assert to include dictionary result

app/tests/test_queries.py
```diff
@@ -0,0 +1,210 @@
...
+
+    query = "junk"
+    search_results = fuzzy_search(query)
+
+    assert len(search_results) == 0
+
+
+def test_fuzzy_search_with_results(client: FlaskClient):
+    """
+    Given a user with a search query
```
anthonyhashemi: can you go through all the docstrings and make sure they are accurately reflect the test.

e.g. here there is no user as its just a function that gets passed a query not a client making a call.

```
Given some data where there is a file with a field matching "test body1" in one of its fields
When fuzzy_search is called with "test body1"
Then it the matching file
```

and then I think its worth asserting all the values on this resut e.g.

```
assert  search_results[0].body_id == 1
assert  search_results[0].transferring_body == "test_transferring_body"
assert  search_results[0].series_id == 4
assert  search_results[0].series == "test_series"
assert  search_results[0].consignment_reference == "test_consignment_reference"
assert  search_results[0].file_name == "file_name"
```

Similarly assert the properties of the results in the other functions too

app/main/db/queries.py
pgandhizaizi: fixed issues with dictionary to refer snake case variable names

anthonyhashemi (approved):

PR #86, opened 2023-12-05
Ayr 506/connect to rds with iam

anthonyhashemi:
## Changes in this PR

- Split out DATABASE URI to separate env vars
- Connect to RDS DB when using AWS Config using rds client to get password through iam
- Add rds permissions to zappa deployment's lambda policy 

Bonus change:
- Removed unnecessary dependencies in non dev section which bloated the zipfile created by zappa when deploying to lambda (playwright is huge)

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-506


colinbowen (approved):
All good, once this is merged in we should update our local envs

PR #88, opened 2023-12-08
Ayr 515/sign out link

colinbowen:
### After


anthonyhashemi (approved):

app/templates/main/signed-out.html
```diff
@@ -16,7 +16,7 @@
         <div class="govuk-grid-column-two-thirds">
           <h1 class="govuk-heading-l">You have successfully signed out</h1>
           <p class="govuk-body-l">Thank you for using Access Your Records.</p>
```
anthonyhashemi, line 18: worth adding a test for each of these hrefs

pgandhizaizi (commented):
looks good, correct link has been assigned now, approved

PR #90, opened 2023-12-11
Update zappa deployment config

anthonyhashemi:
https://national-archives.atlassian.net/browse/AYR-557

## Changes in this PR

- [Specify s3 bucket that zappa lambda zip uploaded to so it doesnt create a new one on each deplyoment update](https://github.com/nationalarchives/da-ayr-beta-webapp/commit/6f58899d5fa77a1ce562951d8755491843c0a79a)
- [Update zappa settings to auto configure lambda to work in externally set up vpc (to work with rds)](https://github.com/nationalarchives/da-ayr-beta-webapp/commit/3393b870543dfbc5b242e8da7e029a624359c40d)
- [Add domain and certificate info to zappa config template and readme (to deploy to a specific domain)](https://github.com/nationalarchives/da-ayr-beta-webapp/commit/1ebff19a9cbf3b11bfd75ff744b60a51230a956c)
- [Add assumption info to the README](https://github.com/nationalarchives/da-ayr-beta-webapp/pull/90/commits/4698cf1f6094bb4e8443645cbddfad35678c6441)
- [Add info about Metadata Store Postgres Database](https://github.com/nationalarchives/da-ayr-beta-webapp/pull/90/commits/f9d0e62e1bbe79345e2fdc119b2fc493bb765344)

### DIFF
--- a/lambda_policy.json
+++ b/lambda_policy.json
@@ -3,6 +3,17 @@
     "Statement": [
         {
             "Effect": "Allow",
+            "Action": [
+              "ec2:DescribeNetworkInterfaces",
+              "ec2:CreateNetworkInterface",
+              "ec2:DeleteNetworkInterface",
+              "ec2:DescribeInstances",
+              "ec2:AttachNetworkInterface"
+            ],
+            "Resource": "*"
+        },
+        {
+            "Effect": "Allow",
             "Action": ["rds-db:connect"],
             "Resource": "*"
         },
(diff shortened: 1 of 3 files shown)


lambda_policy.json
pgandhizaizi, line 1: can the resource be specific AWS component if known instead of ("Resource": "*"), as it widen the accesses assigned in AWS environment

anthonyhashemi: Im not sure if we can know the resource, as I think it is created as another part of the zappa deployment, although I may be wrong.

Either way, Ben has done good work around this for the proper deployments and we will come back to decide whether we want to get rid of these configs from this repo so I dont think its worth investng any more time into this if thats ok

pgandhizaizi: I guess it is fine as you provided a valid reason to implement this approach

ben-cerium (approved):

pgandhizaizi (approved):
looks good , no issues

PR #94, opened 2023-12-12
Ayr 329/playwright tests against deployed env

anthonyhashemi:
## Changes in this PR

- Replaced the apply_block_css_decorator fixture with an overriden page fixture, which was doing something weird, essentially making tests run twice which was breaking any test that required loggin in as by that point we would bypass the login page
- Fixed all out of date urls in e2e tests
- Fixed poc search tests
- Fixed record view tests (but record view still needs some more love)

All tests in e2e_tests folder now pass against local instance and against sandbox. We still want to generalise some of the tests in there / have data setup once we configure CI to run tests but for now this is fine.

I have shown that I can run against a deployed instance, which is the purpose of the ticket.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-329


e2e_tests/conftest.py
colinbowen, line 1: Good work on this, seems all good running local now

colinbowen (approved):

e2e_tests/conftest.py
```diff
@@ -35,7 +34,14 @@ def browser_context_args(browser_context_args):
     }
 
 
-@pytest.fixture(autouse=True)
-def apply_block_css_decorator(request, page):
+@pytest.fixture()
+def page(request, page) -> Page:
```
anthonyhashemi, line 38: The apply_block_css_decorator was doing something weird, essentially making tests run twice which was breaking any test that required loggin in as by that point we would bypass the login page

colinbowen (approved):

PR #96, opened 2023-12-12
added pagination logic and performed unit testing

pgandhizaizi:
## Changes in this PR

1.  added a paginate function to perform pagination
2. modified fuzzy_search and browse_data queries to provide pagination results
3. added a unit testing for pagination for fuzzy search and browse_data functions
4. added new test records in mock_database file to perform pagination testing, unified to one function on mocking data for all test cases
5. modified poc search route to verify pagination functionality

## JIRA ticket

AYR- 518

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -7,7 +7,24 @@ from app.main.authorize.keycloak_manager import (
 from app.main.db.models import Body, Consignment, File, FileMetadata, Series, db
 
 
-def fuzzy_search(query_string):
+def paginate(query, record_count, page, per_page):
+    pages = 1
+    if record_count > per_page:
+        page_mod = record_count % per_page
+        page_cnt = record_count / per_page
+        if page_mod >= 0:
+            pages = int(round(page_cnt, 0))
+        else:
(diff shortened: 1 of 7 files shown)


app/main/db/queries.py
```diff
@@ -62,81 +79,117 @@ def fuzzy_search(query_string):
...
+    if search_results is not None:
+        for r in search_results:
             record = {
-                "transferring_body_id": r.body_id,
-                "transferring_body": r.transferring_body,
-                "series_id": r.series_id,
-                "series": r.series,
-                "consignment_reference": r.consignment_reference,
-                "file_name": r.file_name,
+                "transferring_body_id": r[0],
```
anthonyhashemi, line 105: I thought we decided that we'd stop parsing the query results list to a dictionary when we do the pagination

pgandhizaizi: yes, but default pagination supported by flasks-sqlalchemy run only on base query (i.e. direct model objects) and does not allow the results to paginate with joined queries, I could not find any solution which can help to avoid parsing query result

PR #100, opened 2023-12-15
Configure flask app logger and detail in readme for devs

anthonyhashemi:
## Changes in this PR

- Configure flask app logger and detail in readme for devs

## JIRA ticket

https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?assignee=63b4326678aabbefa9d4400d&selectedIssue=AYR-502

Have verified that the logs come out as expected locally and when running on AWS lambda they come through as expected in cloudwatch logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,48 @@ e.g.:
 def test_css_has_title(page: Page):
 `
 
+## Logging
+
+### Configuration
+
+The application logger configuration is detailed in `setup_logging` in `app/logger_config.py`.
+
+This config includes:
+
+- a formatter specifying request-specific information such as the remote address and URL when a request context is available
+- setting the logging level to `INFO`.
+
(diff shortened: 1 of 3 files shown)


pgandhizaizi (commented):
do we need to configure conftest.py for mocking the logging object ?

anthonyhashemi:
> do we need to configure conftest.py for mocking the logging object ?

Spoken offline, can just use pytests inbuilt fixture, caplog. Will add a line about this

anthonyhashemi:
> > do we need to configure conftest.py for mocking the logging object ?
> 
> Spoken offline, can just use pytests inbuilt fixture, caplog. Will add a line about this

Done

pgandhizaizi (approved):
looks good , approved

PR #104, opened 2023-12-20
SCSS Files reformatted 

zaiziclebby:
## Changes in this PR

Following SCSS files reformatted with BEM:- 

- Header
- Footer
- Signed-out page
- Index 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-550
## Screenshots of UI changes
N/A
### After

### DIFF
--- a/app/static/src/scss/footer.scss
+++ b/app/static/src/scss/footer.scss
@@ -1,3 +1,5 @@
+/* SCSS for Footer */
+
 @mixin on-mobile{
    @media screen and (width <= 640px){
        @content;
--- a/app/static/src/scss/footer.scss
+++ b/app/static/src/scss/footer.scss
@@ -1,57 +1,66 @@
-
-/* CSS for Footer */
+@mixin on-mobile{
+   @media screen and (width <= 640px){
+       @content;
+   }
+}
 
 
 .govuk-footer  {
-    padding-bottom: 40px;
- }
+   padding-bottom: 40px;
 
- .govuk-footer__tna-logo {
-    width: 100px;
- }
 
- .govuk-footer__meta {
-    margin: 0;
-    align-items: center;
-    gap: 40px;
- }
+&__tna-logo {
+   width: 100px;
+}
 
- .govuk-footer__meta-item {
-    display: block;
-    margin: 0;
- }
+&__meta {
+   margin: 0;
+   align-items: center;
+   gap: 40px;
+@include on-mobile {
+flex-direction: column;
+}
+}
 
- .govuk-footer__inline-list-item {
-    margin-bottom: 10px;
- }
+&__meta-item {
+   display: block;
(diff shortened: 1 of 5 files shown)


app/static/src/scss/signed-out.scss
```diff
@@ -1,7 +1,10 @@
...
-    margin-bottom: 10px;
-}
+
+.govuk-button {
+    &--sign-in-again {
+        margin-left: 0;
+        margin-top:0;
+        margin-bottom: 10px;
+    }
+    }
```
anthonyhashemi, line 10:
```suggestion
}
```

app/static/src/scss/footer.scss
```diff
@@ -1,57 +1,68 @@
...
+   width: 100px;
+}
+
+&__meta {
+   margin: 0;
+   align-items: center;
+   gap: 40px;
+@include on-mobile {
+flex-direction: column;
+}
```
anthonyhashemi, line 24: formatting of curly braces looks off but we can enforce the formatting like we discussed by adding prettier to the precommit hooks in the repo:

Add the following to `.pre-commit-config.yaml`


```
- repo: https://github.com/pre-commit/mirrors-prettier
  rev: "v3.1.0"
  hooks:
    - id: prettier
```

app/static/src/scss/footer.scss
```diff
@@ -1,57 +1,68 @@
...
+       display:block;
+       margin-right: 0;
+       margin-bottom: 0;
+       text-align: center;
+       padding: 10px 0;
+}
+}
+
+&__link {
+@include on-mobile {
```
anthonyhashemi, line 44: is this pattern of

```
&__<subclass> {
@include on-mobile {
some_property: value
```

just reducing duplication so that the mobile styling is included on all components? or is this something else?

If so, does that mean we need to always have the 
`@include on-mobile {` line in every scss definition we make?

app/static/src/scss/index.scss
```diff
@@ -1,8 +1,11 @@
-.govuk-button--start {
-  position: relative;
-  padding: 8px 10px 7px 7px;
-  text-align: center;
-  border: 2px solid transparent;
+/* SCSS for Index page */
+
+.govuk-button {
+  &--start {
```
anthonyhashemi, line 4: so does this not override the govuk-button as defined by gds?

anthonyhashemi (approved):

app/static/src/scss/index.scss
anthonyhashemi: confirmed fine on call

app/static/src/scss/footer.scss
anthonyhashemi: clarified on call - reducing duplication

PR #107, opened 2023-12-22
fixed pagination for browse page

pgandhizaizi:
## Changes in this PR
1. modified pagination.html to take dynamic filters
2. modified poc_search.html to refer filters["query"] instead of query object
3. modified routes.py, updated browse function to refer 'filters' for pagination 
4. added function create multiple files under one consignment in mock_database.py file
5. updated test cases in test_browse.py
6. updated test case in test_queries.py --> get_file_metadata

## JIRA ticket

AYR-518 

## Screenshots of UI changes
image

### After


app/tests/mock_database.py
```diff
@@ -754,3 +763,406 @@ def create_multiple_test_records():
         file_10,
         file_11,
     ]
+
+
+def create_multiple_files_for_consignment(consignment_id):
```
anthonyhashemi, line 768: can we use Factory boy and for loops to reduce the coe needed for this setup

anthonyhashemi (approved):
Looks good, but have just suggested to simplify the test fixture created

app/tests/mock_database.py
pgandhizaizi: It is too late in the current sprint to implement, but will refactor with a new ticket in future

PR #112, opened 2023-12-29
Extract get_user_groups out of access_token_sign_in_required and refa…

anthonyhashemi:
## Changes in this PR

- Extract get_user_groups out of access_token_sign_in_required and refactor get_user_accessible_transferring_bodies so that it doesnt need the same logic anymore

## JIRA ticket


pgandhizaizi (approved):
looking fine, no changes required

PR #113, opened 2024-01-02
Limit users to their transferring bodies

anthonyhashemi:
## Changes in this PR

- Added `validate_body_user_groups_or_404` function to be reused by any detail view
- Utilised it in /record/record_id/ view so standard users can only see records from the transferring body they are part of.

Followup PR to come for browse-transferring and browse-series views. 

## JIRA ticket


colinbowen (approved):
Looks good

PR #115, opened 2024-01-02
Fix record route by using get_or_404, not one_or_404

anthonyhashemi:
## Changes in this PR

- Fix record route by using get_or_404, not one_or_404

https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/api/#flask_sqlalchemy.query.Query.get_or_404

is what should be used, not https://flask-sqlalchemy.palletsprojects.com/en/3.0.x/api/#flask_sqlalchemy.query.Query.get_or_404 since one_or_404 will error if more than one File in the db.

## JIRA ticket

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -177,7 +177,7 @@ def record(record_id: uuid.UUID):
     Returns:
         A rendered HTML page with record details.
     """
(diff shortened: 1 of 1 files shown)


colinbowen (approved):
Good change

PR #118, opened 2024-01-03
Set record specific scss

colinbowen:
### After

### DIFF
--- a/app/static/src/scss/record.scss
+++ b/app/static/src/scss/record.scss
@@ -133,7 +133,7 @@
     }
 }
 
-.govuk-summary-list {
+.govuk-summary-list-record {
     table-layout: auto;
 
     &__key {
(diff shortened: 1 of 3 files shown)


app/static/src/scss/record.scss
```diff
@@ -133,7 +133,7 @@
     }
 }
 
-.govuk-summary-list {
+.govuk-summary-list-record {
```
anthonyhashemi, line 136: was this name not following the bem or gds convention?

anthonyhashemi (approved):

app/static/src/scss/record.scss
anthonyhashemi: spoke offline. this name is used elsewhere and so there was conflicting styles. instead extend the class for record page specifically

PR #119, opened 2024-01-03
Ayr-580 implement download mvp

rahmedzaizi:
## Changes in this PR
downlaod_record
- function added
- can be called by clicking download button on record view page

## JIRA ticket
[AYR-580](https://national-archives.atlassian.net/browse/AYR-580)
### After

- [x] Requires env variable(s) to be updated

[AYR-580]: https://national-archives.atlassian.net/browse/AYR-580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/main/routes.py
```diff
@@ -187,6 +189,32 @@ def record(record_id: uuid.UUID):
...
+    file_path = file_metadata["file_path"]
+    key = f'{consignment_reference}/{file_path.rstrip("/")}'
+    file_name = file_metadata["file_name"]
+    try:
+        response = Response(
+            s3.get_object(Bucket=bucket, Key=key)["Body"].read(),
+            headers={"Content-Disposition": "attachment;filename=" + file_name},
+        )
+        return response
+    except Exception as e:
```
anthonyhashemi: This try except is currently unncessary since we just reraise the error.

We probably want to catch the error and then do abort(404) where `from flask import abort`


Also, can we catch the specific exception that we expect might be raised when making the call to s3 and put the try except around the narrowest bit of code we belive the error to eminate from. e.g.

```
    try:
        s3_object = s3.get_object(Bucket=bucket, Key=key)
    except ClientError, e:
        logger.error(e)
        abort(404)        

    response = Response(
        s3_object["Body"].read(),
        headers={"Content-Disposition": "attachment;filename=" + file_name},
    )

    return response
```

app/main/routes.py
```diff
@@ -187,6 +189,32 @@ def record(record_id: uuid.UUID):
     return render_template("record.html", record=file_metadata)
 
 
+@bp.route("/download/<uuid:record_id>")
+@access_token_sign_in_required
+def download_record(record_id: uuid.UUID):
```
anthonyhashemi, line 194: looks good, just need some tests using moto to mock s3:

- user with body A can download record part of body A (i.e correct response object returned)
- user with body A cannot download record part of body B (i.e 404 is raised)
- superuser can download any random record with some body (i.e correct response object returned)
- when s3.get raises an error, then we log the error and raise 404


might also be worth an e2e test using playwright where a user navigates to `/record/<ID>` they have access to and then clicks download.

app/templates/main/record.html
```diff
@@ -108,8 +108,9 @@ <h1 class="govuk-heading-m govuk-heading-m__record-details">Record details</h1>
                 <div class="govuk-grid-column-one-third govuk-grid-column-one-third--record">
                     <div class="rights-container">
                         <h3 class="govuk-heading-m govuk-heading-m__rights-header">Rights to access</h3>
-                        <button class="govuk-button govuk-button__download--record"
-                                data-module="govuk-button">Download record</button>
+                        <a href="{{ url_for('main.download_record', record_id=record['file_id']) }}"
```
anthonyhashemi, line 111: syling needs fixing but not sure exactly whats the issue here, worth chatting with @colinbowen and @zaiziclebby

.env.env_var.template
```diff
@@ -20,3 +20,7 @@ export DEFAULT_PAGE_SIZE=5
 export DEFAULT_DATE_FORMAT="DD/MM/YYYY"
 
 export RATELIMIT_STORAGE_URI=memory://
+
+# record bucket variables
+
+export RECORD_BUCKET_NAME=test-record-download
```
anthonyhashemi: could leave this empty to reduce confusion - devs will most likely need to change this

pgandhizaizi: please update the existing config testing in test_config.py to reflect new changes

app/templates/main/record.html
rahmedzaizi: Fixed in the latest commit

.env.env_var.template
rahmedzaizi: Updated testing in test_config.py to reflect new changes

Removed value

app/tests/test_download.py
```diff
@@ -0,0 +1,124 @@
...
+    And the HTML content should contain specific elements
+        related to the record
+    """
+    create_mock_s3_bucket_with_object()
+    app.config["RECORD_BUCKET_NAME"] = BUCKET
+    file = mock_record()
+    mock_standard_user(client, file.file_consignments.consignment_bodies.Name)
+    response = client.get(f"/download/{file.FileId}")
+
+    assert response.status_code == 200
```
anthonyhashemi, line 92: can you add assertions for 


response header
file content

```
    assert response.headers["Content-Disposition"] == "attachment;filename=test_file.txt" 
    assert response.data == "foobar"
```

app/tests/test_download.py
```diff
@@ -0,0 +1,124 @@
...
+            PropertyName=property_name,
+            Value=value,
+        )
+        for property_name, value in metadata.items()
+    ]
+
+    return file
+
+
+@mock_s3
```
anthonyhashemi: I think we can probably remove the mock_s3 decorator either from here or from the tests that call this, can you give that a go?

app/tests/test_download.py
```diff
@@ -0,0 +1,124 @@
...
+    FileFactory,
+    FileMetadataFactory,
+)
+
+BUCKET = "test-download"
+CONSIGNMENT_REF = "TDR-123-TEST"
+FILE_PATH = "data/content/folder_a/test_file.txt"
+
+
+def test_invalid_id_raises_404(client):
```
anthonyhashemi: Can we put this below the setup functions, alongside the other tests, for consistency / easier to read this file.

app/tests/test_download.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    consignment = ConsignmentFactory(ConsignmentReference=CONSIGNMENT_REF)
+    file = FileFactory(
+        file_consignments=consignment,
+        FileName=FILE_NAME,
+        FilePath=FILE_PATH,
+        FileType="file",
+    )
+
+    metadata = {
```
anthonyhashemi, line 28: I think these download tests can get away without creating the file metadata objects

app/tests/test_download.py
```diff
@@ -0,0 +1,136 @@
...
+@mock_s3
+def test_downloads_record_successfully_for_user_with_access_to_files_transferring_body(
+    app, client, mock_standard_user
+):
+    """
+    Given a File in the database
+    When a standard user with access to the file's transferring body makes a
+        request to view the record page
+    Then the response status code should be 200
+    And the HTML content should contain specific elements
```
anthonyhashemi:
```suggestion
    And the file should contain the expected content
```

app/tests/test_download.py
```diff
@@ -0,0 +1,136 @@
...
+        related to the record
```
anthonyhashemi:
```suggestion
```

app/main/routes.py
```diff
@@ -187,6 +189,33 @@ def record(record_id: uuid.UUID):
...
+    key = f'{consignment_reference}/{file_path.rstrip("/")}'
+    file_name = file_metadata["file_name"]
+    file = s3.get_object(Bucket=bucket, Key=key)
+
+    response = Response(
+        file["Body"].read(),
+        headers={"Content-Disposition": "attachment;filename=" + file_name},
+    )
+    if response.status_code == 200:
+        print("ok")  # success message will go here
```
anthonyhashemi: This response code of this will always be 200 since we are creating it above implicitly with the default status code which is 200.

I think the message should be shown on the record page, not on the download page. 

Will reach out to chat offline.

app/tests/test_download.py
rahmedzaizi: I tried it 
The function requires metadata 
`    s3 = boto3.client("s3")
    bucket = current_app.config["RECORD_BUCKET_NAME"]
    file_metadata = get_file_metadata(record_id)`

e2e_tests/test_download.py
```diff
@@ -0,0 +1,9 @@
+from playwright.sync_api import Page
+
+
+def test_has_title(authenticated_page: Page):
```
anthonyhashemi: can we rename this to test_download_record

and rename the page to test_record.py
?

anthonyhashemi (approved):
Really nice work 👍

PR #121, opened 2024-01-04
fixed bug AYR-597

pgandhizaizi:
1 . updated consignment ref query in queries.py (changed join to left join for FileMetaData table), 
2. added a new unit test in test_browse.py, 
3. modified mock_database.py to add a file record without metadata, 
4. modified corresponding test cases in test_browse.py and test_queries.py, 
5. updated npm package in package.json

## JIRA ticket

AYR-597

### Before

when filemetadata not found for file exists in consignment

image

### After

image

### DIFF
--- a/app/templates/main/browse.html
+++ b/app/templates/main/browse.html
@@ -340,7 +340,7 @@
                                                 {% if record['closure_period'] %}
                                                     {{ record['closure_period'] }} years
                                                 {% else %}
-                                                    -
+                                                    None
                                                 {% endif %}
                                             </p>
                                         </dd>
--- a/app/tests/test_browse.py
+++ b/app/tests/test_browse.py
@@ -71,7 +71,7 @@ def test_browse_get_with_data(client: FlaskClient):
             "'test body1', 'test series1', '01/01/2023', '1', '1', "
             "'test body2', 'test series2', '01/01/2023', '1', '1', "
             "'testing body10', 'test series10', '01/01/2023', '1', '1', "
-            "'testing body11', 'test series11', '01/01/2023', '1', '1', "
+            "'testing body11', 'test series11', '01/01/2023', '2', '1', "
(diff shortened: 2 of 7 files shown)


app/tests/test_browse.py
```diff
@@ -448,6 +448,53 @@ def test_browse_consignment(client: FlaskClient):
...
+
+    expected_results_table = [
+        [
+            "Last modified",
+            "Filename",
+            "Status",
+            "Closure start date",
+            "Closure period",
+        ],
+        ["None", "test_file12.txt", "None", "None", "None"],
```
anthonyhashemi, line 484: nice one 👌

app/templates/main/browse.html
```diff
@@ -340,7 +340,7 @@ <h1 class="govuk-heading-l browse__records-found__text">Records found {{ num_rec
                                                 {% if record['closure_period'] %}
                                                     {{ record['closure_period'] }} years
                                                 {% else %}
-                                                    -
+                                                    None
```
anthonyhashemi, line 343: Do we know if we want to show None or - for empty fields? Can we check with Terry / his designs?

If there is no decision atm I'm happy for you to go with whatever you prefer

anthonyhashemi (approved):
Nice fix. Can we just double check if we have decided to show None or - for non existent fields

app/tests/test_browse.py
pgandhizaizi: thank you

app/templates/main/browse.html
pgandhizaizi: I have kept the output consistent for all fields to show 'None' for now, There is a new ticket [AYR-608 check consignment ref view- without file meta data] has been created to discuss with Terry and Andrew on the design

anthonyhashemi: I'd suggest not applying the styling on the status field if it is None, but not a big deal since we never expect that case

PR #126, opened 2024-01-08
Create AYRUser class and refactor business logic that depends on ayr group logic

anthonyhashemi:
## JIRA ticket

Another prep PR for https://national-archives.atlassian.net/browse/AYR-574.

## Reason for this PR

This groups together all our keycloak user group logic into a new AYRUser class. This will help us maintain any changes to user group structure and assumptions we make between our keycloak implementation and the webapp.

In preparation of PRs for top level browse and search filtering by standard user's transferring body groups

## Changes in this PR
- Created AYRUser class
- Refactored business logic that depends on keycloak group logic

### DIFF
--- a/app/main/authorize/ayr_user.py
+++ b/app/main/authorize/ayr_user.py
@@ -1,0 +1,68 @@
+from typing import List
+
+from sqlalchemy import exc
+
+from app.main.authorize.keycloak_manager import (
+    get_user_transferring_body_keycloak_groups,
+)
+from app.main.db.models import Body, db
+
+
+class AYRUser:
+    def __init__(self, groups: List[str]) -> None:
+        self.groups = groups
+
+    @property
+    def can_access_ayr(self):
+        return self.is_superuser or self.is_standard_user
+
+    @property
+    def is_superuser(self) -> bool:
+        return "/ayr_user_type/view_all" in self.groups
+
+    @property
+    def is_standard_user(self) -> bool:
+        return "/ayr_user_type/view_dept" in self.groups
+
+    @property
+    def transferring_bodies(self) -> str | None:
+        if self.is_superuser or not self.is_standard_user:
+            return None
+
+        return get_user_accessible_transferring_bodies(self.groups)
+
+
+def get_user_accessible_transferring_bodies(groups: List[str]) -> List[str]:
+    user_transferring_bodies = get_user_transferring_body_keycloak_groups(
+        groups
+    )
+
+    if not user_transferring_bodies:
+        return []
+
+    try:
+        query = db.session.query(Body.Name)
+        bodies = db.session.execute(query)
+    except exc.SQLAlchemyError as e:
+        print("Failed to return results from database with error : " + str(e))
+        return []
+
+    user_accessible_transferring_bodies = []
+
+    for body in bodies:
+        body_name = body.Name
+        if _body_in_users_groups(body_name, user_transferring_bodies):
+            user_accessible_transferring_bodies.append(body_name)
+
+    return user_accessible_transferring_bodies
+
+
+def _body_in_users_groups(body, user_transferring_body_keycloak_groups):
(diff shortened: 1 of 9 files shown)


app/tests/conftest.py
```diff
@@ -9,14 +10,12 @@
 from configs.testing_config import TestingConfig
 
 
-def mock_standard_user(client: FlaskClient, body: str):
+def mock_standard_user(client: FlaskClient, bodies: List[str]):
```
anthonyhashemi, line 13: After chatting with @Terry-Price @pgandhizaizi @colinbowen this morning, we decided that for now we will assume that a standard user might have access to more than one transferring body, so have adjusted this to allow for it.

This is reflected in the AYRUser class also, where we have the `transferring_bodies` property, not `transferring_body`

app/main/authorize/ayr_user.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def transferring_bodies(self) -> str | None:
+        if self.is_superuser or not self.is_standard_user:
+            return None
+
+        return get_user_accessible_transferring_bodies(self.groups)
+
+
+def get_user_accessible_transferring_bodies(groups: List[str]) -> List[str]:
```
anthonyhashemi, line 33: Was thinking of refactoring this a bit more later to make it that the AYRUser doesnt have to know about the db but will leave for now.

app/main/authorize/ayr_user.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+
+    if not user_transferring_bodies:
+        return []
+
+    try:
+        query = db.session.query(Body.Name)
+        bodies = db.session.execute(query)
+    except exc.SQLAlchemyError as e:
+        print("Failed to return results from database with error : " + str(e))
```
colinbowen: Would it be better to use the app logger that was set up instead of this print?

app/main/authorize/ayr_user.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, groups: List[str]) -> None:
+        self.groups = groups
+
+    @property
+    def can_access_ayr(self):
+        return self.is_superuser or self.is_standard_user
+
+    @property
+    def is_superuser(self) -> bool:
+        return "/ayr_user_type/view_all" in self.groups
```
colinbowen, line 19: Do we have the user types defined anywhere? Open to discussing this a bit more but might be good to define all the available user types if we know what the possible options are

app/main/authorize/ayr_user.py
anthonyhashemi: yeah great point. This was just moved wholesale from another file in the refactor so I hadnt really inspected it.

will change to use logger 👍

colinbowen (approved):

app/main/authorize/ayr_user.py
anthonyhashemi: As discussed offline, these values are only used here and so I felt this sufficient as opposed to creating a constant that only gets used in one location.

This AYRUser class should then be the source of truth for all user groups logic.

Based off the decision in ADR 9 linked in slack.

app/main/authorize/ayr_user.py
anthonyhashemi: In fact I will remove this try except as if there is a db connection error then I think the user should see a internal server as that is completely unexpected behaviour.

colinbowen (approved):

pgandhizaizi:
"TNA User is logged in but does not have access to AYR. Please contact your admin." in init.py , I cannot see any unit testing performed on this

anthonyhashemi:
> "TNA User is logged in but does not have access to AYR. Please contact your admin." in init.py , I cannot see any unit testing performed on this

As discussed offline, this line was not changed and the function that it shows in was just refactored and so no tests surrounding it in this PR

PR #130, opened 2024-01-09
Fix mock user fixtures

anthonyhashemi:
We noticed that there were some tests that were showing up as PASSING when running together in CI but which were in fact broken when run individually.

After investigating we noticed this was due to bugs in the helper functions `mock_standard_user` and `mock_superuser`.

## Changes in this PR

- Changed `mock_standard_user` and `mock_superuser` to fixtures and fixed them so when running mutliple tests that use them at once, the permissions mocking doesn't persist across tests unintentionally.
- Fixed the tests that were showing up as PASSING when running together in CI before fixing the above but which were in fact broken when run individually
- Added info about `mock_standard_user` and `mock_superuser` to README

## JIRA ticket

### DIFF
--- a/app/tests/conftest.py
+++ b/app/tests/conftest.py
@@ -10,25 +10,39 @@ from app.main.db.models import db
 from configs.testing_config import TestingConfig
 
 
-def mock_standard_user(client: FlaskClient, bodies: List[str]):
-    with client.session_transaction() as session:
-        session["access_token"] = "valid_token"
-
-    groups = [f"/transferring_body_user/{body}" for body in bodies]
-    groups.append("/ayr_user_type/view_dept")
-
+@pytest.fixture(scope="function")
+def mock_standard_user():
     patcher = patch("app.main.authorize.permissions_helpers.get_user_groups")
     mock_get_user_groups = patcher.start()
-    mock_get_user_groups.return_value = groups
+
+    def _mock_standard_user(client: FlaskClient, bodies: List[str]):
+        with client.session_transaction() as session:
+            session["access_token"] = "valid_token"
+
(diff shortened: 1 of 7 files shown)


app/tests/conftest.py
```diff
@@ -10,25 +10,39 @@
...
+        groups.append("/ayr_user_type/view_dept")
+
+        mock_get_user_groups.return_value = groups
+
+    yield _mock_standard_user
 
-def mock_superuser(client: FlaskClient):
-    with client.session_transaction() as session:
-        session["access_token"] = "valid_token"
+    patcher.stop()
```
anthonyhashemi, line 29: Previously, we weren't stopping the patcher at the end of each test before so if multiple tests run in a seingle pytest run, where the first used this mocker function, but the second didn't, then the second test would have the mocked permissions from the first anyway.

rahmedzaizi (commented):
LGTM

app/tests/conftest.py
```diff
@@ -10,25 +10,39 @@
...
+    def _mock_standard_user(client: FlaskClient, bodies: List[str]):
+        with client.session_transaction() as session:
+            session["access_token"] = "valid_token"
+
+        groups = [f"/transferring_body_user/{body}" for body in bodies]
+        groups.append("/ayr_user_type/view_dept")
+
+        mock_get_user_groups.return_value = groups
+
+    yield _mock_standard_user
```
anthonyhashemi, line 27: Anything that comes after a yield ina pytest fixture is treated as teardown, i.e gets run when the fixture is being destroyed. In this casxe, since the scope is function, patcher.stop will get called at the end of each test

rahmedzaizi (approved):

PR #131, opened 2024-01-10
Ayr 606/consignment view filter

pgandhizaizi:
## Changes in this PR

1. queires.py --> modified browse function 

- added browse_type and filters parameters, added logic for filters in _build_consignment_filter_query to support filters
- added sorting_orders parameter, to support sorting options on consignment view

2. test_queries.py --> updated existing unit test cases to support filters, added new test cases for filters, added a new test cases for sorting options

3. routes.py --> modified browse route to differentiate between browse parameters through query string and filters supplied in a specific view

- added sample filters to use for consignment view
- added sample sorting options to use for consignment view

## JIRA ticket
AYR - 606 - consignment view filter
AYR - 601 - consignment sorting options

### After


app/main/db/queries.py
```diff
@@ -19,43 +19,44 @@ def fuzzy_search(query: str, page: int, per_page: int):
...
+    filters=None,
+    sorting_orders=None,
     transferring_body_id=None,
     series_id=None,
     consignment_id=None,
-    date_range=None,
-    date_filter_field=None,
 ):
-    if transferring_body_id:
+    if browse_type == "transferring_body":
```
anthonyhashemi, line 29: Nice change.

Just rehighlighting the point made in slack yesterday about the urls. I think in the next sprint we can start to split out the browse views into their own distinct routes,e.g.

```
@route("/browse")
def browse():
```


```
@route("/browse/consignment/<uuid>:consignment_id")
def browse_consignment(consignment_id):
```

```
@route("/browse/series/<uuid>:series_id")
def browse_consignment(series_id):
```

```
@route("/browse/transferring_body/<uuid>:transferring_body_id")
def browse_transferring_body(transferring_body_id):
```

so we have 
```
/browse/
/browse/transferring_body/id
/browse/consignment/id
/browse/series/id
```


or even 

```
/browse/
/transferring_body/id/browse
/consignment/id/browse
/series/id/browse
```

could even be better 

but yeah, should be easy to break this function out into each of those when we create those tickets.

app/main/db/queries.py
```diff
@@ -277,16 +280,68 @@ def _build_consignment_filter_query(
...
     )
 
-    if date_range and date_filter_field is not None:
-        dt_range = validate_date_range(date_range)
-        if date_filter_field.lower() == "date_last_modified":
-            date_filter = _build_date_range_filter(
-                sub_query.c.date_last_modified,
-                dt_range["date_from"],
-                dt_range["date_to"],
+    if filters:
```
anthonyhashemi, line 283: loving the organisation 👌

app/main/db/queries.py
pgandhizaizi: new user story AYR-620 has been created in a backlog

app/main/db/queries.py
pgandhizaizi: thank you for positive feedback

app/tests/test_queries.py
```diff
@@ -327,7 +327,212 @@ def test_browse_data_get_specific_page_results(self, client: FlaskClient):
...
+        Given multiple File objects in the database and consignment transfer complete date
+            that matches all of them
+        When browse_data is called with date_from and date_to
+        Then it returns a list containing multiple dictionary for the matching record with
+            date in between the range
+        """
+        files = create_multiple_test_records()
+
+        filters = {
+            "date_range": {"date_from": "01/02/2023", "date_to": "28/02/2023"}
```
anthonyhashemi, line 463: it would be ideal to change the date range or the setup data so that the objects returned here make it clear that the from and to filters are working in conjunction. Currently this results in the same assertion as in `def test_browse_with_date_from_filter(self, client: FlaskClient):`

app/main/db/queries.py
```diff
@@ -277,16 +280,68 @@ def _build_consignment_filter_query(
...
+                date_filter = _build_date_range_filter(
+                    sub_query.c.date_last_modified,
+                    dt_range["date_from"],
+                    dt_range["date_to"],
+                )
+                query = query.filter(date_filter)
+    return query
+
 
+def _build_consignment_sorting_orders(query, sub_query, sorting_orders):
```
anthonyhashemi, line 326: can refactor this to:

```

def _build_consignment_sorting_orders(query, sub_query, sorting_orders):
    for field, order in sorting_orders.items():
        column = getattr(sub_query.c, field, None)
        if column is not None:
            query = query.order_by(desc(column)) if order == 'desc' else query.order_by(column)
    return query
```

app/main/db/queries.py
```diff
@@ -277,16 +280,68 @@ def _build_consignment_filter_query(
...
-            query = query.filter(date_filter)
+    if "file_type" in filters:
+        file_type = filters["file_type"]
+        filter_value = str(f"%{file_type}").lower()  # "ppt"  # None  # "docx"
+        query = query.filter(
+            func.lower(sub_query.c.file_name).like(filter_value)
+        )
+    if "date_range" in filters:
+        date_range = filters["date_range"]
+        date_filter_field = None
```
anthonyhashemi: can condense 306-320 into

```
    date_range = filters.get("date_range")
        if date_range:
            date_filter_field = filters.get("date_filter_field")
            if date_filter_field and date_filter_field.lower() == "date_last_modified":
                dt_range = validate_date_range(date_range)
                date_filter = _build_date_range_filter(
                    sub_query.c.date_last_modified,
                    dt_range["date_from"],
                    dt_range["date_to"],
                )
                query = query.filter(date_filter)
```

app/main/db/queries.py
```diff
@@ -204,8 +205,10 @@ def _build_series_filter_query(series_id):
     return query
 
 
-def _build_consignment_filter_query(
-    consignment_id: uuid.UUID, date_range=None, date_filter_field=None
+def _build_consignment_view_query(
```
anthonyhashemi, line 208: I think it could be worth making a few changes to help us with testing maintainability and readability:

- make `_build_consignment_filters` a public function `build_consignment_filters` which we can then test all the edge case combination of filters in.

- make `_build_consignment_sorting_orders` a public function `build_consignment_sorting_orders` which we can then test all the edge case combination of sorting in.

- make `_build_consignment_view_query` a public function, `build_consignment_view_query` which we write 1 or 2 integration tests.

- then only need 1 or 2 top level browse_consignment tests which care about user permissions etc. but not too bothered by all the edge cases which will be dealt progressively at lower levels.

This will help us keep more in line with the test pyramid, where there are more unit tests and fewer, more complex integration tests; in particular this will help since most of the new tests we are adding to the browse_consignment level have a huge load of test setup to cover all the edge cases and are harder to debug.

app/tests/test_queries.py
pgandhizaizi: changes will be made to the test in AYR-604 branch

app/main/db/queries.py
anthonyhashemi: spoken on slack, will create a refactor  ticket (or 2) for this.

anthonyhashemi (approved):
Nice 👍

PR #133, opened 2024-01-12
Ayr 604/browse view combine filter

pgandhizaizi:
## Changes in this PR
1. added new pytest fixture in conftest.py to support browse filter and sorting testing
2. modified browse_data function in queries.py , added a new parameter (filters) to support filters operation , added new function _build_browse_filters to perform filter operation
3. added sample usage of filters in browse route in routes.py 
4. re-organised test cases in individual classes for browse type and added new test cases for browse filters in test_queries.py

## JIRA ticket
AYR-604  - browse view filters on all, transferring body and series view

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -208,17 +208,19 @@ def _build_series_view_query(series_id):
 
 
 def _build_browse_filters(query, filters):
-    if "transferring_body" in filters:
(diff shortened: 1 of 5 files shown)


anthonyhashemi:
should we close this PR for now to make sure nobody accidentally reviews it as we still need to wait for clear requirements for it which will probably be in next sprint?

PR #134, opened 2024-01-12
Create AYRUser.from_access_token static method to concentrate calls to keycloak

anthonyhashemi:
## Changes in this PR

- Create AYRUser.from_access_token static method to concentrate calls to keycloak

There will be a followup PR which removes `FORCE_AUTHENTICATION_FOR_IN_TESTING` from test config where we will add mock_standard_user or mock_superuser to all route tests that require authentication, rather than skipping auth on any of theirs. This PR makes it so we only need to mock one thing rather than 2 in that followup and is a wortwhile refactor in itself to move towards concentrating all calls to keycloak to one place to help decouple the app.

## JIRA ticket


colinbowen (approved):

pgandhizaizi (commented):
looks good

PR #135, opened 2024-01-12
Removed FORCE_AUTHENTICATION_FOR_IN_TESTING

anthonyhashemi:
## Changes in this PR

- Removed FORCE_AUTHENTICATION_FOR_IN_TESTING use mock_standard_user where necessary, adjusting _mock_standard_user to have default bodies list so dont need to specify when we dont care about db details.

Reason for this is now we have  mock_standard_user and mock_superuser fixtures which should be used in all route tests, i.e those that utilise the flask `client` 

## JIRA ticket


pgandhizaizi (approved):
looks good, approved

PR #136, opened 2024-01-12
Users can only have one body associated

anthonyhashemi:
## Changes in this PR

- Adjusted AYRUser from having `transferring_bodies` to `transferring_body`, taking the first body if the user has multiple bodies specified. We will have a ticket addressing validity of a user where we can enforce that a user should only ever have one, but need to think about how we want to enforce that.

- Updated mock_standard_user fixture to reflect the above also.

- Updated tests appropriately

## JIRA ticket

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ We have 2 fixtures used for mocking a user and their permissions in our non end 
 
 which can be called like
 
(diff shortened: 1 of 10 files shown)


pgandhizaizi (approved):
looks good

colinbowen (approved):

PR #137, opened 2024-01-12
Redirect standard user's to transferring body browse page when trying…

anthonyhashemi:
## Changes in this PR

- Redirect standard user's to transferring body browse page when trying to access the browse page
- Create a browse view test with mock_superuser to show that the user is redirected to their transferring body browse page
- Update all normal browse view tests to use mock_superuser

## JIRA ticket


pgandhizaizi (approved):
looks good

PR #140, opened 2024-01-16
Refactor / fix models and factories

anthonyhashemi:
## Changes in this PR

- Refactor / fix models and factories.

The factories were creating two body objects due to the Consignment model having related series and body even though we get related body from the related series also!

As I was here, I renamed the relations to be clearer also.

## JIRA ticket


colinbowen (approved):
Good change

PR #141, opened 2024-01-16
Refactor fuzzy search query

anthonyhashemi:
## Changes in this PR

Refactor fuzzy search query:

- paginate in the search route, not the query building function
- added tests for fuzzy query edge cases

## JIRA ticket

### DIFF
--- a/app/tests/test_queries.py
+++ b/app/tests/test_queries.py
@@ -261,23 +261,6 @@ class TestFuzzySearch:
         assert results == []
 
 
-#     # def test_build_fuzzy_search_query_multiple_matches(self):
-#     #     # Test when the query string matches multiple fields
-#     #     # ...
-
-#     # def test_build_fuzzy_search_query_case_insensitivity(self):
-#     #     # Test when the query string is case-insensitive
-#     #     # ...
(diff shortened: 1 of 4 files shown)


app/tests/test_queries.py
```diff
@@ -1,210 +1,281 @@
...
-                "test_file9.txt",
-                files[8].consignment.series.body.BodyId,
-                files[8].consignment.series.SeriesId,
-            ),
-        ]
-        assert result.pages > 0
-        assert result.has_next is False
-        assert result.has_prev is True
+
+#     # def test_build_fuzzy_search_query_multiple_matches(self):
```
colinbowen: Possible to remove these if they're not needed?

colinbowen (approved):
PR looks good just the commented out tests which could be removed

app/tests/test_queries.py
anthonyhashemi: ah yeah thanks for spotting these, I had put them as placeholders. will remove for now to get unblocked, can add more later

pgandhizaizi (approved):
looks good

PR #142, opened 2024-01-16
Filter search route for standard users by their transferring body

anthonyhashemi:
## Changes in this PR

- Filter search route for standard users by their transferring body

## JIRA ticket

### Before
Screenshot 2024-01-16 at 15 42 45

### After

Screenshot 2024-01-16 at 15 42 25


colinbowen (approved):
lgtm

PR #145, opened 2024-01-17
browse view all (top level) sorting functionality

pgandhizaizi:
## Changes in this PR

1. added a new pytest fixture to create browse file records for testing in conftest.py, 
2. modified browse_data query to accommodate sorting orders, added a new function _build_browse_sorting_orders in queries.py, 
3. modified /browse route to put example sorting order details in routes.py, 
4. created new test cases for browse sorting in test_queries.py

## JIRA ticket

AYR-598

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -51,6 +51,12 @@ def browse_data(
                     dt_range["date_to"],
                 )
                 browse_query = browse_query.filter(date_filter)
+        if sorting_orders:
+            browse_query = _build_browse_sorting_orders(
+                browse_query, sorting_orders
+            )
+        else:
+            browse_query = browse_query.order_by(Body.Name, Series.Name)
 
     return browse_query.paginate(page=page, per_page=per_page)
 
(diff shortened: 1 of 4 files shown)


app/main/db/queries.py
```diff
@@ -151,8 +168,8 @@ def _build_transferring_body_view_query(transferring_body_id):
             func.count(func.distinct(File.FileId)).label("records_held"),
         )
         .join(Consignment, Consignment.ConsignmentId == File.ConsignmentId)
-        .join(Body, Body.BodyId == Consignment.BodyId)
         .join(Series, Series.SeriesId == Consignment.SeriesId)
+        .join(Body, Body.BodyId == Series.BodyId)
```
anthonyhashemi, line 172: thanks for fixing these joins

app/tests/test_queries.py
```diff
@@ -584,6 +584,462 @@ def test_browse_with_date_from_and_to_filter_no_result(
 
         assert result.items == []
 
+    def test_browse_with_transferring_body_sorting_a_to_z(
```
anthonyhashemi, line 587: Like we discussed before, these tests could possibly be simpler and as effective at a lower level by moving the execution of the query (using pagination) to the route itself. Then we could test all these sorting edge cases without needing to worry about pagination and requiring lots of test objects.

This can be addressed when we tackle the ticket to break out the body route into distinct routes

anthonyhashemi (approved):
Looks good to me 👍

app/main/db/queries.py
pgandhizaizi: you are welcome

app/tests/test_queries.py
pgandhizaizi: yes, I agree , we should address when we create distinct routes

PR #146, opened 2024-01-17
Fix browse redirection

anthonyhashemi:
## Changes in this PR

- Fix browse redirection by redirecting to correct url
Previously we were getting 429 due to infinite loop of redirection

Going to make a separate ticket to update browse and search e2e tests to handle multiple users, as for AYR-574 we haven't updated these. Currently we only can handle one user we specify in our env vars but now we need one which is a superuser and one which is a standard user.

## JIRA ticket


pgandhizaizi (approved):
looks good

PR #148, opened 2024-01-17
Change to table component, update tests

colinbowen:
## Changes in this PR

- Switch from using govuk-summary-list for the browse table to using govuk-table component. 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-588

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/static/src/scss/base.scss
+++ b/app/static/src/scss/base.scss
@@ -11,7 +11,7 @@
 }
 
 
-@media screen and (width <=768px) {
+@media screen and (width <=810px) {
     .sign-out {
         margin: 0.5rem 0;
         align-self: flex-end;
--- a/app/static/src/scss/browse.scss
+++ b/app/static/src/scss/browse.scss
@@ -1,6 +1,6 @@
 /* Default styles for desktop */
 @mixin on-mobile {
-  @media screen and (width <=768px) {
+  @media screen and (width <=810px) {
     @content;
   }
 }
@@ -18,7 +18,12 @@
 // DESKTOP STYLES
 
 
+
 .govuk {
+  &-grid-column-full__browse-details {
+    padding: 0;
+  }
+
   &-form-group {
     &__search-form {
       margin-bottom: 0;
@@ -309,7 +314,7 @@
 
 // MOBILE STYLES
 
-@media screen and (width <=768px) {
+@media screen and (width <=810px) {
   .desktop {
     display: none;
   }
--- a/app/static/src/scss/browse.scss
+++ b/app/static/src/scss/browse.scss
@@ -130,6 +130,16 @@
 
 
 .browse-grid {
+  &__table {
+    font-size: 1rem;
+    border-bottom: 2px solid black;
+
+    thead {
+      border-bottom: 2px solid black;
+    }
+  }
+
+
   &--two-thirds {
     padding: 0;
   }
@@ -431,6 +441,14 @@
   .mobile-filters-form {
     display: block;
     width: 100%;
+  }
+
+  .govuk-button__filters-form-apply-button {
+    width: 50%;
+  }
+
+  .govuk-button__sort-container-update-button {
+    width: 50%;
   }
 
   .sort-container__form {
--- a/app/templates/main/browse.html
+++ b/app/templates/main/browse.html
@@ -320,7 +320,13 @@
                                                     {{ record['closure_type'] }}
                                                 </strong>
                                             </td>
-                                            <td class="govuk-table__cell">{{ record["closure_start_date"] }}</td>
+                                            <td class="govuk-table__cell">
+                                                {% if record['closure_start_date'] %}
+                                                    {{ record['closure_start_date'] }}
+                                                {% else %}
+                                                    -
+                                                {% endif %}
+                                            </td>
                                             <td class="govuk-table__cell">
                                                 {% if record['closure_period'] %}
                                                     {{ record['closure_period'] }} years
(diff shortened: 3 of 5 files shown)


app/static/src/scss/base.scss
```diff
@@ -11,7 +11,7 @@
 }
 
 
-@media screen and (width <=768px) {
+@media screen and (width <=810px) {
```
anthonyhashemi, line 14: what informed the mobile threshold change?

app/static/src/scss/browse.scss
```diff
@@ -130,6 +135,16 @@
 
 
 .browse-grid {
+  &__table {
+    font-size: 1rem;
```
anthonyhashemi, line 139: how comes there isnt any equivalent summary list styles to remove?

app/tests/test_browse.py
```diff
@@ -568,19 +568,20 @@ def test_browse_consignment_filter_display_multiple_pages(
             "Closure start date",
             "Closure period",
         ],
-        ["-", "test_file6.txt", "closed", "-", "-"],
-        ["-", "test_file7.png", "closed", "-", "-"],
+        ["15/12/2023", "test_file6.txt", "closed", "05/11/2023", "-"],
```
anthonyhashemi, line 575: whats the equivalent change in the code that led to these dates displaying as non-null?

app/static/src/scss/base.scss
colinbowen: This should account for both tablet & mobile sizing. The summary-list is a bit more fluid when it comes to different widths

app/static/src/scss/browse.scss
colinbowen: There is browse__body which previously defined the 1rem sizing for text and was used on the summary list body text. This class name is still being used for other body text on the browse page so I've kept it in for now. Open to changing class names if we would like different naming.

app/templates/main/browse.html
```diff
@@ -193,168 +195,155 @@ <h1 class="govuk-heading-l browse__records-found__text">Records found {{ num_rec
...
+                                                {% else %}
+                                                    -
+                                                {% endif %}
+                                            </td>
+                                            <td class="govuk-table__cell">
                                                 {% if record['closure_period'] %}
                                                     {{ record['closure_period'] }} years
                                                 {% else %}
-                                                    None
+                                                    -
```
anthonyhashemi, line 336: I think there was an open question about whether we want to show None or -. Did this get resolved?

app/tests/mock_database.py
```diff
@@ -762,6 +762,33 @@ def create_multiple_test_records():
     db.session.add(file_12)
     db.session.commit()
 
+    file_12_metadata = FileMetadata(
```
anthonyhashemi: What are these new objects for?

I'd suggest for new tests / changes to tests, we stop expanding and eventually stop using this function and use factory boy to create nested objects in addition to separating our test cases so that we dont need to have a huge test setup which is hard to debug

app/tests/test_browse.py
```diff
@@ -513,18 +513,18 @@ def test_browse_consignment_with_missing_file(
...
-    for index, row in enumerate(rows):
-        values = row.find_all("dd")
-        assert [
-            value.text.replace("\n", " ").strip(" ") for value in values
-        ] == expected_results_table[index + 1]
+    for index, row in enumerate(rows[1:]):
+        values = [value.text.strip() for value in row.find_all("td")]
+        expected_values = expected_results_table[index + 1]
+
+        assert values == expected_values
```
anthonyhashemi: why did these asserts change structure, but not the one in `test_browse_consignment_filter_display_multiple_pages` for example

app/static/src/scss/browse.scss
anthonyhashemi: no thats fine, was just wondering as I had imagined that we were using some styling for summary list before, but if it was shared, that makes sense

app/static/src/scss/base.scss
anthonyhashemi: ah I see, thanks for clarifying

app/templates/main/browse.html
colinbowen: It should be an "-" instead of None / not available. We've got an updated ticket for the actual change of this work to be completed in & the closure period is due to be calculated see:

https://national-archives.atlassian.net/browse/AYR-608
https://national-archives.atlassian.net/browse/AYR-654

app/tests/mock_database.py
colinbowen: I added this in as file_12 was missing valid metadata but yes we should switch to use factory boy for this.

app/templates/main/browse.html
anthonyhashemi: nice one, cheers

app/tests/mock_database.py
anthonyhashemi: maybe a better question is which test(s) does this change impact?

app/tests/test_browse.py
colinbowen: There's now a <strong> which contains the data. It was just easier for me to understand the assertion. Assertions should be the same

app/tests/test_browse.py
colinbowen: Resolved over call. Flaky test

app/tests/test_browse.py
colinbowen: Revert the format of this test

anthonyhashemi (approved):
Nice one

PR #154, opened 2024-01-19
browse series view sorting

pgandhizaizi:
## Changes in this PR
added a sorting functionality for browse series view

1. update logic for browse series view sorting in queries.py, 
2. updated example usage in routes.py, 
3. added new test cases for sorting for browse series view in test_queries.py

## JIRA ticket

AYR-600

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -27,7 +27,9 @@ def browse_data(
     elif browse_type == "series":
         series = Series.query.get_or_404(series_id)
         validate_body_user_groups_or_404(series.body.Name)
-        browse_query = _build_series_view_query(series_id)
+        browse_query = _build_series_view_query(
+            series_id, filters, sorting_orders
+        )
     elif browse_type == "consignment":
         consignment = Consignment.query.get_or_404(consignment_id)
         validate_body_user_groups_or_404(consignment.series.body.Name)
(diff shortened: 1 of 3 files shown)


app/main/db/queries.py
```diff
@@ -181,17 +183,16 @@ def _build_transferring_body_view_query(transferring_body_id):
...
         db.session.query(
             Body.BodyId.label("transferring_body_id"),
             Body.Name.label("transferring_body"),
             Series.SeriesId.label("series_id"),
             Series.Name.label("series"),
-            func.to_char(
-                func.cast(func.max(Consignment.TransferCompleteDatetime), DATE),
-                current_app.config["DEFAULT_DATE_FORMAT"],
-            ).label("last_record_transferred"),
+            func.max(Consignment.TransferCompleteDatetime).label(
```
anthonyhashemi, line 193: without this change does the sorting not work as expected due to the format of the date?

app/main/routes.py
```diff
@@ -133,6 +133,11 @@ def browse():
     elif series_id:
         browse_type = "series"
         browse_parameters["series_id"] = series_id
+        # sorting_orders["last_record_transferred"] = "asc"  # oldest first
```
anthonyhashemi, line 136: so for series we just sort by last record transferred date and num records held. Cool

anthonyhashemi (approved):

app/main/routes.py
pgandhizaizi: yes, that is was requested as per design by Terry

app/main/db/queries.py
pgandhizaizi: Yes, with DD/MM/YYYY format does not allow correct sorting as postgres DB support yyyy-mm-dd as a default format

PR #155, opened 2024-01-22
Swap parameter store out for secrets manager

anthonyhashemi:
We decided to swap out storing our config values in parameter store for secrets manager. In particular we are going with the pattern of a secrets manager key value pair, where the value is a JSON blob containing key value pairs of all the config values we currently store in parameter store.

## Changes in this PR

- Swapped out AWSParameterStoreConfig for AWSSecretsManagerConfig

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-594

Have tested this works on sandbox


configs/aws_secrets_manager_config.py
```diff
@@ -0,0 +1,49 @@
...
+        Get string value of `secret_name` in Secrets Manager.
+        :param key: Name of key whose value will be returned.
+        :return: String value of requested Parameter Store key.
+        """
+        client = boto3.client(
+            service_name="secretsmanager",
+        )
+
+        secret_value_json_string = client.get_secret_value(
+            SecretId="ayr-test-one-vars"  # pragma: allowlist secret
```
rahmedzaizi, line 31: we should parameterise env name (os.getenv)
get value from GitHub Action envVariables

lambda_policy.json
```diff
@@ -39,9 +39,9 @@
         {
             "Effect": "Allow",
             "Action": [
-                "ssm:GetParameter"
+                "s3:GetObject"
```
rahmedzaizi, line 42: Does this update the lambda_execution_role policy ?

anthonyhashemi: yeah, it is the policy itself but will remove once we sort our deplyoment out so this repo can be deployment agnostic

configs/aws_secrets_manager_config.py
anthonyhashemi: will leave this for followup ticket as we will need to deal with the terraform repo and I want to get the non prod deploiyment ticket unblocked for now

anthonyhashemi: https://national-archives.atlassian.net/browse/AYR-669

lambda_policy.json
anthonyhashemi: https://national-archives.atlassian.net/browse/AYR-668

rahmedzaizi (approved):

configs/aws_secrets_manager_config.py
anthonyhashemi: https://national-archives.atlassian.net/browse/AYR-668

PR #156, opened 2024-01-22
browse transferring body view sorting

pgandhizaizi:
## Changes in this PR
1. updated logic for browse series view sorting in queries.py,
2. updated example usage in routes.py,
3. added new test cases for sorting for browse series view in test_queries.py , 
4. added a new pytest fixture in conftest.py

## JIRA ticket

AYR- 599

### After


anthonyhashemi (approved):
Nice

PR #158, opened 2024-01-22
browse view all, transferring body and series view filters

pgandhizaizi:
## Changes in this PR

1.  added filter logic in queries.py, 
2. updated example filter options in routes.py, 
3. added new test cases for browse all filters, transfer_body filters and series filters in test_queries.py

## JIRA ticket

AYR-604

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -135,15 +135,7 @@ def _build_browse_everything_query(filters, sorting_orders):
     )
 
     if filters:
-        if "date_range" in filters:
-            dt_range = validate_date_range(filters["date_range"])
-
-            date_filter = _build_date_range_filter(
-                sub_query.c.last_record_transferred,
-                dt_range["date_from"],
-                dt_range["date_to"],
-            )
-            query = query.filter(date_filter)
+        query = _build_browse_filters(query, sub_query, filters)
 
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
@@ -195,6 +187,9 @@ def _build_transferring_body_view_query(
         sub_query.c.records_held,
     )
 
+    if filters:
+        query = _build_browse_filters(query, sub_query, filters)
+
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
     else:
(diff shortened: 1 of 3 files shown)


app/main/db/queries.py
```diff
@@ -243,12 +238,43 @@ def _build_series_view_query(series_id, filters, sorting_orders):
...
+        query = _build_browse_filters(query, sub_query, filters)
+
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
     else:
         query = query.order_by(
-            sub_query.c.transferring_body, sub_query.c.series
+            sub_query.c.transferring_body,
+            sub_query.c.series,
+            sub_query.c.consignment_reference,
```
anthonyhashemi, line 250: This is just an extra bit of ordering you added whilst working on this area, as opposed to anything to do with filtering, right?

app/main/db/queries.py
```diff
@@ -243,12 +238,43 @@ def _build_series_view_query(series_id, filters, sorting_orders):
...
+        )
+    print(query)
```
anthonyhashemi, line 252: can we remove this print statment

app/main/db/queries.py
```diff
@@ -243,12 +238,43 @@ def _build_series_view_query(series_id, filters, sorting_orders):
...
+    return query
+
+
+def _build_browse_filters(query, sub_query, filters):
```
anthonyhashemi, line 256: nice abstraction

app/main/db/queries.py
pgandhizaizi: This sorting was there before in the original query, this will make sure that if no sorting option selected, at least the data is sorted with body, series and consignment ref when the look at it

app/tests/test_queries.py
```diff
@@ -262,348 +261,20 @@ def test_build_fuzzy_search_query_no_results(self, client: FlaskClient):
...
-        create_multiple_test_records()
-
-        filters = {
-            "date_range": {"date_from": "01/03/2023", "date_to": "31/05/2023"}
-        }
-        result = browse_data(page=1, per_page=per_page, filters=filters)
-
-        assert result.items == []
-
-    def test_browse_with_transferring_body_sorting_a_to_z(
```
anthonyhashemi, line 587: github's PR diff of the changes ended up being poor due to the similar tests :( so looks like there's lots of new tests which are in fact these tests incorrectly marked as changed.

app/main/db/queries.py
pgandhizaizi: thank you

app/main/db/queries.py
anthonyhashemi: I don't see it in original file? It looks like it has just been added. But sorting by consignment ref after body and series seems sensible as a default

app/tests/test_queries.py
```diff
@@ -1539,44 +2192,220 @@ def test_browse_transferring_body_with_records_held_sorting_most_first(
...
+                browse_files[3].consignment.ConsignmentId,
+                browse_files[3].consignment.ConsignmentReference,
+            ),
         ]
 
+        results = pagination_object.items
+
+        assert results == expected_results
+
+    def test_browse_series_with_date_consignment_transferred_filter_using_date_from_and_date_to_option_no_result(
```
anthonyhashemi, line 2381: This file is even huuger than before. We should start breaking out test_queries.py into one test file per query function at minimum

anthonyhashemi: Like mentioned before, we would be able to reduce the amount of tests and cover more permutations more efficiently by testing various edge cases at a lower unit test level and then having one or two integration tests per route route / query and a function composition test to show that queries are built up by main query, filter and sort. 
We have a ticket for that but just wanted to raise it again so we keep it in mind

app/tests/test_queries.py
anthonyhashemi: I have ended up skimming over the tests and they generally look okay but I wouldnt fully trust it, because there were a lot of lines changed and so I may have missed some things

app/tests/test_queries.py
pgandhizaizi: I agree, need to split into four views test cases separately,

app/tests/test_queries.py
pgandhizaizi: without filter and with filter are new test cases added for this ticket, rest of sorting ones were done with sorting tickets

anthonyhashemi (approved):

app/main/db/queries.py
pgandhizaizi: as discussed it make sense to put consignment reference sorting as default

PR #163, opened 2024-01-25
Fix static content in lambda

anthonyhashemi:
## Changes in this PR

- Stopped creating bundled and minified css and js files at runtime
- Include all scss files in one main.scss file so we only need to include one file
- Fixed imports so that we import main.css that is created after running npm run build as per readme and also the cookies javascript is imported directly as opposed to the old minified version from flaskassets,
- Removed now redundant flask-assets package
- Updated readme appropriately

**Note:**
This shouldnt change what you need to do when developing, will still need to run npm run build / dev but will just have to make sure to adhere to the common pattern for including scss files in one main scss file of naming all scss files with_ so they are recognised as a partial stylesheet and then include them in main.scss e.g. for
`_accessibility.scss` -> add `@import "includes/accessibility";` to main.scss

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-676


README.md
```diff
@@ -545,16 +555,6 @@ With pytest we can assert the logs we expect to be written by utilising pytest's
...
-
-CSS is [minified](https://en.wikipedia.org/wiki/Minification_(programming)) using [CSSMin](https://github.com/zacharyvoase/cssmin) and JavaScript is minified using [JSMin](https://github.com/tikitu/jsmin/). This removes all whitespace characters, comments and line breaks to reduce the size of the source code, making its transmission over a network more efficient.
-
-### Cache busting
-
-Merged and compressed assets are browser cache busted on update by modifying their URL with their MD5 hash using [Flask Assets](https://flask-assets.readthedocs.io/en/latest/) and [Webassets](https://webassets.readthedocs.io/en/latest/). The MD5 hash is appended to the file name, for example `custom-d41d8cd9.css` instead of a query string, to support certain older browsers and proxies that ignore the querystring in their caching behaviour.
```
anthonyhashemi, line 556: I know we no longer have this as removed flask assets and currently we just build main.css file from main.scss but this shouldnt affect us in our deployment due to running in lambda. 

We may even move our assets to cloudfront anyway so we'd be able to deal with that there if we decide to. @colinbowen thoughts?

colinbowen (approved):
Good, reviewed during huddle to discuss changes.

README.md
colinbowen: Yeah definitely we should look to having at least the error pages set up to be retrieved from CloudFront. Worth having a discussion with Ben / Rahat too

pgandhizaizi (approved):
looks good

README.md
pgandhizaizi: I agree, that we should put static content in Cloudfront, having it cached will improve the performance too.

PR #164, opened 2024-01-25
browse view bread crumbs fix

pgandhizaizi:
## Changes in this PR

1. modified templates to add breadcrumb details for mobile and desktop view, 
2. modified consignment query in queries.py to include additional columns for breadcrumbs, 
3. amended test cases in test_browse.py to check breadcrumb asserts are working as expected, 
4. updated consignment test cases to use pytest fixture in test_queries.py

## JIRA ticket

AYR-595

### Before

transferring body view not showing breadcrumb from everything

image

series view not showing breadcrumbs for everything and transferring body

image

consignment view not showing breadcrumbs for everything, transferring body and series
image

### After

transferring body view showing breadcrumb for everything view with a link

image

series view showing breadcrumbs for everything and transferring body with a link

image

consignment views showing breadcrumbs for everything, transferring body and series with a link

image

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -249,7 +249,7 @@ def _build_series_view_query(series_id, filters, sorting_orders):
             sub_query.c.series,
             sub_query.c.consignment_reference,
         )
-    print(query)
+
     return query
 
 
@@ -323,6 +323,12 @@ def _build_consignment_view_query(
                 else_=None,
             ),
         ).label("closure_period"),
+        Consignment.ConsignmentId.label("consignment_id"),
+        Consignment.ConsignmentReference.label("consignment_reference"),
+        Body.Name.label("transferring_body"),
+        Body.BodyId.label("transferring_body_id"),
+        Series.SeriesId.label("series_id"),
+        Series.Name.label("series"),
     )
 
     query_filters = [
@@ -334,8 +340,19 @@ def _build_consignment_view_query(
         select.join(
             FileMetadata, File.FileId == FileMetadata.FileId, isouter=True
         )
+        .join(Consignment, File.ConsignmentId == Consignment.ConsignmentId)
+        .join(
+            Series,
+            Series.SeriesId == Consignment.SeriesId,
+        )
+        .join(
+            Body,
+            Body.BodyId == Series.BodyId,
+        )
         .filter(*query_filters)
-        .group_by(File.FileId)
+        .group_by(
+            File.FileId, Body.BodyId, Series.SeriesId, Consignment.ConsignmentId
+        )
         .order_by(File.FileName)
     ).subquery()
 
@@ -352,13 +369,20 @@ def _build_consignment_view_query(
             current_app.config["DEFAULT_DATE_FORMAT"],
         ).label("closure_start_date"),
         sub_query.c.closure_period,
+        sub_query.c.transferring_body_id,
+        sub_query.c.transferring_body,
+        sub_query.c.series_id,
+        sub_query.c.series,
+        sub_query.c.consignment_id,
+        sub_query.c.consignment_reference,
     )
 
     if filters:
         query = _build_consignment_filters(query, sub_query, filters)
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
-
+    else:
+        query = query.order_by(sub_query.c.file_name)
     return query
 
 
--- a/app/templates/main/breadcrumb.html
+++ b/app/templates/main/breadcrumb.html
@@ -1,0 +1,30 @@
+{% set browse_url = "/browse" %}
+{% set transferring_body_url = browse_url ~ "?transferring_body_id=" %}
+{% set series_url = browse_url ~ "?series_id=" %}
+{% set consignment_url = browse_url ~ "?consignment_id=" %}
+<div class="govuk-breadcrumbs">
+    <ol class="govuk-breadcrumbs__list">
+        {% for key, value in dict.items() %}
+            {% if loop.index != dict|length %}
+                <li class="govuk-breadcrumbs__list-item">
+                    {% if key == "everything" %}
+                        <a class="govuk-breadcrumbs__link--record" href="{{ browse_url }}">{{ value }}</a>
+                    {% elif key == "body" %}
+                        <a class="govuk-breadcrumbs__link--record"
+                           href="{{ transferring_body_url ~ value[0] }}">{{ value[1] }}</a>
+                    {% elif key == "series" %}
+                        <a class="govuk-breadcrumbs__link--record"
+                           href="{{ series_url ~ value[0] }}">{{ value[1] }}</a>
+                    {% elif key == "consignment" %}
+                        <a class="govuk-breadcrumbs__link--record"
+                           href="{{ consignment_url ~ value[0] }}">{{ value[1] }}</a>
+                    {% endif %}
+                </li>
+            {% else %}
+                <li class="govuk-breadcrumbs__list-item">
+                    <p class="govuk-breadcrumbs__link--record">{{ value }}</p>
+                </li>
+            {% endif %}
+        {% endfor %}
+    </ol>
+</div>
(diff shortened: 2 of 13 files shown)


app/templates/main/browse-series.html
```diff
@@ -334,7 +373,6 @@
             </div>
         </details>
     </div>
-    <!-- END MOBILE FILTERS -->
```
anthonyhashemi, line 337: do we want these block comments or not?

pgandhizaizi: I think the top block comment should suffice and they get differentiated with other block start comment

app/templates/main/browse-series.html
```diff
@@ -1,4 +1,25 @@
 {% if view_type == "desktop" %}
+    <!-- BREAD CRUMB -->
```
anthonyhashemi: we could make a breadcrumbs html file which takes in variable:

something like 
https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/main/ds_caselaw_editor_ui/templates/includes/document_breadcrumbs.html

but I think we could make it simpler by making it a for loop with the last element treated differently

e.g. for series
parent_structure_list = [{"Everything": everyting_url, {"Body A": body_a_url}]
end_node = "Series Name"

with this structure we could easily make the breadcrumbs a generic include.

Could even have it at the browse.html level and then just pass in `parent_structure_list`, `end_node` as parameters to the render function

app/main/db/queries.py
```diff
@@ -323,6 +323,12 @@ def _build_consignment_view_query(
                 else_=None,
             ),
         ).label("closure_period"),
+        Consignment.ConsignmentId.label("consignment_id"),
+        Consignment.ConsignmentReference.label("consignment_reference"),
```
anthonyhashemi, line 327: do we need all of these fields?

pgandhizaizi: yes to support bread crumb on consignment view, keep the consistent approach to columns across all queries

app/templates/main/browse-series.html
pgandhizaizi: separated breadcrumb to a new template and updated all browse templates (incl. record template) to use that template

app/main/db/queries.py
anthonyhashemi: I thought this query was intended for the rows in the table. I think its fine for now but might be worth us revisting all these queries if we have time in a couple sprints time

app/templates/main/browse-all.html
```diff
@@ -51,145 +55,135 @@
...
-            <div class="govuk-form-group filters-form__group filters-form__series-group">
-                <label class="govuk-label" for="browse_filter_series"></label>
-                <input class="govuk-input filters-form__series--input"
-                       id="browse_filter_series"
-                       name="width10"
-                       type="text">
+        <form method="post" action="{{ url_for('main.browse') }}">
+            {{ form.csrf_token }}
+            <div class="browse-filter-container">
+                <div class="browse-filter__header">
```
anthonyhashemi: whats the change here about form?

app/templates/main/browse.html
```diff
@@ -29,37 +29,36 @@
             </div>
         </div>
         {% if results %}
-            <div class="govuk-grid-column-full govuk-grid-column-full__browse-details">
```
anthonyhashemi, line 32: why has `<div class="govuk-grid-column-full govuk-grid-column-full__browse-details">` been moved inside of the div starting after line ?

app/templates/main/breadcrumb.html
```diff
@@ -0,0 +1,30 @@
+{% set browse_url = "/browse" %}
```
anthonyhashemi, line 1: this abstraction is really clean 👌

app/templates/main/browse-all.html
pgandhizaizi: I think I was working on the implementation of browse filters , that by mistake checked in , please let me know if you want me to remove

pgandhizaizi: removed the changes of  <form> tag not required  in browse_all.html page

app/templates/main/browse.html
pgandhizaizi: removed extra div not required in browse.html for both desktop and mobile view

app/tests/test_browse.py
```diff
@@ -553,7 +562,9 @@ def test_browse_consignment_filter_display_multiple_pages(
     assert response.status_code == 200
     assert b"Search for digital records" in response.data
     assert b"You are viewing" in response.data
-    assert b"Everything available to you" in response.data
+    assert b"test body1" in response.data
```
anthonyhashemi, line 704: could we make the assertions clearer that these build up breadcrumbs. Currently this just assrts that this data is foundf anywhere on the page.

Can we assert the full breadcrumb html element is found and is as expected?

Similarlyl for all new breadcrumb assertions in the other tests

app/main/db/queries.py
pgandhizaizi: the query is intended get file details, but you do need body, series and consignment ref details for breadcrumb in consignment view, due to that these fields added

app/tests/test_browse.py
pgandhizaizi: sure, I can add new separate test cases for breadcrumbs

anthonyhashemi (approved):
Looks good. I did notice the spacing between the breadcrumbs is a bit large but you said the styling for the breadcrumbs would addressed in a separate bug ticket so that can be dealt with there also

app/tests/test_browse.py
anthonyhashemi: I wasnt saying new test cases necessarily, just that these assertions of test bodt 1, testseries 1 etc dont show that they should be part of a breadcrum b structure. You can just replace these assertions you had added with one assertion per test changed like

assert b" < BREADCRUMB STRUCTURE>EVERYTHING,BODY,SERIES,ETC.</BREADCRUMB_STRUCTURE>" in response.data

app/main/db/queries.py
anthonyhashemi: Yeah that's understandable but it seems a bit hacky. It's fine for now but could be worth revisiting later if we have time

app/tests/test_browse.py
pgandhizaizi: new test cases added for breadcrumb html

app/main/db/queries.py
pgandhizaizi: I agree, we have to revisit all queries during refactoring the code

app/tests/test_browse.py
```diff
@@ -480,6 +581,56 @@ def test_browse_consignment(client: FlaskClient, mock_standard_user):
...
+    mock_standard_user(client, file.consignment.series.body.Name)
+
+    response = client.get(f"/browse?consignment_id={consignment_id}")
+
+    assert response.status_code == 200
+    assert b"Search for digital records" in response.data
+    assert b"You are viewing" in response.data
+
+    html = response.data.decode()
+    print(html)
```
anthonyhashemi: can we remove print

app/tests/test_browse.py
```diff
@@ -480,6 +581,56 @@ def test_browse_consignment(client: FlaskClient, mock_standard_user):
...
+                <a class="govuk-breadcrumbs__link--record"
+                    href="/browse?series_id={file.consignment.series.SeriesId}">{file.consignment.series.Name}</a>
+            </li>
+            <li class="govuk-breadcrumbs__list-item">
+                <p class="govuk-breadcrumbs__link--record">{file.consignment.ConsignmentReference}</p>
+            </li>
+        </ol>
+    </div>
+    """
+    print(expected_breadcrumbs_html)
```
anthonyhashemi: likewise this print

app/tests/test_browse.py
pgandhizaizi: removed print statements

app/tests/test_browse.py
pgandhizaizi: removed print

PR #167, opened 2024-01-29
Add revised header, fix margin bug

colinbowen:
## Changes in this PR
- Use revised header (Keycloak header)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-570

### Before
Screenshot 2024-01-29 at 16 34 47
Screenshot 2024-01-29 at 16 34 38

### After
Screenshot 2024-01-29 at 16 32 35

Screenshot 2024-01-29 at 16 32 25

### DIFF
--- a/app/static/src/scss/includes/_header.scss
+++ b/app/static/src/scss/includes/_header.scss
@@ -1,39 +1,40 @@
 @mixin on-mobile {
-    @media screen and (width <= 768px) {
+    @media screen and (width <=768px) {
         @content;
     }
 }
 
 .govuk-header {
-    &__logo {
+    font-family: Arial, sans-serif !important;
+
+    &__logo--ayr {
+
         @include on-mobile {
             padding: 0;
             margin: 0;
         }
     }
 
-    &__link--tna {
+    &__link--ayr {
         font-weight: 700;
     }
 
     &__logotype--ayr {
         white-space: nowrap;
-        word-spacing: 0.2rem;
         margin-top: 0;
         margin-bottom: 0;
 
         @include on-mobile {
-            font-size: 1.7rem;
+            font-size: 1.8rem;
+
         }
 
     }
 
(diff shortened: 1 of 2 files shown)


app/static/src/scss/header.scss
```diff
@@ -1,39 +1,40 @@
...
+    &__logo--ayr {
+
         @include on-mobile {
             padding: 0;
             margin: 0;
         }
     }
 
-    &__link--tna {
+    &__link--ayr {
```
anthonyhashemi: (for my understanding) was the --tna suffix just an oversight when it should have always been --ayr since that is the bem pattern we agreed for modifing the govuk styles?

app/static/src/scss/header.scss
```diff
@@ -1,39 +1,40 @@
...
 
     }
 
-    &__content--tna {
+    &__content--ayr {
         text-align: right;
         width: 100%;
-        padding-bottom: 0.5rem;
         margin-top: 0.75rem;
-        word-spacing: 0.2rem;
```
anthonyhashemi: I trust all these make sense as long as the page banner visually matches what was expected in the ticket

colinbowen: Yeah, went through these changes with Terry who's signed off on these changes. These new styles are the same as what we will have in keycloak. I'll add a screenshot.

app/static/src/scss/header.scss
colinbowen: I believe the tna suffix was set as it's for "Delivered by The National Archives". Happy to revert back but i've set the main header logo to be under logo--ayr & the sublink to be link--ayr. Both could be reasonable choices

anthonyhashemi: your changes makes sense to me 👍

anthonyhashemi (approved):
Nice 👌

PR #169, opened 2024-01-29
BROWSE view all (top level) filters and Integration with backend

pgandhizaizi:
## Changes in this PR

1. removed references of mock_database functions and replaced test cases with pytest fixture in test_browse.py, 
2. added new filter test cases in test_browse.py, 
3. added a new css _browse-all.scss for styling of browse all view, 
4. fixed record_status=all value comparison with lowercase in queries.py, 
5. updated logic to read filters from the query string in routes.py, 
6. added filter integration changes in browse-all.html, 
7. added a new property transferring_bodies to get a list of transferring bodies to all access user has access to, added new test cases in test_ayr_user.py

## JIRA ticket

https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-605 

### After


app/tests/test_queries.py
```diff
@@ -261,6 +262,26 @@ def test_build_fuzzy_search_query_no_results(self, client: FlaskClient):
 
 
 class TestBrowse:
+    def test_browse_get_all_transferring_bodies(
+        self, client, mock_superuser, browse_files
```
anthonyhashemi: Could do away with browse_files here, we don't need the 27 file objects, we just need to create a few Body objects.

```
        bodies = BodyFactory.create_batch(3)
        mock_superuser(client)
        body_names = get_all_transferring_bodies()

        assert body_names == [bodies[0].Name, bodies[1].Name, bodies[2].Name]
```

app/tests/test_queries.py
```diff
@@ -261,6 +262,26 @@ def test_build_fuzzy_search_query_no_results(self, client: FlaskClient):
...
+    ):
+        """
+        Given 27 file objects with all file type as 'file'
+        And the session contains user info for a superuser with access
+        When I call the 'get_all_transferring_bodies' function
+        Then it should return list of all transferring body names available in database
+        """
+        mock_superuser(client)
```
anthonyhashemi: This function doesn't depend on a user / session info, so can remove mock_superuser

app/main/routes.py
```diff
@@ -193,10 +198,49 @@ def browse():
...
+        filter_items = []
+        filters = {}
+        transferring_body = request.args.get(
+            "transferring_body_filter", ""
+        ).lower()
+        series = request.args.get("series_filter", "").lower()
+        date_from_day = request.args.get("date_from_day", "")
+        date_from_month = request.args.get("date_from_month", "")
+        date_from_year = request.args.get("date_from_year", "")
+        date_from = date_from_day + "/" + date_from_month + "/" + date_from_year
```
anthonyhashemi: I was a little confused about the "//" checks below, then realised they are accounting for the possibility of these string concatenations resulting in "//"

What if someone puts:
- day, month, BUT no year
- month, year BUT no day
- etc.

I think we should probably set defaults for the Day, month or year, depending on the other values specified already. I remember us doing something similar in caselaw... https://github.com/nationalarchives/ds-caselaw-public-ui/blob/7005a0ab1e52b6143951f83b3c2887e7a448434e/judgments/views/advanced_search.py#L32

It could be worth creating a `_build_date_range_filter()` to abstract out all this date logic into one place to.

app/tests/test_browse.py
```diff
@@ -30,715 +26,1594 @@ def test_standard_user_redirected_to_browse_transferring_body_when_accessing_bro
...
-    mock_superuser(client)
-    create_multiple_test_records()
-
-    query = "test"
-    response = client.post("/browse", data={"query": query})
-
-    assert response.status_code == 200
-    assert b"Search for digital records" in response.data
-    assert b"Records found 11" in response.data
+class TestBrowse:
```
anthonyhashemi, line 138: can we move this test class / all browse all tests to their own test file since this file is really big

app/tests/test_browse.py
```diff
@@ -30,715 +26,1594 @@ def test_standard_user_redirected_to_browse_transferring_body_when_accessing_bro
...
+            [
+                "'fifth_body', 'fifth_series', '21/09/2023', '6', '2', "
+                "'first_body', 'first_series', '07/02/2023', '3', '2', "
+                "'fourth_body', 'fourth_series', '03/08/2023', '6', '2', "
+                "'second_body', 'second_series', '26/04/2023', '7', '2', "
+                "'sixth_body', 'sixth_series', '14/10/2023', '2', '1'"
+            ],
+        ]
 
+        assert [
```
anthonyhashemi: can we create an assertion function or 2 functions for the lines between 140 and 149 since we repeat these in many tests

app/tests/test_browse.py
```diff
@@ -30,715 +26,1594 @@ def test_standard_user_redirected_to_browse_transferring_body_when_accessing_bro
...
+                "Consignments within series",
+            ],
+            [
+                "'fifth_body', 'fifth_series', '21/09/2023', '6', '2', "
+                "'first_body', 'first_series', '07/02/2023', '3', '2', "
+                "'fourth_body', 'fourth_series', '03/08/2023', '6', '2', "
+                "'second_body', 'second_series', '26/04/2023', '7', '2', "
+                "'sixth_body', 'sixth_series', '14/10/2023', '2', '1'"
+            ],
+        ]
```
anthonyhashemi, line 232: can we reduce a lot of this code by parametrizing the tests for the query params.

something along the lines of

```
@pytest.mark.parametrize(
        "query_params, ", expected_rows",
[
("date_from_day=1&date_from_month=2&date_from_year=2022&date_to_day=3&date_to_month=6&date_to_year=2023",
            [
                "'fifth_body', 'fifth_series', '21/09/2023', '6', '2', "
                "'first_body', 'first_series', '07/02/2023', '3', '2', "
                "'fourth_body', 'fourth_series', '03/08/2023', '6', '2', "
                "'second_body', 'second_series', '26/04/2023', '7', '2', "
                "'sixth_body', 'sixth_series', '14/10/2023', '2', '1'"
            ],
    ),
("series_filter=fi",
            [
                "'fifth_body', 'fifth_series', '21/09/2023', '6', '2', "
                "'first_body', 'first_series', '07/02/2023', '3', '2', "
                "'fourth_body', 'fourth_series', '03/08/2023', '6', '2', "
                "'second_body', 'second_series', '26/04/2023', '7', '2', "
                "'sixth_body', 'sixth_series', '14/10/2023', '2', '1'"
            ],
etc.
)
]
```

app/main/routes.py
pgandhizaizi: we should be checking the frontend for basic date validation and highlighting any error to the user to amend it when clicking on apply a filter , alternatively,
e.g. if the user enters only year as 2023 - the date from should default to 01/01/2023, the same way if the user enters the only year 2023 - the date to should default to 31/12/2023. and if not data entered for date to, it should default to the current date, this can be implemented, but we might need to spend more time putting the functionality - this is based on a discussion we had today with Terry

app/tests/test_browse.py
pgandhizaizi: I separated them with different classes for now, will pick up the split test cases in different files when splitting the browse routes

app/tests/test_queries.py
pgandhizaizi: fixed the test case as requested

app/tests/test_queries.py
pgandhizaizi: removed mock_superuser from test case

app/main/routes.py
pgandhizaizi: at the moment separated date filters in _build_date_range_filter , but we need to finalise and agree on what approach we want to take for date parameters

1. we can add validation on front end to check date filters have a valid date value using flaskform
2. we need to create separate ticket for validations

app/tests/test_browse.py
pgandhizaizi: separated in a function to verify header and data rows and updated all test cases accordingly.

app/tests/test_browse.py
pgandhizaizi: not quite clear on what will be an advantage having parameterised test fixture, we can look at to gather when you are back on Thursday

app/main/routes.py
pgandhizaizi: image

pgandhizaizi: we need to provide the validations according to Terry's design
https://national-archives.atlassian.net/wiki/spaces/DAYR/pages/204275723/Validation+messaging+matrix+for+AYR

app/main/routes.py
```diff
@@ -193,10 +198,55 @@ def browse():
         filters=browse_parameters,
         browse_type=browse_type,
         results=browse_results,
+        transferring_bodies=transferring_bodies,
+        user_filters=filters,
         num_records_found=num_records_found,
     )
 
 
+def _build_browse_all_filters():
```
colinbowen, line 207: We could move these to a utils file to keep the routes.py file strictly for routes

app/tests/test_browse.py
anthonyhashemi: ok

app/tests/test_browse.py
anthonyhashemi: it will save us having all these various tests which are the same apart from the query paramaters and expected rows

app/main/routes.py
pgandhizaizi: I moved the changes to a new util file for both filters and sorting logic in a new PR 177

anthonyhashemi (approved):
LGTM 👍

app/main/routes.py
anthonyhashemi: we agreed we will do date validation in a separate ticket. This ticket is just considering happy path

PR #173, opened 2024-01-31
browse all view sorting integration

pgandhizaizi:
## Changes in this PR

performed sorting for each sorting option provided in the dropdown on browse all view for all access user (superuser)

## JIRA ticket

AYR - 647

### After


anthonyhashemi (approved):
Nice one 👍

PR #174, opened 2024-01-31
browse transferring body filters integration

pgandhizaizi:
## Changes in this PR

performed filters integration from frontend to backend for transferring body view for standard user

## JIRA ticket

AYR - 662

### After

### DIFF
--- a/app/templates/main/browse-transferring-body.html
+++ b/app/templates/main/browse-transferring-body.html
@@ -9,22 +9,39 @@
 {% if view_type == "desktop" %}
     <!-- SORT -->
     <div class="govuk-form-group sort-container__form">
-        <div class="browse__sort-container">
-            <label class="govuk-label" for="sort">Sort by</label>
-            <select class="govuk-select govuk-select__sort-container-select"
-                    id="sort"
-                    name="sort">
-                <option value="body-a">Transferring body (A to Z)</option>
-                <option value="body-b" selected>Transferring body (Z to A)</option>
-                <option value="series-a">Series (A to Z)</option>
-                <option value="series-b">Series (Z to A)</option>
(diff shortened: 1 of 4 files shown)


anthonyhashemi:
This PR is dependent on https://github.com/nationalarchives/da-ayr-beta-webapp/pull/175 so worth pointing this PR to `AYR-665/browse-transferring-body-sorting-integration` rather than main until that one is merged in.

Also, seems like this PR is about sorting, not filtering as the name suggests. I think the 2 PRs were named the wrong way around

app/templates/main/browse-transferring-body.html
```diff
@@ -319,22 +336,39 @@ <h3 class="govuk-heading-s govuk-heading-s--browse-all-filter">Date consignment
...
-                <option value="series-a">Series (A to Z)</option>
-                <option value="series-b">Series (Z to A)</option>
-                <option value="date-first">Date record transferred (most recent first)</option>
-                <option value="date-last">Date record transferred (oldest first)</option>
-            </select>
-            <button class="govuk-button govuk-button__sort-container-update-button"
-                    id="sort-update-button"
-                    data-module="govuk-button">Apply</button>
-        </div>
+        <form action="{{ url_for('main.browse') }}" method="get">
```
anthonyhashemi, line 339: the mobile and desktop  forms and all nested html looks very similar. Are we able to extract the shared parts to reduce duplication due to the mobile view. Maybe Im missing something that differs though. @colinbowen thoughts?

anthonyhashemi: spoke on offline, going to be dealt with in a refactor PR

anthonyhashemi (approved):
LGTM

PR #175, opened 2024-01-31
browse transferring body sorting integration

pgandhizaizi:
## Changes in this PR
performed sorting on transferring body view for standard user

## JIRA ticket

AYR - 665

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -130,10 +130,10 @@ def browse():
         sorting_orders = build_sorting_orders()
 
     if transferring_body_id:
+        print("i was here")
         browse_type = "transferring_body"
         browse_parameters["transferring_body_id"] = transferring_body_id
-        # filters["series"] = "tsta1"
-        # filters["date_range"] = {"date_from": "01/08/2022", "date_to": "31/08/2022"}
+        filters = _build_browse_all_filters()
         # sorting_orders["series"] = "asc"  # A to Z
         # sorting_orders["series"] = "desc"  # Z to A
         # sorting_orders["last_record_transferred"] = "asc"  # oldest first
--- a/app/templates/main/browse-transferring-body.html
+++ b/app/templates/main/browse-transferring-body.html
@@ -57,141 +57,130 @@
         {% endif %}
     </div>
     <!-- FILTERS -->
-    <div class="govuk-grid-column-one-third filters-form">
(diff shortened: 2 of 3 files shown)


app/main/routes.py
```diff
@@ -130,10 +130,10 @@ def browse():
         sorting_orders = build_sorting_orders()
 
     if transferring_body_id:
+        print("i was here")
         browse_type = "transferring_body"
         browse_parameters["transferring_body_id"] = transferring_body_id
-        # filters["series"] = "tsta1"
-        # filters["date_range"] = {"date_from": "01/08/2022", "date_to": "31/08/2022"}
+        filters = _build_browse_all_filters()
```
anthonyhashemi, line 136: Should we rename `_build_browse_all_filters` since it is used here as well and will be used in browse series too?

app/templates/main/browse-transferring-body.html
```diff
@@ -57,286 +57,264 @@
...
-                                <input class="govuk-input govuk-date-input__input govuk-input--width-2"
-                                       id="date-to-day"
-                                       name="date-to-day"
-                                       type="text"
-                                       inputmode="numeric">
-                                <label class="govuk-label govuk-date-input__label filters-form__date__label"
-                                       for="date-to-day">(DD)</label>
+                    </fieldset>
+                </div>
+                <div class="browse-all-filter-to-dates">
```
anthonyhashemi, line 131: can we extract the datae range component into its own html to make this easier to read and reduce duplication?

Like wise with all other shared filter components, e.g. series filter, and apply button div etc.

anthonyhashemi:
seems like this PR is about filtering, not sorting like the name suggests?

app/main/routes.py
pgandhizaizi: this has been renamed to moved to separate file in PR 177

app/templates/main/browse-transferring-body.html
pgandhizaizi: I was thinking of separating both sort and filter into a component which can work for both mobile and desktop view

app/main/routes.py
anthonyhashemi: yeah  just saw that now 👍

anthonyhashemi (approved):
LGTM

PR #176, opened 2024-01-31
browse series filters integration

pgandhizaizi:
## Changes in this PR

perform filter operation on browse series view as a standard user

## JIRA ticket

AYR - 663

### After

### DIFF
--- a/app/templates/main/browse-series.html
+++ b/app/templates/main/browse-series.html
@@ -58,141 +58,119 @@
         {% endif %}
     </div>
     <!-- FILTERS -->
-    <div class="govuk-grid-column-one-third filters-form">
-        <div class="browse-filter-container">
-            <div class="browse-filter__header">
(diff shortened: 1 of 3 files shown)


app/templates/main/browse-series.html
```diff
@@ -58,286 +58,240 @@
...
+                </div>
             </div>
-            <div class="filters-form__buttons">
-                <button type="button"
-                        class="govuk-button govuk-button__filters-form-apply-button"
-                        data-module="govuk-button">Apply filters</button>
-                <button type="button"
-                        class="govuk-button__filters-form-clear-button"
-                        data-module="govuk-button">Clear all filters</button>
+            <div class="browse-all-filter-container">
```
anthonyhashemi, line 165: can we rename the browse-all classes since they are not unique to the browse all view

anthonyhashemi (commented):
Nice one. Will approve once all the dependent prs merged in

PR #177, opened 2024-02-02
browse series sorting integration

pgandhizaizi:
## Changes in this PR
perform sorting on the browse series view, extracted filter and sorting logic in separate util file, updated browse view and browse transferring body with new sorting logic

## JIRA ticket
AYR - 666

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -247,7 +247,7 @@ def _build_series_view_query(series_id, filters, sorting_orders):
         query = query.order_by(
             sub_query.c.transferring_body,
             sub_query.c.series,
-            sub_query.c.consignment_reference,
+            desc(sub_query.c.last_record_transferred),
         )
 
     return query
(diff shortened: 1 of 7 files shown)


app/main/db/queries.py
```diff
@@ -247,7 +247,7 @@ def _build_series_view_query(series_id, filters, sorting_orders):
         query = query.order_by(
             sub_query.c.transferring_body,
             sub_query.c.series,
-            sub_query.c.consignment_reference,
+            desc(sub_query.c.last_record_transferred),
```
anthonyhashemi, line 250: what prompted the change from default sorting of consignment reference asc to last_rec_transf desc?

pgandhizaizi: consignment_reference sort was set earlier in a query, as the consignment view now requires default sort by record transferred, so amended the query accordingly

anthonyhashemi (approved):

PR #179, opened 2024-02-02
Integrate sort options, set default sort

colinbowen:
## Changes in this PR
- Integrate sort options to consignment view
- Add tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-667

### After


anthonyhashemi:
Currently review the chain of filter and sort PRs Pavan created. Once those are all merged in, I will review this one as I think there will be a shared build sort function

app/main/util/queries.py
```diff
@@ -0,0 +1,18 @@
+def build_sorting_orders(request):
```
anthonyhashemi: might be worth looking at the build sort order function @pgandhizaizi made? Maybe its not entirely appropriate for this route since its quite different but maybe theres some similarity

anthonyhashemi (approved):
Generally looks good but just suggest checking if there is any shared logic with the other browse views

app/main/util/queries.py
colinbowen: Thanks, in the process of writing some tests for both filter & sort but will use the updated function.

pgandhizaizi: Hi, noticed  build_sorting_orders_consignment(request) function, it is not required as I wrote a generic sort function which will work for all views, please look at the function build_sorting_orders

colinbowen: @pgandhizaizi That's been removed and it's now using the shared sorting func.

PR #185, opened 2024-02-06
Add auth refresh flow

anthonyhashemi:
## Changes in this PR

- Add auth refresh flow

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-420

### DIFF
--- a/app/main/authorize/access_token_sign_in_required.py
+++ b/app/main/authorize/access_token_sign_in_required.py
@@ -40,7 +40,7 @@ def access_token_sign_in_required(view_func):
         g.access_token_sign_in_required = True  # Set attribute on g
 
         try:
-            if not _validate_access_token():
+            if not _validate_or_refresh_access_token():
                 session.clear()
                 return redirect(url_for("main.sign_in"))
 
@@ -63,7 +63,7 @@ def access_token_sign_in_required(view_func):
     return decorated_view
 
 
-def _validate_access_token():
+def _validate_or_refresh_access_token():
     is_valid_access_token = False
 
     keycloak_openid = get_keycloak_instance_from_flask_config()
@@ -76,5 +76,23 @@ def _validate_access_token():
             session["user_groups"] = user_groups
             if user_groups:
                 is_valid_access_token = True
+
+    # if access token not valid, attempt to refresh it with the refresh_token if present
+    if not is_valid_access_token:
+        refresh_token = session.get("refresh_token")
+        if refresh_token:
+            refreshed_token_response = keycloak_openid.refresh_token(
+                refresh_token
+            )
+            session["access_token"] = refreshed_token_response["access_token"]
+            session["refresh_token"] = refreshed_token_response["refresh_token"]
+            decoded_refreshed_token = keycloak_openid.introspect(
+                refreshed_token_response["access_token"]
+            )
+            if decoded_refreshed_token["active"] is True:
+                refreshed_user_groups = decoded_refreshed_token["groups"]
+                session["user_groups"] = refreshed_user_groups
+                if refreshed_user_groups:
+                    is_valid_access_token = True
 
     return is_valid_access_token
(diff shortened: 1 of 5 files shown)


app/main/routes.py
```diff
@@ -82,6 +82,8 @@ def callback():
 
     session["access_token"] = access_token_response["access_token"]
     session["refresh_token"] = access_token_response["refresh_token"]
+    decoded_access_token = keycloak_openid.introspect(session["access_token"])
+    session["user_groups"] = decoded_access_token["groups"]
```
anthonyhashemi, line 86: noticed that neither sign-in nor callback routes are covered by normal pytest flask tests, only by e2e playwright tests. The fact that all our permission based browse and search tests and the sign in tests work in e2e is a good coverage already but I may add a couple of simple tests for these routes in a followup PR which mock keycloack and assert on the session.

I will override the code cov check for this line because of this

app/main/authorize/access_token_sign_in_required.py
```diff
@@ -40,10 +41,25 @@ def decorated_view(*args, **kwargs):
...
         try:
-            if not _validate_access_token():
+            access_token = session["access_token"]
+            refresh_token = session["refresh_token"]
+            try:
+                (
+                    access_token,
+                    refresh_token,
+                    tokens_are_refreshed,
+                ) = _validate_or_refresh_tokens(access_token, refresh_token)
```
anthonyhashemi, line 51: _validate_or_refresh_tokens doesnt concern itself with sessions.

Did consider just testing all the edge cases in that function itself, making it a public function but after tidying up the test cases I think its just about okay how it is, if it got any bigger I would do that.

app/main/authorize/access_token_sign_in_required.py
```diff
@@ -63,18 +79,30 @@ def decorated_view(*args, **kwargs):
     return decorated_view
 
 
-def _validate_access_token():
-    is_valid_access_token = False
+def _validate_or_refresh_tokens(access_token, refresh_token):
+    tokens_are_refreshed = False
+
+    if not (access_token and refresh_token):
```
anthonyhashemi, line 85: I made the assumption that we should always have access_token and refresh_token set due to us setting them both in the callback and nobody can manipulate the session due to the flask secret.

This assumption keeps the test cases simpler

app/main/authorize/access_token_sign_in_required.py
```diff
@@ -40,10 +41,25 @@ def decorated_view(*args, **kwargs):
...
+                (
+                    access_token,
+                    refresh_token,
+                    tokens_are_refreshed,
+                ) = _validate_or_refresh_tokens(access_token, refresh_token)
+            except InvalidAccessToken:
                 session.clear()
                 return redirect(url_for("main.sign_in"))
 
+            if tokens_are_refreshed:
```
anthonyhashemi, line 56: no point to update the session if they werent refreshed, as theyd be the same values as before

app/main/routes.py
pgandhizaizi: I agree, we should have at least basic tests for code coverage for sign-in and callback routes

app/main/authorize/access_token_sign_in_required.py
pgandhizaizi: good and precise clarification, thank you

app/main/authorize/access_token_sign_in_required.py
pgandhizaizi: image

anthonyhashemi:
Just realised that even with this solution, after the refresh token expires due to the client expiry time, once we are redirected back to the sign-in page, we are redirected back to the keycloak login page which seems to auto create a new token and allow us back in to the app.

Screenshot 2024-02-08 at 18 35 33

anthonyhashemi:
> Just realised that even with this solution, after the refresh token expires due to the client expiry time, once we are redirected back to the sign-in page, we are redirected back to the keycloak login page which seems to auto create a new token and allow us back in to the app.
> 
> Screenshot 2024-02-08 at 18 35 33

Actually, after reading the keycloak documentation, I realise that it was how I tested this out that caused this issue. I had changed the client idle time but this is optional. Without it, the refresh token expires when the sso session expires, and I think that is fine for us, we don't need to log out a user from this client specifically.

pgandhizaizi (approved):
looks good, but we need to create a separate ticket to write test cases for sign in and callback route

PR #189, opened 2024-02-07
combine sort filter all views

pgandhizaizi:
## Changes in this PR

1. created new components for sorting lists and date filters
2. combined sort and filter functionality in the browse, transferring body and series view
3. added test cases for sorting+filter and end-to-end test cases for browse, transferring body and series view
4. updated readme.md to add additional parameter when running playwright codegen to ignore https errors

## JIRA ticket

AYR 707

### After

### DIFF
--- a/app/tests/test_browse.py
+++ b/app/tests/test_browse.py
@@ -2078,41 +2078,8 @@ class TestSeries:
         """
         Given a superuser accessing the browse page
         When they make a GET request with a series id
-        and select sorting option as consignment reference ascending
-        Then they should see records sorted by consignment reference in alphabetic order most recent first
-        on browse page content.
-        """
-        mock_standard_user(
-            client,
-            browse_files[0].consignment.series.body.Name,
-        )
-
-        series_id = browse_files[0].consignment.series.SeriesId
-
-        response = client.get(
-            f"/browse?series_id={series_id}&sort=consignment_reference-asc"
-        )
-
-        assert response.status_code == 200
-
-        expected_rows = [
-            [
-                "'first_body', 'first_series', '13/01/2023', '1', 'TDR-2023-FI1', "
-                "'first_body', 'first_series', '07/02/2023', '2', 'TDR-2023-SE2'"
-            ],
-        ]
-
-        verify_series_view_header_row(response.data)
-        verify_data_rows(response.data, expected_rows)
-
-    def test_browse_series_with_consignment_reference_sorting_oldest_first(
-        self, client: FlaskClient, mock_standard_user, browse_files
-    ):
-        """
-        Given a superuser accessing the browse page
-        When they make a GET request with a series id
         and select sorting option as consignment reference descending
-        Then they should see records sorted by consignment reference in reverse alphabetic order oldest first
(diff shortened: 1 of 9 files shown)


e2e_tests/test_browse.py
```diff
@@ -10,25 +10,318 @@ def test_has_title(authenticated_page: Page):
...
-def test_sort_dropdown(authenticated_page: Page):
-    authenticated_page.goto("/browse")
-    authenticated_page.select_option("#sort", "body-a")
-    authenticated_page.select_option("#sort", "body-b")
-    authenticated_page.select_option("#sort", "series-a")
-    authenticated_page.select_option("#sort", "series-b")
-    authenticated_page.select_option("#sort", "date-first")
-    authenticated_page.select_option("#sort", "date-last")
+class TestBrowse:
+    def test_browse_sort_dropdown(self, authenticated_page: Page):
```
anthonyhashemi: tests are looking good.

I'd just suggest that we could maybe get away without needing these sort dropdown e2e tests. 
I cant see much benefit since the dropdown is tested as part of the sort and filter e2e tests and then all the edge cases of the options are tested in all the standard pytest tests you have written

e2e_tests/test_browse.py
```diff
@@ -10,25 +10,318 @@ def test_has_title(authenticated_page: Page):
...
+            "Transferring body",
+            "Series",
+            "Last record transferred",
+            "Records held",
+            "Consignments within series",
+        ]
+
+        assert rows.count() == 1
+
+        if cols.count() > 0:
```
anthonyhashemi: we should not have conditional statements in tests especially around assertions as this will mean we cant trust 2 runs of the test behave the same.

There's no reason cols.count should be 0, if it is there is a problem and we want the test to error out and let us know there is a problem. Currently it will silently (and incorrectly) pass

app/tests/test_browse.py
```diff
@@ -234,6 +234,152 @@ def test_browse_get_without_filter(
         verify_browse_view_header_row(response.data)
         verify_data_rows(response.data, expected_rows)
 
+    def test_browse_get_with_transferring_body_filter_and_transferring_body_sorting_z_to_a(
```
anthonyhashemi, line 237: As suggested before, I think we could reduce a lot of this test code by paramterizing the query parameters passed into the client.get request and the expected rows

anthonyhashemi (changes requested):
Generally looks really good. Have left some suggestions about about reducing size of test code and removing potentially unnecessary e2e tests.

Have requested changes specifically due to the conditional statements in tests, as explained in the comment.

e2e_tests/test_browse.py
pgandhizaizi: removed sort dropdown test cases as requested

e2e_tests/test_browse.py
pgandhizaizi: removed condition as requested

app/tests/test_browse.py
pgandhizaizi: we need to look during the refactor and work out best way to reduce the test cases

anthonyhashemi (approved):
Nice one, looks good 👌

PR #190, opened 2024-02-07
Ayr 712/filter with any character

pgandhizaizi:
### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -256,14 +256,14 @@ def _build_series_view_query(series_id, filters, sorting_orders):
 def _build_browse_filters(query, sub_query, filters):
     transferring_body = filters.get("transferring_body")
     if transferring_body:
-        filter_value = str(f"{transferring_body}%").lower()
+        filter_value = str(f"%{transferring_body}%").lower()
         query = query.filter(
             func.lower(sub_query.c.transferring_body).like(filter_value)
         )
 
     series = filters.get("series")
     if series:
-        filter_value = str(f"{series}%").lower()
+        filter_value = str(f"%{series}%").lower()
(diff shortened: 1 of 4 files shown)


app/main/db/queries.py
```diff
@@ -256,14 +256,14 @@ def _build_series_view_query(series_id, filters, sorting_orders):
 def _build_browse_filters(query, sub_query, filters):
     transferring_body = filters.get("transferring_body")
     if transferring_body:
-        filter_value = str(f"{transferring_body}%").lower()
+        filter_value = str(f"%{transferring_body}%").lower()
```
anthonyhashemi, line 259: so I assume %STRING% is the format for a CONTAINS string filter in sqlalchmey

pgandhizaizi: yes, it refers to like operator of database

pgandhizaizi: if you want to match characters that are contained in a given string, you surround the string with percentage signs (%) on both sides of the filter value string. i.e. %t% will get results for test, star, dest, into and so on

anthonyhashemi: perfect, thanks for explaining 👌

e2e_tests/test_browse.py
```diff
@@ -74,6 +74,46 @@ def test_browse_sort_and_filter_functionality(
...
+            "Transferring body",
+            "Series",
+            "Last record transferred",
+            "Records held",
+            "Consignments within series",
+        ]
+
+        assert rows.count() == 2
+
+        if cols.count() > 0:
```
anthonyhashemi: can we remove these conditions like in the last PR?

pgandhizaizi: condition removed as requested

anthonyhashemi (approved):

PR #191, opened 2024-02-07
Ayr 664/consignment filter

colinbowen:
## Changes in this PR
- Add consignment filter query
- Update consignment tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-664

### After
Screenshot 2024-02-07 at 17 03 18


colinbowen:
I've temporarily refactored the browse func due to flake8 complexity until we do the major refactor.

pgandhizaizi:
not sure if poetry.lock require to check in, unless there are any package changes

pgandhizaizi:
my branch for combining sort and filter has been merged to the main, please refer the refactor component to user sorting-list.html for sorting and date-filters.html to use for date filters

pgandhizaizi:
browse.html was updated by Colin to remove csrf token references in browse.html, browse_series.html has one line of unwanted code - which I removed

app/templates/main/browse-consignment.html
```diff
@@ -99,340 +71,247 @@
...
-                        <div class="govuk-form-group filters-form__group filters-form__consignment-ref-group">
-                            <label class="govuk-label" for="browse_filter_consignment-ref"></label>
-                            <input class="govuk-input filters-form__consignment-ref--input"
-                                   id="browse_filter_consignment-ref"
-                                   name="width10"
-                                   type="text">
+                        <div class="govuk-form-group govuk-form-group--consignment-filter">
+                            <input class="govuk-input govuk-!-width-full govuk-input--consignment-input"
+                                   id="full"
+                                   name="file_type"
```
anthonyhashemi: did we not agree to remove file type?

app/templates/main/browse-consignment.html
```diff
@@ -99,340 +71,247 @@
...
+                        <h3 class="govuk-heading-s govuk-heading-s--consignment-filter">Dates</h3>
+                        <div class="govuk-form-group govuk-form-group__consignment-form-group">
+                            <fieldset class="govuk-fieldset">
+                                <div class="govuk-radios govuk-radios--small" data-module="govuk-radios">
+                                    <div class="govuk-radios__item">
+                                        <input class="govuk-radios__input" id="date_last_modified" name="date_filter_field"
+                                            type="radio" value="date_last_modified" {% if
+                                            user_filters["date_filter_field"]=='date_last_modified' %}checked{% endif %}>
+                                            <label class="govuk-label govuk-radios__label govuk-radios__label--consignment"
+                                                   for="date_last_modified">File last modified</label>
```
anthonyhashemi, line 130: I think we have agreed on Date last modified being called Record date in the frontend

app/templates/main/browse.html
```diff
@@ -9,7 +9,6 @@
             <div class="search__container__content">
                 <p class="govuk-body search__heading">Search for digital records</p>
                 <form method="post" action="{{ url_for('main.search') }}">
-                    {{ form.csrf_token }}
```
anthonyhashemi, line 12: why are we removing the csrf_token from these forms?

app/static/src/scss/includes/_browse-consignment.scss
```diff
@@ -0,0 +1,245 @@
+.consignment-filter {
+    &-container {
```
anthonyhashemi, line 2: are all of these styles unique to the consignment filter, or are some similar to the other filter pages? If so, we should reuse/ share styles that are used between multiple components

app/main/util/filter_sort_builder.py
```diff
@@ -41,6 +41,29 @@ def build_sorting_orders(args):
     return sorting_orders
 
 
+def build_browse_consignment_filters(args):
```
anthonyhashemi, line 44: there is an existing `_build_consignment_filters` which is called inside `_build_consignment_view_query` which is called in `browse_data`. Can we consolidate these / delete what is redundant?

anthonyhashemi:
> not sure if poetry.lock require to check in, unless there are any package changes

agreed, the poetry.lock change is probably unnecessary

app/templates/main/browse-consignment.html
pgandhizaizi: we have to work on separate ticket for it as it is not part of current sprint

app/templates/main/browse-consignment.html
pgandhizaizi: Colin is already working on this ticket and will pick up tomorrow my latest changes on consignment view

app/templates/main/browse.html
pgandhizaizi: I think they appear in query string parameter when form get function called

app/static/src/scss/includes/_browse-consignment.scss
pgandhizaizi: I think most of them being added by Candace and Colin, need to be revisited during refactoring and make it one generic style which can be used across different browse views

app/templates/main/browse.html
anthonyhashemi: could you show me?

app/static/src/scss/includes/_browse-consignment.scss
anthonyhashemi: okay, if we could, it might be easier to refactor as we go but dont want to block this PR

app/templates/main/browse-consignment.html
anthonyhashemi: as in @colinbowen is working on https://national-archives.atlassian.net/browse/AYR-718 ?

app/templates/main/browse-consignment.html
anthonyhashemi: not sure if its worth adding this here just to remove it in a following PR

app/main/util/filter_sort_builder.py
pgandhizaizi: removed function _build_consignment_filters and moved logic to _build_consignment_view_query as requested.

app/templates/main/browse-consignment.html
pgandhizaizi: file type has been removed from the consignment view as a filter - updated query, test cases and removed file type references from every where

app/templates/main/browse.html
pgandhizaizi: done

app/templates/main/browse-consignment.html
colinbowen: Will be picked up as part of AYR-718 & AYR-714

app/main/util/filter_sort_builder.py
anthonyhashemi: I think there was some confusion, there is still 2 lots of filtering for the conginment browse

anthonyhashemi (approved):

PR #193, opened 2024-02-08
browse show empty table for no result

pgandhizaizi:
## Changes in this PR

modified browse views all, transferring body, series to allow the user to show empty table when no data returned
modified breadcrumb logic to read values from database

## JIRA ticket

AYR- 715

### After


app/main/db/queries.py
```diff
@@ -560,3 +560,46 @@ def get_all_transferring_bodies():
...
+            .join(Body, Body.BodyId == Series.BodyId)
+            .filter(Series.SeriesId == series_id)
+            .first()
+        )
+
+        dict_values[0] = {"transferring_body_id": result[0]}
+        dict_values[1] = {"transferring_body": result[1]}
+        dict_values[2] = {"series": result[2]}
+    elif consignment_id:
+        result = (
```
anthonyhashemi: as discussed offline, the consignment breadcrumb logic is currently unused but once the consignment browse view has been updated, it will be updated to use this

app/main/db/queries.py
```diff
@@ -560,3 +560,46 @@ def get_all_transferring_bodies():
     for body in Body.query.all():
         bodies.append(body.Name)
     return bodies
+
+
+def get_breadcrumb_values(
```
anthonyhashemi: can we just split this out into 3 different functions, 1 for body, 1 for series, 1 for consignment since there is no shared logic for any of these?

We will also need one for the record page to be consistent.

app/main/db/queries.py
```diff
@@ -560,3 +560,46 @@ def get_all_transferring_bodies():
...
+):
+    dict_values = {}
+    if transferring_body_id:
+        transferring_body = (
+            Body.query.filter(Body.BodyId == transferring_body_id).first()
+        ).Name
+        dict_values[0] = {"transferring_body": transferring_body}
+    elif series_id:
+        result = (
+            db.session.query(Body.BodyId, Body.Name, Series.Name)
```
anthonyhashemi: [Suggestion] Instead of session.query you could use just do

```
series = Series.query.get(series_id)
body = series.body
```

I think it is more concise/readable, what do you think?

app/main/db/queries.py
```diff
@@ -560,3 +560,46 @@ def get_all_transferring_bodies():
...
     return bodies
+
+
+def get_breadcrumb_values(
+    transferring_body_id=None, series_id=None, consignment_id=None
+):
+    dict_values = {}
+    if transferring_body_id:
+        transferring_body = (
+            Body.query.filter(Body.BodyId == transferring_body_id).first()
```
anthonyhashemi: Since BodyId is the primary key for the Body Table, it is unique and so I think we can just do

```suggestion
            Body.query.get(transferring_body_id)
```

app/main/db/queries.py
```diff
@@ -560,3 +560,46 @@ def get_all_transferring_bodies():
...
+            .first()
+        )
+
+        dict_values[0] = {"transferring_body_id": result[0]}
+        dict_values[1] = {"transferring_body": result[1]}
+        dict_values[2] = {"series": result[2]}
+    elif consignment_id:
+        result = (
+            db.session.query(
+                Body.BodyId,
```
anthonyhashemi: similarly:

```
consignment = Consignment.query.get(consignment_id)
series = consignment.series
body = series.body
```

anthonyhashemi (commented):
Looks good, just left a few suggestions

app/main/db/queries.py
pgandhizaizi: yes, I am working on consignment view, will merge the changes from main and validate the functionality

app/main/db/queries.py
pgandhizaizi: separated them into four function for each browse view and record view

app/main/db/queries.py
pgandhizaizi: updated logic to follow as mentioned above

app/main/db/queries.py
pgandhizaizi: updated logic to follow above

app/main/db/queries.py
pgandhizaizi: updated logic to follow above

pgandhizaizi:
image

I had to fix this issue for commit in record.html page

app/main/db/queries.py
```diff
@@ -560,3 +560,54 @@ def get_all_transferring_bodies():
     for body in Body.query.all():
         bodies.append(body.Name)
     return bodies
+
+
```
anthonyhashemi: can you add tests for these functions and/or adjust/add 1 test for each browse view at the route level to show that the breadcrumbs are in the html for the browse, browse_body, browse_series, browse_consignment, record

pgandhizaizi: test cases are already written for each browse view to verify breadcrumb values

anthonyhashemi (approved):

PR #194, opened 2024-02-08
read user selected sort for clear filter

pgandhizaizi:
## Changes in this PR

if user has selected sort option and then click on clear filters, option, the sort value retains and sort the result according the selected option

## JIRA ticket

AYR - 713

### Before

without sort selected by user

image

### After

with sort selected by user and then click on clear filters option

image

### DIFF
--- a/app/static/src/scss/includes/_browse-all.scss
+++ b/app/static/src/scss/includes/_browse-all.scss
@@ -1,7 +1,6 @@
 
 .browse-all-filter {
     &-container {
-
       background-color: #f3f2f1;
       padding: 0.5rem;
       min-width: 100%;
@@ -11,86 +10,70 @@
     }
 
     &__header {
-
       display: flex;
     }
 
     &__icon {
-
       justify-content: space-between;
       width: 2rem;
       margin-left: auto;
     }
 
     &-from-dates {
-
       display: inline-flex;
-      margin-bottom: 15px;
     }
 
     &-to-dates {
-
       display: inline-flex;
     }
 
     &-container--series {
-
       min-width: 100%;
       display: inline-flex;
     }
   }
 
   .filters-form {
-
     float: right;
     font-size: 1rem;
 
     &__group {
-
       margin: 0.5rem 0;
     }
 
     &__title {
-
       font-weight: 700;
       font-size: 1.5rem;
     }
 
     &__transferring-body-group {
-
       display: flex;
       align-items: baseline;
     }
 
     &__series-group {
-
       display: flex;
     }
 
     &__browse-all-ref-group {
-
       display: flex;
     }
 
     &__date__label {
-
       font-size: 1rem;
     }
 
     &__date-from-group {
-
       display: flex;
       justify-content: space-between;
     }
 
     &__date-to-group {
-
       display: flex;
       justify-content: space-between;
     }
 
     &__series__container {
-
       background-color: #f3f2f1;
       padding: 0.5rem;
       min-width: 100%;
@@ -98,7 +81,6 @@
     }
 
     &__browse-all-ref__container {
-
       background-color: #f3f2f1;
       padding: 0.5rem;
       min-width: 100%;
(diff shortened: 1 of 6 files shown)


app/templates/main/date-filters.html
```diff
@@ -96,5 +96,6 @@
...
             data-module="govuk-button">Apply filters</button>
-    <a href="{% if view_name == 'browse' %} {{ browse_url }} {% elif view_name == 'transferring_body' %} {{ transferring_body_url ~ transferring_body_id }} {% elif view_name == 'series' %} {{ series_url ~ series_id }} {% elif view_name == 'consignment' %} {{ consigmnent_url ~ consignment_id }} {% endif %}">clear filters</a>
+    <a class="govuk-link govuk-link--browse-all-filter"
+       href="{% if view_name == 'browse' %} {% if sorting_order %} {{ browse_url ~ sorting_order }} {% else %} {{ browse_url }} {% endif %} {% elif view_name == 'transferring_body' %} {% if sorting_order %} {{ transferring_body_url ~ transferring_body_id  ~ sorting_order }} {% else %} {{ transferring_body_url ~ transferring_body_id }} {% endif %} {% elif view_name == 'series' %} {% if sorting_order %} {{ series_url ~ series_id ~ sorting_order }} {% else %} {{ series_url ~ series_id }} {% endif %} {% elif view_name == 'consignment' %} {% if sorting_order %} {{ consigmnent_url ~ consignment_id ~ sorting_order }} {% else %} {{ consignment_url ~ consignment_id }} {% endif %} {% endif %}">Clear filters</a>
```
anthonyhashemi: worth expanding / adding an assertion to the clear filter e2e playwright test to show that the sorting order persists after clearing? I believe you already have the test for clearing so just need an extra assertion to show the order persists

app/static/src/scss/includes/_browse-all.scss
```diff
@@ -197,43 +166,29 @@
...
+}
 
 .govuk-button {
   &__browse-all-filters-form-apply-button {
-
     margin-left: 0;
     margin-bottom: 10px;
     margin-top: 0;
   }
-
```
anthonyhashemi, line 222: was `browse-all-filters-form-clear-button` unused?

I see we are now applying `govuk-link`, whats the visual difference this causes? They look the same in the screenshots

app/templates/main/date-filters.html
```diff
@@ -96,5 +96,6 @@
     <button type="submit"
             class="govuk-button govuk-button__browse-all-filters-form-apply-button"
             data-module="govuk-button">Apply filters</button>
-    <a href="{% if view_name == 'browse' %} {{ browse_url }} {% elif view_name == 'transferring_body' %} {{ transferring_body_url ~ transferring_body_id }} {% elif view_name == 'series' %} {{ series_url ~ series_id }} {% elif view_name == 'consignment' %} {{ consigmnent_url ~ consignment_id }} {% endif %}">clear filters</a>
+    <a class="govuk-link govuk-link--browse-all-filter"
```
anthonyhashemi: I feel like the apply filters and clear filters html shouldnt be part of this date_filters template.

Also, I think we can simplify this href by doing something like 

```
<a href="{% if sorting_order %}{{ url_prefix ~ view_id ~ sorting_order }}{% else %}{{ url_prefix ~ view_id }}{% endif %}">Clear filters</a>
```

app/static/src/scss/includes/_browse-all.scss
pgandhizaizi: clear filter has been set as <a> tag and that is why css has to be changed

app/templates/main/date-filters.html
pgandhizaizi: updated clear filter test cases as requested

app/templates/main/date-filters.html
pgandhizaizi: removed apply filters and clear filters div from date-filters.html, added in individual browse view - all, transferring body and series view, will update for consignment view in AYR - 664 ticket

app/static/src/scss/includes/_browse-all.scss
pgandhizaizi: Anchor tag <a> applied for clear filter instead of button

anthonyhashemi (approved):
Chatted on slack. Wrong screenshot in after section in PR but have checked out the branch and seen the difference in styling. Looks good!

PR #201, opened 2024-02-13
use record opening date label

colinbowen:
## Changes in this PR
- Remove Closure Expiry date and use record opening date
- Added tests for closure expiry date combinations

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-714

### Before
Screenshot 2024-02-13 at 18 30 04

### After

Screenshot 2024-02-13 at 18 27 59


anthonyhashemi (approved):
Nice one

app/templates/main/browse-consignment.html
```diff
@@ -7,7 +7,7 @@
 {% set sorting_list = {
     "Record status (closed first)": "closure_type-asc",
     "Record status (open first)": "closure_type-desc",
-    "Closure expiry date (soonest first)": "closure_expiry_date-asc",
+    "Record opening date (soonest first)": "opening_date-asc",
     "Date last modified (most recent first)": "date_last_modified-desc",
     "Date last modified (oldest first)": "date_last_modified-asc",
     "Record filename (A to Z)": "file_name-asc",
```
pgandhizaizi, line 13: Please replace 'Date last modified' labels with 'Record date' as per terry's design and replace corresponding test cases too

pgandhizaizi:
noticed all unit testing is using date_last_modified as date filter, it would be good if one or two test cases being added for 'record opening date' as date filter and sorting

anthonyhashemi:
> noticed all unit testing is using date_last_modified as date filter, it would be good if one or two test cases being added for 'record opening date' as date filter and sorting

As mentioned above I dont think this PR is concerned with the label updates, just the closure consolidation.

If there were any tests for closure start and closure period, those should be updated to record opening date equivalent.

app/templates/main/browse-consignment.html
anthonyhashemi: I think this PR is just for the combination of closure start date and period into record opening date

The label changes Terry wants will happen in another PR/ ticket I believe

colinbowen: Label changes are to be done in AYR-718

colinbowen:
> noticed all unit testing is using date_last_modified as date filter, it would be good if one or two test cases being added for 'record opening date' as date filter and sorting

Added tests to check that results are sorted by opening date and filters

anthonyhashemi (approved):

PR #204, opened 2024-02-14
Redirect to sign in if no access or refresh token in session, dont error

anthonyhashemi:
Thanks to @colinbowen  for spotting this bug that I introduced when I was working on the refresh flow work. 

## Changes in this PR

Redirect to sign in if no access or refresh token in session, dont error

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-420


colinbowen (approved):
Looks good 👍

pgandhizaizi (approved):

PR #208, opened 2024-02-14
Ayr 711/breadcrumb fix || WIP

zaiziclebby:
## Changes in this PR

- Updated breadcrumb styling for the record page so that last child falls to next line irregardless of file name length. 
- Updated the breadcrumb styling across all views in the mobile view so they are consistent and match the designs provided. 
- Fixed the issue with hyperlinks in incognito and across Safari and Firefox;  they now remain the consistent light blue colour no matter the state, as per Terrys' designs. 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-711
### Before
Screenshot 2024-02-12 at 10 58 38

### After
Screenshot 2024-02-14 at 15 04 16


app/templates/main/breadcrumb.html
```diff
@@ -2,7 +2,7 @@
 {% set transferring_body_url = browse_url ~ "?transferring_body_id=" %}
 {% set series_url = browse_url ~ "?series_id=" %}
 {% set consignment_url = browse_url ~ "?consignment_id=" %}
-<div class="govuk-breadcrumbs">
+<div class="govuk-breadcrumbs  {{ 'govuk-breadcrumbs--file' if record_breadcrumbs }}">
```
anthonyhashemi: instead of introducing `record_breadcrumbs` we could just do if `if "File" in dict`

corresponding to `File` set in the top of record.html
```
{% set breadcrumbs = {
    "everything":"Everything",
    "body": [breadcrumb_values[0]["transferring_body_id"], breadcrumb_values[1]["transferring_body"]],
    "series": [breadcrumb_values[2]["series_id"], breadcrumb_values[3]["series"]],
    "consignment": [breadcrumb_values[4]["consignment_id"], breadcrumb_values[5]["consignment_reference"]],
    "File": breadcrumb_values[6]['file_name']
} %}
```

which will keep things consistent with how we check for values inside the breadcrumb dict below.

Also, whilst might be nice to rename "File" to "record" for consistency

anthonyhashemi (commented):
There's a lot of formatting changes in this PR so its a little difficult to see what changes to focus on, so maybe having a call to go through what was needed to be changed would be good

app/templates/main/breadcrumb.html
zaiziclebby: updated.  Not sure on last part of comment, so renamed to 'record', if it's incorrect, will change it @anthonyhashemi

anthonyhashemi: that was exactly what I meant, nice one

anthonyhashemi (approved):
Nice one, I haven't reviewed the scss itself but the rest looks good, might be worth getting @Terry-Price  to review the scss

PR #214, opened 2024-02-16
Update s3 key pattern to use FileId, not FilePath

anthonyhashemi:
## Changes in this PR

- Updated s3 key pattern to use FileId, not FilePath to be in line with new format of file keys of objects we store in s3

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-706


pgandhizaizi (approved):
looks good

PR #219, opened 2024-02-19
split browse view to distinct routes

pgandhizaizi:
## Changes in this PR

1. separated route for each browse view
2. separated test cases and e2e tests for each browse view
3. fixed route locations of each browse view in every possible places in code base
4.  created parameterised test cases for browse views

## JIRA ticket

AYR-620

### After


app/main/routes.py
```diff
@@ -92,106 +96,232 @@ def accessibility():
...
+            num_records_found=num_records_found,
+            dict_args={
+                k: v for k, v in request.args.items() if k not in "page"
+            },
+        )
+
+
+@bp.route("/browse/transferring_body/<uuid:_id>", methods=["GET"])
+@access_token_sign_in_required
+def browse_transferring_body(_id: uuid.UUID):
```
anthonyhashemi, line 158: why `_id` and not just `id`?

app/main/routes.py
```diff
@@ -92,106 +96,232 @@ def accessibility():
...
+    """
+    Render the browse transferring body view page.
+
+    This function retrieves search results for a specific
+    record(s) based on the 'transferring_body_id' provided in the query parameters, and renders
```
```suggestion
    record(s) based on the transferring_body 'id' provided
```

as its not from the query params anymore

app/main/routes.py
```diff
@@ -92,106 +96,232 @@ def accessibility():
...
+    # set default sort for consignment view
+    if len(sorting_orders) == 0:
+        sorting_orders["closure_type"] = "asc"
+
+    query = build_browse_consignment_query(
+        consignment_id=_id,
+        filters=filters,
+        sorting_orders=sorting_orders,
+    )
+    print(query.subquery())
```
anthonyhashemi: lets remove this print statement

app/main/routes.py
```diff
@@ -92,106 +96,232 @@ def accessibility():
...
+        2: {"series_id": series.SeriesId},
+        3: {"series": series.Name},
+        4: {"consignment_reference": consignment.ConsignmentReference},
+    }
+
+    filters = build_browse_consignment_filters(request.args)
+    sorting_orders = build_sorting_orders(request.args)
+
+    # set default sort for consignment view
+    if len(sorting_orders) == 0:
```
anthonyhashemi, line 312: is there something unique about consingments sorting that means that we need to set the default sort explicitly out here as opposed to inside build_sorting_orders like the other route?

app/templates/main/browse-series.html
```diff
@@ -124,25 +131,26 @@ <h3 class="govuk-heading-s govuk-heading-s--browse-all-filter">Date consignment
...
-                                <a href="{{ url_for('main.browse', consignment_id=record['consignment_id']) }}">{{ record["consignment_reference"] }}</a>
+                                <a href="{{ url_for('main.browse_consignment', _id=record['consignment_id']) }}">{{ record["consignment_reference"] }}</a>
                             </div>
                         </td>
                     </tr>
                 {% endfor %}
             </tbody>
         </table>
+        <!-- PAGINATION -->
+        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds">
```
anthonyhashemi, line 149: so were we previously completely missing pagination for all these views?

app/templates/main/pagination.html
```diff
@@ -5,7 +5,7 @@
         {% if results.has_prev == True %}
             <div class="govuk-pagination__prev">
                 <a class="govuk-link govuk-pagination__link"
-                   href="{{ url_for(view_name, page=current_page-1, **filters) }}"
+                   href="{% if id != None %} {{ url_for(view_name, _id=id, page=current_page-1, **dict_args) }} {% else %} {{ url_for(view_name, page=current_page-1, **dict_args) }} {% endif %}"
```
anthonyhashemi: could we rename dict_args something a bit more meaningful as its quite vague at the moment?

app/main/routes.py
pgandhizaizi: id is a reserved keyword in Python and is conflicting when used

app/main/routes.py
pgandhizaizi: as we are using generic sort from filter_sort_builder, if we want a default sorting on specific browse view, we have to define in the route of that view

app/templates/main/browse-series.html
pgandhizaizi: pagination was on browse.html for common use, it has been moved to individual browse view to pass in _id parameter based on that view

app/templates/main/pagination.html
pgandhizaizi: please suggest the name and I will replace it

app/main/routes.py
pgandhizaizi: updated comment for transferring body, series and consignment routes as requested

app/main/routes.py
pgandhizaizi: removed print statement

app/templates/main/pagination.html
pgandhizaizi: renamed dict_args to query_string_parameters at all places where dict_args used

app/main/routes.py
anthonyhashemi: Are you sure, I didn't think it was:
Screenshot 2024-02-21 at 14 16 22

app/main/routes.py
anthonyhashemi: cheers

app/main/routes.py
anthonyhashemi: thanks

app/main/routes.py
anthonyhashemi: Okay, I noticed the other routes do not set a default sort order. Does that mean we don't care the default order for those other pages? I feel like we definitely need the default sort order of results to match the default sort option in the sort dropdown. Worth double checking this is correct and if anything needs fixing we can create a bug ticket for it

app/templates/main/pagination.html
anthonyhashemi: query_string_parameters is much clearer, thank you

app/tests/test_browse_consignment.py
```diff
@@ -0,0 +1,340 @@
...
+                [
+                    [
+                        "'12/04/2023', 'fourth_file.xls', 'Closed', '12/04/2023', "
+                        "'10/03/2023', 'third_file.docx', 'Closed', '10/03/2023'"
+                    ],
+                ],
+            ),
+        ],
+    )
+    def test_browse_consignment_full_test(
```
anthonyhashemi, line 310: paramterization is helping greatly, makes these tests much easier to follow! Like we discussed, if we get a chance after delivering MVP we may revist to move the testing of edge case of filters and sort combinations to a lower level and leave fewer higher level route tests but already the paramterization is a big help.

anthonyhashemi (approved):
Really nice work, the paramrixation of tests is making the tests of these routes so much easier to digest as is the main route logic now that the browse routes are split out!

app/templates/main/browse-series.html
anthonyhashemi: Ah yes, I see 👌

app/main/routes.py
```diff
@@ -92,106 +96,238 @@ def accessibility():
...
-            4: {"consignment_reference": consignment.ConsignmentReference},
-        }
-
-        filters = build_browse_consignment_filters(request.args)
-        sorting_orders = build_sorting_orders(request.args)
+        total_records = db.session.query(
+            func.sum(query.subquery().c.records_held)
+        ).scalar()
+        if total_records:
+            num_records_found = total_records
```
anthonyhashemi, line 136: (same as comment on search PR) Isnt the only option here for query to be a positive integer or 0 so we should be fine by just doing `num_record_found = db.session.query(
            func.sum(query.subquery().c.records_held)
        ).scalar()` or could this return a different falsey value such as None?

app/tests/test_browse_consignment.py
pgandhizaizi: yes, I agree, we can perform sort and filter combinations on lower level

app/main/routes.py
pgandhizaizi: refactored num_record_found as requested

app/main/routes.py
pgandhizaizi: as discussed on slack 'id' keyword shadow 

image

anthonyhashemi: for others, id is a built in python function https://docs.python.org/3/library/functions.html.

I think due to its low usage in python code it is often as a variable name but making it distinct here avoids possible confusion 👌

app/main/routes.py
pgandhizaizi: refactored num_record_found as it was before

app/main/routes.py
pgandhizaizi: each browse view has been set with default sort order for consistency

anthonyhashemi: ended up setting the default sort options in each view, so not creating ticket 👍

PR #225, opened 2024-02-20
update search query based on a new design for search view

pgandhizaizi:
## Changes in this PR
1. modified fuzzy search query to accommodate new columns as per the new design of the search view
2. modified test cases to reflect new changes and checked all test cases passed with different sorting options
3. modified routes.py to follow the same approach as search to get total record count from query
4. updated top-search.html to display the value if request.args.get('query') has value

## JIRA ticket

AYR - 757

### Before

image

### After

image

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -347,7 +347,6 @@ def browse_consignment(_id: uuid.UUID):
 def search():
     form = SearchForm()
     search_results = None
-    sorting_orders = {}
     per_page = int(current_app.config["DEFAULT_PAGE_SIZE"])
     num_records_found = 0
     query = (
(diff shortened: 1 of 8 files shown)


app/main/routes.py
```diff
@@ -205,18 +205,25 @@ def search():
     )
     page = int(request.args.get("page", 1))
     filters = {"query": query}
+    sorting_orders = build_sorting_orders(request.args)
 
     if query:
-        fuzzy_search_query = build_fuzzy_search_query(query)
+        query = build_fuzzy_search_query(
```
anthonyhashemi: I'd suggest against reusing query for the sqlalchemy query object when it was also used as the query string and that could lead to confusion

Maybe we should rename query on line 202 to query_string to differentiate it if we don't want to use fuzzy_search_query for the query object.

app/main/routes.py
```diff
@@ -205,18 +205,25 @@ def search():
...
-            )
-        search_results = fuzzy_search_query.paginate(
-            page=page, per_page=per_page
-        )
-        num_records_found = search_results.total
+            query = query.where(Body.Name == ayr_user.transferring_body.Name)
+
+        search_results = query.paginate(page=page, per_page=per_page)
+
+        total_records = query.count()
```
anthonyhashemi: Isnt the only option here for query to be a positive integer or 0 so we should be fine by just doing `num_record_found = query.count()`. Just noticed this here, will make same comment on the browse PR.

pgandhizaizi: kept the same as browse consignment view for consistency

app/main/routes.py
pgandhizaizi: renamed query to fuzzy_search_query as how it was before to avoid confusion

app/main/routes.py
anthonyhashemi: ok

anthonyhashemi (approved):
Nice one

PR #227, opened 2024-02-21
search results summary query

pgandhizaizi:
## Changes in this PR

1. created a new search results summary screen
2. created a new search results summary query in queries.py
3. updated search route in routes.py for all access users to redirect on search results summary screen
4. added test cases for build_fuzzy_search_results_summary_query
5. added test cases for search results summary screen in test_search.py

## JIRA ticket

AYR - 761

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -94,6 +94,39 @@ def build_fuzzy_search_query(query_string: str, sorting_orders=None):
     return query
 
 
+def build_fuzzy_search_summary_query(query_string: str):
+    filter_value = str(f"%{query_string}%").lower()
+
+    fuzzy_filters = or_(
+        func.lower(Consignment.ConsignmentReference).like(filter_value),
+        func.lower(Body.Name).like(filter_value),
+        func.lower(Body.Description).like(filter_value),
+        func.lower(Series.Name).like(filter_value),
+        func.lower(Series.Description).like(filter_value),
+        func.lower(File.FileName).like(filter_value),
+    )
+
+    query = (
+        db.session.query(
+            Body.BodyId.label("transferring_body_id"),
+            Body.Name.label("transferring_body"),
+            func.count(File.FileName.label("file_name")).label("records_held"),
+        )
+        .join(Series, Series.BodyId == Body.BodyId)
+        .join(
+            Consignment,
+            Consignment.SeriesId == Series.SeriesId,
+        )
+        .join(File, File.ConsignmentId == Consignment.ConsignmentId)
+        .join(FileMetadata, File.FileId == FileMetadata.FileId, isouter=True)
+        .where(and_(func.lower(File.FileType) == "file", fuzzy_filters))
+        .group_by(Body.BodyId)
+        .order_by(Body.Name)
+    )
+
+    return query
+
+
 def build_browse_all_query(filters=None, sorting_orders=None):
     sub_query = (
         db.session.query(
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -7,7 +7,9 @@ from app.main.db.models import Body, Consignment, File, FileMetadata, Series, db
 from app.main.util.date_formatter import validate_date_range
 
 
-def build_fuzzy_search_query(query_string: str, sorting_orders=None):
+def build_fuzzy_search_query(
+    query_string: str, transferring_body_id=None, sorting_orders=None
+):
     filter_value = str(f"%{query_string}%").lower()
 
     fuzzy_filters = or_(
@@ -81,15 +83,20 @@ def build_fuzzy_search_query(query_string: str, sorting_orders=None):
         ).label("opening_date"),
     )
 
+    if transferring_body_id:
+        query = query.filter(
+            sub_query.c.transferring_body_id == transferring_body_id
+        )
+
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
-    else:
-        query = query.order_by(
-            sub_query.c.transferring_body,
-            sub_query.c.series,
-            sub_query.c.consignment_reference,
-            sub_query.c.file_name,
-        )
+    # else:
+    #    query = query.order_by(
+    #        sub_query.c.transferring_body,
+    #        sub_query.c.series,
+    #        sub_query.c.consignment_reference,
+    #        sub_query.c.file_name,
+    #    )
 
     return query
 
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -32,6 +32,7 @@ from app.main.db.queries import (
     build_browse_series_query,
     build_browse_transferring_body_query,
     build_fuzzy_search_query,
+    build_fuzzy_search_summary_query,
     get_file_metadata,
 )
 from app.main.flask_config_helpers import (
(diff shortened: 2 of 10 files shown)


app/main/routes.py
```diff
@@ -205,18 +206,43 @@ def search():
...
+    sorting_orders = build_sorting_orders(request.args)
 
+    ayr_user = AYRUser(session.get("user_groups"))
     if query:
-        fuzzy_search_query = build_fuzzy_search_query(query)
-        ayr_user = AYRUser(session.get("user_groups"))
-        if ayr_user.is_standard_user:
-            fuzzy_search_query = fuzzy_search_query.where(
-                Body.Name == ayr_user.transferring_body.Name
+        if ayr_user.is_superuser:
```
anthonyhashemi: I think the summary page should probably be in a separate route of its own which we redirect to here as opposed to having the logic in here and rendering the template.

Otherwise, how would an all access user get to the normal, non-summary search results page? I'd imagine wed have to use query params but I think a separate route would be cleaner

app/templates/main/search-results-summary.html
```diff
@@ -0,0 +1,43 @@
...
+                                <tr class="govuk-table__row govuk-!-font-size-14">
+                                    <th scope="col" class="govuk-table__header govuk-!-font-weight-bold">Results found within each Transferring body</th>
+                                    <th scope="col" class="govuk-table__header govuk-!-font-weight-bold">Records found</th>
+                                </tr>
+                            </thead>
+                            <tbody class="govuk-table__body">
+                                {% for record in results %}
+                                    <tr class="govuk-table__row">
+                                        <td class="govuk-table__cell govuk-body govuk-!-font-size-14">
+                                            <a href="{{ url_for('main.browse', transferring_body_id=record['transferring_body_id']) }}">{{ record['transferring_body'] }}</a>
```
anthonyhashemi: we need to link to the search results that a standard user would see for a given body, not the browse route

pgandhizaizi: fixed link and pointing to search-transferring-body page

app/main/routes.py
pgandhizaizi: created search-results-summary page for AAU users and renamed search.html to search-transferring-body.html to standard user

app/main/db/queries.py
```diff
@@ -81,15 +83,53 @@ def build_fuzzy_search_query(query_string: str, sorting_orders=None):
...
+
     if sorting_orders:
         query = _build_sorting_orders(query, sub_query, sorting_orders)
-    else:
-        query = query.order_by(
-            sub_query.c.transferring_body,
-            sub_query.c.series,
-            sub_query.c.consignment_reference,
-            sub_query.c.file_name,
+    # else:
```
anthonyhashemi, line 93: can this be removed?

app/main/db/queries.py
```diff
@@ -81,15 +83,53 @@ def build_fuzzy_search_query(query_string: str, sorting_orders=None):
         ).label("opening_date"),
     )
 
+    if transferring_body_id:
```
anthonyhashemi, line 86: not a big deal but was there a particular reason for moving this filter that depends on the user being a AAU or standard user from the route itself?

app/main/db/queries.py
```diff
@@ -81,15 +83,53 @@ def build_fuzzy_search_query(query_string: str, sorting_orders=None):
...
+    #        sub_query.c.file_name,
+    #    )
+
+    return query
+
+
+def build_fuzzy_search_summary_query(query_string: str):
+    filter_value = str(f"%{query_string}%").lower()
+
+    fuzzy_filters = or_(
```
anthonyhashemi, line 107: we could reduce duplication of the shared parts of the summary query and the normal query by extracting those parts into a private function

app/main/routes.py
```diff
@@ -381,12 +463,15 @@ def search():
         total_records = fuzzy_search_query.count()
         if total_records:
             num_records_found = total_records
+        else:
+            num_records_found = 0
```
anthonyhashemi, line 467: I know already discussed, but count will always return a number so no need to check if total records. could just do `num_record_found = fuzzy_search_query.count()` https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.count

app/main/routes.py
```diff
@@ -359,20 +422,39 @@ def search():
...
 
+    breadcrumb_values = {
+        0: {"transferring_body_id": _id},
+        1: {"transferring_body": Body.query.get(_id).Name},
+        2: {"query": ""},
+        3: {"search_terms": ""},
+    }
+
     if query:
+        str_items = ""
```
anthonyhashemi, line 437: I see we are building the comma separated search terms for the body search results breadcrumbs but neither of the search queries are separating the terms - that needs to be done like we discussed building up lots of OR or AND filter combinations through a for loop across each term to match this breadcrumb separation

anthonyhashemi: Had call with Terry and it seems like you are doing this as part of the search within search ticket and it will be an AND combination

app/tests/test_queries.py
```diff
@@ -49,9 +53,8 @@ def test_build_fuzzy_search_query_with_results(
 
     def test_build_fuzzy_search_query_no_results(self, client: FlaskClient):
         """
-        Given a filter value that does not match any field used for search terms
-            in any file in the database
-        When build_browse_all_query is called with it and is executed
+        Given a filter value that does not match transferring body in the database
```
anthonyhashemi, line 56: can you revert back the description to what it was:
```
        Given a filter value that does not match any field used for search terms
            in any file in the database
        When build_browse_all_query is called with it and is executed
```

I updated it when I fixed the flakiness issue

app/main/db/queries.py
pgandhizaizi: I removed it in the new ticket AYR-765 which I am currently working on if that is ok

app/main/db/queries.py
pgandhizaizi: there are two separate queries for search transferring body and search results summary, due to that added transferring_body_id as a filter which will be needed when accessing search transferring body screen

app/main/db/queries.py
pgandhizaizi: I have reduced the duplication in AYR-765 ticket where I will be using the common function to extract search_terms and interact them with both queries

app/main/routes.py
pgandhizaizi: Yes, I am covering it in AYR-765 ticket

app/tests/test_queries.py
pgandhizaizi: I will revert the comment in AYR-765, if that is ok as I am also renaming the test case
image

app/main/routes.py
pgandhizaizi: I can do it in AYR-765 ticket if that works for you
image

app/main/routes.py
```diff
@@ -359,20 +422,39 @@ def search():
...
+        query_items = query.strip().split(",")
+        # exclude any empty items
+        query_items = list(filter(None, query_items))
+
+        if len(query_items) > 0:
```
anthonyhashemi, line 442: worth extracting into its own function but happy for it to be done in followup PR alongside the othe r changes

anthonyhashemi (approved):

PR #228, opened 2024-02-21
fix: display pagination on mobile

colinbowen:
## Changes in this PR
- remove display:none for pagination on mobile sizes
- 

## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-742

### After
Screenshot 2024-02-21 at 14 27 19

### DIFF
--- a/app/static/src/scss/includes/_browse.scss
+++ b/app/static/src/scss/includes/_browse.scss
@@ -310,10 +310,6 @@
     display: block;
   }
 
-  .browse-grid--two-thirds {
-    display: none;
-  }
-
   .govuk-template__body {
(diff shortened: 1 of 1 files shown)


app/static/src/scss/includes/_browse.scss
```diff
@@ -310,10 +310,6 @@
     display: block;
   }
 
-  .browse-grid--two-thirds {
```
anthonyhashemi, line 313: I wonder why we had that there 🤔

anthonyhashemi (approved):

PR #230, opened 2024-02-22
Ayr 725/bug/revert back but fix flaky test

anthonyhashemi:
## Changes in this PR

- revert back `test_build_fuzzy_search_query_no_results` to have data in the setup so that the test shows that a search term that doesnt match any search fields of data in the db returns nothing, as opposed to searching any term against a db with no data returns nothing.
- fixed its previous flakiness by specifying all the search fields explcitly. It was previously flaky because "bar" was sometimes contained in the random strings created by the data factories.
- added pytest-repeat to dev depdencies so we can run multiple runs of a test easily when looking for flakiness

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-725


colinbowen (approved):
lgtm, nice

PR #236, opened 2024-02-26
search within search query changes

pgandhizaizi:
## Changes in this PR

1. updated search_results_summary and search_transferring_body query to support search within search
2. added new test cases in test_queries.py and test_search.py
3. fixed bug in routes.py which was causing for AAU users when they search in search results summary screen 
4. updated top-search.html to hold transferring body value if it has one

## JIRA ticket

AYR - 765

## Screenshots of UI changes

image

### After


pgandhizaizi:
fixed record link on the search page, it required updating queries.py to retrieve FileId from database

app/main/db/queries.py
```diff
@@ -7,28 +7,32 @@
...
         db.session.query(
             Body.BodyId.label("transferring_body_id"),
             Body.Name.label("transferring_body"),
+            Body.Description.label("transferring_body_description"),
             Series.SeriesId.label("series_id"),
             Series.Name.label("series"),
+            Series.Description.label("series_description"),
             Consignment.ConsignmentId.label("consignment_id"),
             Consignment.ConsignmentReference.label("consignment_reference"),
+            File.FileId.label("file_id"),
```
anthonyhashemi, line 23: why was file_id added?

app/main/db/queries.py
```diff
@@ -125,12 +129,43 @@ def build_fuzzy_search_summary_query(query_string: str):
...
+        .join(File, File.ConsignmentId == Consignment.ConsignmentId)
+        .where(func.lower(File.FileType) == "file")
+        .group_by(
+            sub_query.c.transferring_body_id, sub_query.c.transferring_body
+        )
     )
 
+    search_terms = _get_search_terms(query_string)
+    if len(search_terms) > 0:
+        for term in search_terms:
```
anthonyhashemi: in both build_fuzzy_search_summary_query and build_fuzzy_search_transferring_body_query we do this same filtering for loop/ Can we reduce this duplication so that any further changes to this only needs to be done in 1 place / less likely to introduce bugs etc.

app/main/routes.py
```diff
@@ -387,26 +389,32 @@ def search_results_summary():
...
 
-    total_records = db.session.query(
-        func.sum(query.subquery().c.records_held)
-    ).scalar()
-    if total_records:
-        num_records_found = total_records
-    else:
-        num_records_found = 0
+        total_records = db.session.query(
+            func.sum(fuzzy_search_summary_query.subquery().c.records_held)
```
anthonyhashemi, line 405: Is the following sufficient

```
total_records = db.session.query(
    func.sum(fuzzy_search_summary_query.records_held)
).scalar()
```

app/main/db/queries.py
pgandhizaizi: File_Id needed for the record page link as needs to pass in FileId as parameter, at some point during the query change we would have removed the column, 

image

app/main/routes.py
```diff
@@ -433,24 +441,26 @@ def search_transferring_body(_id: uuid.UUID):
         3: {"search_terms": ""},
     }
 
-    if query:
-        str_items = ""
+    if query.strip():
```
anthonyhashemi: since you now have `_get_search_terms`, I htink we should adjust that and reuse it here if necessary instead of having 2 lots of similar search term separation logic

app/templates/main/search-results-summary.html
```diff
@@ -11,7 +11,7 @@
             <div class="mobile__search__container-block"></div>
             <div class="govuk-grid-row">
                 <div class="govuk-grid-column-two-thirds">
-                    {% if filters["query"]|length >0 %}
+                    {% if filters["query"].strip() |length >0 %}
```
anthonyhashemi: I notice we call .strip alot both in the templates and the routes, maybe we can just do it once at the top of each relevant route?

app/templates/main/search-transferring-body.html
```diff
@@ -2,12 +2,23 @@
     "everything": "Everything",
     "summary": [breadcrumb_values[2]["query"], "Results summary"],
     "body": [breadcrumb_values[0]["transferring_body_id"], breadcrumb_values[1]["transferring_body"]],
-    "search_terms": breadcrumb_values[3]["search_terms"],
 } %}
 {% set standard_user_breadcrumbs = {
     "body": [breadcrumb_values[0]["transferring_body_id"], breadcrumb_values[1]["transferring_body"]],
-    "search_terms": breadcrumb_values[3]["search_terms"],
 } %}
+{% if breadcrumb_values[3]["search_terms"] | length > 0 %}
```
anthonyhashemi: what was wrong with the breadcrumbs before?

app/main/db/queries.py
pgandhizaizi: I wanted to separate as internal function, the issue is both queries are using similar fields but different sources of fields, one is using subquery and other uses direct table

app/main/routes.py
pgandhizaizi: the scalar() query return 'None' if no data returned, and will cause to overwrite num_records_found value of zero , which will show 'None' in frontend

app/templates/main/search-transferring-body.html
pgandhizaizi: there is no validation on search box at the moment so user can enter space or , in the search box and click on search button to fetch data, it was causing failure if on breadcrumb to show up correctly if there is not value assigned to search term (query)

app/main/routes.py
pgandhizaizi: both the functionalities are different from each other, will have a look at how can I combine to gather

app/main/db/queries.py
anthonyhashemi: ah right, cheers

app/main/routes.py
anthonyhashemi: not sure if I follow, will ping you

app/main/routes.py
anthonyhashemi: yeah they are slightly different but seem similar enough to extract the similar parts so yeah worth giving it a go.

I was thinking that youd just have a function that returns a list of pure search terms
"foo,bar,car"-> ["foo", "bar", "car"]. 

Thinking about it, is it not as simple as `query_string.strip().split(",")`

and then anything that needs to be done like the %% for the like operator or anything else is done with the output of this function

anthonyhashemi (approved):

PR #238, opened 2024-02-27
Search results table filters tray

zaiziclebby:
## Changes in this PR

-  HTML for table and filter tray added
- SCSS file updated with styles 
- Search HTML/SCSS files renamed to 'search-transferring-body'
- Added test for search filter tray

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-759
## Screenshots of UI changes
Screenshot 2024-02-28 at 14 56 51

### After
Screenshot 2024-02-28 at 14 56 51


colinbowen (approved):

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -0,0 +1,218 @@
+/* CSS for Proof of concept Search Page */
```
anthonyhashemi: can we update this comment

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -0,0 +1,218 @@
...
+  }
+}
+
+.govuk-input {
+  &--search-all-input {
+    width: 95% !important;
+  }
+}
+
+.box {
```
anthonyhashemi: is this following the css naming convention we agreed on @colinbowen ?

colinbowen: .box can be renamed to something matching bem formatting.

anthonyhashemi:
I checked out the repo to see what this looks like, generally looks good.

Main question I have is about the placeholder search terms and the funcionality.

Is there going to be a followup PR which adds the actual search terms and hooks up the delete functionality (at the moment when I click x, nothing happens)

Screenshot 2024-02-28 at 16 35 04

colinbowen:
> I checked out the repo to see what this looks like, generally looks good.
> 
> Main question I have is about the placeholder search terms and the funcionality.
> 
> Is there going to be a followup PR which adds the actual search terms and hooks up the delete functionality (at the moment when I click x, nothing happens)
> 
> Screenshot 2024-02-28 at 16 35 04

I believe there is going to be an integration follow up. The test case is reflecting the dummy data which should be noted also has to be updated

app/static/src/scss/includes/_search-transferring-body.scss
pgandhizaizi: comment has been removed

app/static/src/scss/includes/_search-transferring-body.scss
pgandhizaizi: class has been renamed to .search-term

pgandhizaizi:
static data has been put for filter tray to pass the test case, it will be removed in integration ticket as we will dynamically load the search terms based on user input in search box

PR #244, opened 2024-03-05
filter tray search term styling updated

zaiziclebby:
## Changes in this PR

- Updated the filter tray, specifically the search term box styling. 
- When a users' cursor hovers on the search term box it turns black, and so does the cancel-filter icon. 
- Updated search-.py test cases on the search filter tray. 
- Fixed record.py test (test_record_summary_list_open_file) 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-780
### Before
Screenshot 2024-03-05 at 18 05 17

### After
Screenshot 2024-03-05 at 18 05 25

### DIFF
--- a/app/static/src/scss/includes/_search-transferring-body.scss
+++ b/app/static/src/scss/includes/_search-transferring-body.scss
@@ -46,6 +46,26 @@
   }
 }
 
+.button-search-term {
+  border: none;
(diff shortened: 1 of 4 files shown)


pgandhizaizi (approved):
looks good

PR #245, opened 2024-03-05
AYR-737/refactor mobile views

colinbowen:
## Changes in this PR
- Remove mobile specific code for browse / record routes
- Update CSS / Tests
- refactor summary-list component for record route

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-737

### After
Screenshot 2024-03-05 at 18 50 50
Screenshot 2024-03-05 at 18 50 41
Screenshot 2024-03-05 at 18 50 31
Screenshot 2024-03-05 at 18 50 25
Screenshot 2024-03-05 at 18 50 16
Screenshot 2024-03-05 at 18 50 02
Screenshot 2024-03-05 at 18 49 51
Screenshot 2024-03-05 at 18 49 40
Screenshot 2024-03-05 at 18 49 32
Screenshot 2024-03-05 at 18 49 22

### DIFF
--- a/app/templates/main/browse-transferring-body.html
+++ b/app/templates/main/browse-transferring-body.html
@@ -56,7 +56,7 @@
             </table>
         </div>
         <!-- FILTERS -->
-        <div class="govuk-grid-column-one-third govuk-grid-column-one-third--browse-all-filters mobile-filters">
+        <div class="govuk-grid-column-one-third govuk-grid-column-one-third--browse-all-filters">
             <div class="browse-all-filter-container">
                 <div class="browse-filter__header">
                     <h2 class="govuk-heading-m govuk-heading-m--browse-all-filter-title">Filter</h2>
--- a/app/templates/main/browse-transferring-body.html
+++ b/app/templates/main/browse-transferring-body.html
@@ -24,7 +24,7 @@
             {% endwith %}
         </div>
         <!-- TABLE -->
-        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds mobile-table">
+        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds browse-grid--mobile-table">
             <table class="govuk-table browse-grid__table" id="tbl_result">
                 <thead class="govuk-table__head">
                     <tr class="govuk-table__row">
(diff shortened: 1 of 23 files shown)


app/templates/main/browse-transferring-body.html
```diff
@@ -16,20 +16,22 @@
...
+    <div class="browse-layout">
         <!-- SORT -->
         <div class="govuk-form-group sort-container__form">
             {% with dict = sorting_list %}
                 {% include "sorting-list.html" %}
             {% endwith %}
         </div>
         <!-- TABLE -->
-        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds">
+        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds mobile-table">
```
anthonyhashemi: would it make more sense to rename mobile-table to something different since its a bit misleading when it is applied for both desktop and mobile.

likewise the other class names like mobile-filters etc.

anthonyhashemi (dismissed):
just suggested considering class renames but looks much better now 👌

pgandhizaizi:
verified search terms are not getting overwritten when user enters multiple search terms, apart from them it looks good

pgandhizaizi (dismissed):
approved please do resolve the conflict issues before merging

app/templates/main/browse-transferring-body.html
colinbowen: I did keep the mobile class names (mobile-table, mobile-filters, mobile-record-details...) in as they are restricted to only applied when the screen viewport is at a certain width. I'm open to suggestions on different naming for these

anthonyhashemi: I was initially thinking that creating a class for this type of table, where we overide the mobile properties we care about for the mobile witdth. 

But, this is fine, its like a mixin

colinbowen: Removed mobile-filters & now using browse-grid--mobile-table to keep in sync with other class names

colinbowen: Ah gotcha, I think I've done that for the filters panel and table now

pgandhizaizi (approved):
looks good

anthonyhashemi (approved):
nice one

PR #246, opened 2024-03-06
Downloaded file's filename is citeable reference, fallbacks to filename

anthonyhashemi:
## Changes in this PR

- Make downloaded file's filename be the file's citeable reference if it exists, otherwise fallback to filename

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-646


colinbowen (approved):
Looks good

PR #248, opened 2024-03-06
Remove csrf token references from app due to no state changing api

anthonyhashemi:
## Changes in this PR

- Removed csrf token references from app due to no state changing api

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-762


pgandhizaizi (approved):
looks good

PR #253, opened 2024-03-07
AYR-708 - date filters validation | WIP

pgandhizaizi:
## Changes in this PR

1. renamed date_formatter.py to date_validator.py, all date validation code has been put in here
2. modified browse routes to perform date validation and return date validation errors to screen if there are any
3. updated test cases in test_date_validator.py for all date validation functionality
4. streamline db_date_format and python_date_format to support easy change

## JIRA ticket

AYR - 708

### After

### DIFF
--- a/app/static/src/scss/includes/_browse.scss
+++ b/app/static/src/scss/includes/_browse.scss
@@ -175,6 +175,7 @@
 
     &__content {
       margin: 0 auto;
+      width: 80%;
     }
   }
 


app/tests/test_date_validator.py
```diff
@@ -0,0 +1,363 @@
...
+            ),
+            (
+                "date_from_day=01&date_from_month=08&date_from_year=2023&"
+                "date_to_day=31&date_to_month=10&date_to_year=2023",
+                {"date_from": "01/08/2023", "date_to": "31/10/2023"},
+            ),
+            (
+                "date_from_day=&date_from_month=&date_from_year=&"
+                "date_to_day=&date_to_month=&date_to_year=",
+                {"date_from": "##/##/####", "date_to": "##/##/####"},
```
anthonyhashemi: why are we returning hashes?

app/tests/test_date_validator.py
```diff
@@ -0,0 +1,363 @@
...
+        client: FlaskClient,
+        mock_all_access_user,
+        query_params,
+        expected_results,
+    ):
+        """
+        Given on different request arguments for date filters
+        When generate_default_date_values function called
+        Then it returns a dictionary object with 6 values
+        with valid date from values get assigned to 1 to 3 items and date to values get assigned to 4 to 6 items
```
anthonyhashemi: not sure if I understand this line, I think it is a leftover from copyu paste from the above test?

app/templates/main/date-filters.html
```diff
@@ -57,7 +69,7 @@
                            id="date_to_day"
                            name="date_to_day"
                            type="text"
-                           value="{% if filters['date_range'] %}{% if filters['date_range']['date_to'] %}{{ filters['date_range']['date_to'][0:2] }}{% endif %}{% endif %}"
+                           value="{% if filters['date_range'] %}{% if filters['date_range']['date_to'] %}{{ filters['date_range']['date_to'][0:2].replace('#', '') }}{% endif %}{% endif %}"
```
anthonyhashemi: whats this replacement of hashes for?

app/tests/test_date_validator.py
pgandhizaizi: to show the correct values on date filter inputs are we are reading [0:2], [3:5] , [6:10] values in date-filters.html

app/templates/main/date-filters.html
pgandhizaizi: as the code returns '#' for blank values to set in correct format, replace require to change from '#' to '' value

app/tests/test_date_validator.py
pgandhizaizi: comments have been updated

app/main/util/date_validator.py
```diff
@@ -0,0 +1,334 @@
...
+            "5": int(date_to[1]) if date_from[1].isnumeric() else 0,
+            "6": int(date_to[2]) if date_from[2].isnumeric() else 0,
+        }
+    )
+
+    return date_values
+
+
+def generate_default_date_values(args, start_date=None):
+    global default_start_date
```
anthonyhashemi: Do we need to set default_start_date as a global variable? I'd suggest avoiding the use of global variables unless a really good reason for it, since it makes things much harder to debug and maintain

app/main/util/date_validator.py
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+    dates = _get_dates_based_on_year(date_values)
+
+    if dates:
+        return {"date_from": dates["date_from"], "date_to": dates["date_to"]}
+
+
+def validate_dates(args, validate_future_date=True):
+    global check_future_date
```
anthonyhashemi: same point here about global

app/main/util/date_validator.py
```diff
@@ -0,0 +1,334 @@
...
+
+from flask import current_app
+
+db_date_format = "%Y-%m-%d"
+python_date_format = "%d/%m/%Y"
+default_start_date = None
+check_future_date = True
+
+
+def validate_date_range(date_range):
```
anthonyhashemi: Could we now remove this validation function which is used in the query building if we validate the request args now?

app/main/routes.py
```diff
@@ -310,6 +333,25 @@ def browse_consignment(_id: uuid.UUID):
...
+    if len(date_filter_field) == 0 and any(
+        value != 0 for value in date_values.values()
+    ):
+        date_validation_errors = {
+            "date_filter_field": "Select either ‘Record date’ or ‘Record opening date’"
+        }
+    else:
+        if any(value != 0 for value in date_values.values()):
+            validate_future_date = (
+                True if not date_filter_field == "opening_date" else False
```
```suggestion
                date_filter_field != "opening_date"
```

app/main/util/date_validator.py
```diff
@@ -0,0 +1,334 @@
...
+
+    return errors
+
+
+def _validate_date(
+    day,
+    month,
+    year,
+    key,
+    value,
```
anthonyhashemi, line 199: would `date_input` be a clearer variable name than value, as its not a value?

PR #257, opened 2024-03-11
Ayr 785/record desktop and mobile view styling

zaiziclebby:
## Changes in this PR
- Styled record page in line with designs
- Styled record mobile table view via media queries 
- Updated record.py tests 
## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-785
### After

Screenshot 2024-03-11 at 13 42 54
Screenshot 2024-03-11 at 13 43 07
Screenshot 2024-03-11 at 13 43 17
Screenshot 2024-03-11 at 13 43 31

### DIFF
--- a/app/static/src/scss/includes/_browse.scss
+++ b/app/static/src/scss/includes/_browse.scss
@@ -460,7 +460,7 @@
   }
 
   .govuk-summary-list__row {
-    display: flex;
+    /* display: flex; */
     flex-wrap: wrap;
     justify-content: space-between;
   }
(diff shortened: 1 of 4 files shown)


app/static/src/scss/includes/_browse.scss
```diff
@@ -394,7 +394,7 @@
   }
 
   .govuk-summary-list__row {
-    display: flex;
```
pgandhizaizi, line 463: just suggestion, not sure if this is applied on base level, if you can copy the css and put in _record.scss and undo _browse.scss changes, can that not work ?

PR #266, opened 2024-03-13
Bug: Fix authorization on browse routes

anthonyhashemi:
## Changes in this PR
Fix:
- Made sure browse_consignment, browse_series and browse_transferring_body routes return 404 if a standard user tries to access a resource that is part of a transferrring body they do not have access to

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-819


colinbowen (approved):
Looks ok to me, good getting this in while the pentest is still going on.

PR #267, opened 2024-03-13
AYR-818/date-filter-fix-consignment

pgandhizaizi:
… date_filters to be used if not passed correct date, fixed the test case

## Changes in this PR

1. fixed issue in queries.py for build_browse_consignment_query to avoid date_filters to be used if not passed correct date, 
2. fixed the test case in test_browse_consignment.py

## JIRA ticket

AYR - 818

### After

### DIFF
--- a/app/main/db/queries.py
+++ b/app/main/db/queries.py
@@ -408,6 +408,7 @@ def build_browse_consignment_query(
                     == record_status.lower()
                 )
 
+        date_filter = None
         date_filter_field = filters.get("date_filter_field")
         if (
             date_filter_field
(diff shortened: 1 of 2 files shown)


anthonyhashemi (dismissed):

app/tests/test_browse_consignment.py
```diff
@@ -99,7 +99,7 @@ def test_browse_consignment_breadcrumb(
         "query_params, expected_results",
         [
             (
-                "date_filter_field=date_last_modified&date_from_day=01&date_from_month=05&date_from_year=2024",
+                "date_filter_field=date_last_modified&date_from_day=01&date_from_month=03&date_from_year=2024",
```
anthonyhashemi: This made me think....

do we have any tests for date in the future which raises error for being future? 

Either way for that type of test we should mock datetime so we have a fixed test where it wont change with the date it was run.

pgandhizaizi: added logic to mock date.today in test_date_validator.py for all test cases

anthonyhashemi (approved):

PR #273, opened 2024-03-14
Ayr 721/add no caching headers to routes that require authentication

anthonyhashemi:
## Changes in this PR
- added "no caching" headers to routes that require authentication to instruct browsers not to cache our authenticated pages so that when a user logs out and clicks back on browser, they cant see the private data. 

Of course the browser could ignore these instructions and we cant confirm 100% that a page wasnt cached by any server along the way but its as good as we can do.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-721

### DIFF
--- a/app/main/authorize/access_token_sign_in_required.py
+++ b/app/main/authorize/access_token_sign_in_required.py
@@ -40,44 +40,39 @@ def access_token_sign_in_required(view_func):
     def decorated_view(*args, **kwargs):
         g.access_token_sign_in_required = True  # Set attribute on g
 
+        access_token = session.get("access_token")
+        refresh_token = session.get("refresh_token")
         try:
-            access_token = session.get("access_token")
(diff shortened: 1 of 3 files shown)


app/main/authorize/access_token_sign_in_required.py
```diff
@@ -40,44 +40,39 @@ def protected_route():
     def decorated_view(*args, **kwargs):
         g.access_token_sign_in_required = True  # Set attribute on g
 
+        access_token = session.get("access_token")
```
anthonyhashemi, line 43: the only change in this file is the removal of the try except. We dont need to set `access_token_sign_in_required` bavk to False at the end of the request since flask's global object `g` gets reset on each request, so I just simplified this.

pgandhizaizi (approved):
looks good

PR #278, opened 2024-03-17
AYR-643/Search - E2E tests with standard user

pgandhizaizi:
## Changes in this PR

1. modified .env.e2e_tests.template to provide support for both AAU and standard users login, 
2. updated README.md to update environment variables, 
3. modified conftest.py to have a separate AAU and standard user fixture for the authenticated page, 
4. added table ref. search-transferring-body.html and search-results-summary.html, 
5. updated existing 'authenticate_page' ref to 'aau_user_page' in all test files in e2e_tests folder

## JIRA ticket

AYR -643

### After


anthonyhashemi (dismissed):

pgandhizaizi:
updated column header to 'Consignment reference' for consistency as similar one used for browse series view and other places too

anthonyhashemi (approved):

PR #279, opened 2024-03-18
AYR - 829- e2e tests for browse views

pgandhizaizi:
## Changes in this PR

1. added e2e tests for browse all, transferring body, series and consignment views including (title ,breadcrumb , pagination (no pagination, previous, next) (only for consignment view) , no/invalid access (except browse all view) , filter value in query string parameter, no results found , sorting , different filter options, date filter validations, clear filters)
3. removed unwanted <tr> tag on browse_consignment.html

## JIRA ticket

AYR - 829

### After

### DIFF
--- a/app/templates/main/browse-consignment.html
+++ b/app/templates/main/browse-consignment.html
@@ -60,14 +60,6 @@
                                     {% endif %}
                                 </td>
                             </tr>
-                            <tr class="govuk-table__row govuk-table__row__browse--mobile">
-                                <td class="govuk-table__cell browse__mobile-table__bottom-row">
-                                    <a href="{{ url_for('main.record', record_id=record['file_id']) }}">{{ record["file_name"] }}</a>
-                                </td>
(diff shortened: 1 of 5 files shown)


app/templates/main/browse-consignment.html
```diff
@@ -60,14 +60,6 @@
                                     {% endif %}
                                 </td>
                             </tr>
-                            <tr class="govuk-table__row govuk-table__row__browse--mobile">
```
anthonyhashemi, line 63: was this dead code?

pgandhizaizi: yes, that code was not required for mobile view

anthonyhashemi (approved):
once MVP is delivered let's evaluate our tests to see whether we need so many tests at an end to end level for things like checking for titles, whether pagination is as epxected on these different pages etc. as opposed to keeping it more for user scenarios

PR #280, opened 2024-03-19
Ayr 769/mobile search tb results

colinbowen:
## Changes in this PR
- Add mobile search layout

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-769

## Screen shots of UI changes
Screenshot 2024-03-19 at 11 56 18

Screenshot 2024-03-19 at 11 56 24

### After
Screenshot 2024-03-19 at 11 54 39
Screenshot 2024-03-19 at 11 54 47

### DIFF
--- a/app/templates/main/search-transferring-body.html
+++ b/app/templates/main/search-transferring-body.html
@@ -46,7 +46,7 @@
                                 </th>
                                 <th scope="col"
                                     class="govuk-table__header govuk-table__header--search-header search__mobile-heading">
-                                    Series / File name / Consignment ref
+                                    Series / File name / Consignment reference
                                 </th>
                                 <th scope="col"
                                     class="govuk-table__header govuk-table__header--search-header search__table__mobile--hidden search__desktop-heading">
--- a/app/templates/main/search-transferring-body.html
+++ b/app/templates/main/search-transferring-body.html
@@ -49,11 +49,11 @@
                                     Series / File name / Consignment reference
                                 </th>
                                 <th scope="col"
-                                    class="govuk-table__header govuk-table__header--search-header search__table__mobile--hidden search__desktop-heading">
+                                    class="govuk-table__header govuk-table__header--search-header search__desktop-heading">
                                     Consignment reference
                                 </th>
                                 <th scope="col"
-                                    class="govuk-table__header govuk-table__header--search-header govuk-table__header--search-header-title search__table__mobile--hidden search__desktop-heading">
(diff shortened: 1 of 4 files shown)


app/tests/test_search.py
```diff
@@ -18,6 +18,7 @@ def verify_search_transferring_body_header_row(data):
     expected_row = (
         [
             "Series",
+            "Series / File name / Consignment ref",
```
anthonyhashemi: so we expect to see words series, file name and consignment twice?

colinbowen: We're using CSS to only display one or the other for Desktop / Mobile. The HTML does contain both headers though to account for this.

anthonyhashemi: so does it look funky if user has css disabled?

pgandhizaizi: this does not look correct, should not be two sets for column headers , also the column header should be 'Consignment reference' when ever used

app/templates/main/search-transferring-body.html
```diff
@@ -26,121 +26,145 @@
...
-                                    {% endif %}
-                                </td>
+                                <th scope="col"
+                                    class="govuk-table__header govuk-table__header--search-header search__desktop-heading">
+                                    Series
+                                </th>
+                                <th scope="col"
+                                    class="govuk-table__header govuk-table__header--search-header search__mobile-heading">
+                                    Series / File name / Consignment ref
+                                </th>
```
pgandhizaizi, line 50: why don't you introduce view type of something condition based to show the dynamically

anthonyhashemi (dismissed):

pgandhizaizi:
as discussed on the call with Terry and Colin, it has been agreed to perform separate test cases for desktop and mobile views and compare the single set of column headers at all the time

app/tests/test_search.py
colinbowen: @pgandhizaizi Updated to test for desktop column headers

app/templates/main/search-transferring-body.html
colinbowen: View_type cannot account for screen sizes.

pgandhizaizi: after discussion it is clear , no need to implement view type

app/tests/test_search.py
pgandhizaizi: cleared the doubts after a call, approving the PR

pgandhizaizi (approved):

PR #285, opened 2024-03-20
Ayr 787/date filter validation styling

zaiziclebby:
## Changes in this PR
1. updated date_validator.py to retrieve specific date field errors, 

2. updated date_filters_validator.py to return the specific date field errors, 

3. updated routes.py to return date field errors to html template, 

4. updated date-filters.html to show an error class if the date field returned error, 

5. updated test cases in test_date_validator.py, test_date_filters_validator.py, 

6. amended basic styling in _browse-consignment.scss to align date_from and date_to in filter tray

7. updated env.vars and [readme.md](http://readme.md/) to set correct variables

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-787
### After
Screenshot 2024-03-19 at 18 31 17
Screenshot 2024-03-19 at 18 31 27
Screenshot 2024-03-19 at 18 31 43


pgandhizaizi:
updated .env.e2e_tests.template in this PR as it was not allowing e2e tests to work, please use the updated template to run e2e tests

app/main/util/date_validator.py
```diff
@@ -101,85 +116,116 @@ def validate_date(  # noqa: C901
     check_future_date=True,
 ):
     errors = []
+    error_field = (
```
anthonyhashemi, line 110: Instead of removing the formatting this here, let's just change what we pass in.

Instead of passing in "’Date from’", pass in "date_from"

app/main/util/date_validator.py
```diff
@@ -17,33 +17,40 @@ def validate_dates(args, browse_consignment=False):
...
+    to_day, to_month, to_year, date_to_errs, error_field = validate_date(
         args.get("date_to_day"),
         args.get("date_to_month"),
         args.get("date_to_year"),
         "‘Date to’",
         check_future_date,
     )
-    if not date_to_errs and (to_day or to_month or to_year):
+
+    if not date_to_errs:
```
anthonyhashemi, line 39: Why did we remove this condition? We shouldn't complete the date if we dont have any of to_day, to_month or to_year

app/main/util/date_filters_validator.py
```diff
@@ -15,14 +15,30 @@ def validate_date_filters(args, browse_consignment=False):
         to_month,
         to_year,
         date_validation_errors,
+        error_field,
     ) = validate_dates(args, browse_consignment=browse_consignment)
 
     if browse_consignment and "date_filter_field" in date_validation_errors:
-        return date_validation_errors, from_date, to_date, date_filters
+        date_filters = {
+            "from_day": args.get("date_from_day"),
```
anthonyhashemi: what is this change for?

app/main/routes.py
```diff
@@ -358,11 +376,17 @@ def browse_consignment(_id: uuid.UUID):
     from_date = None
     to_date = None
     date_filters = {}
+    date_error_fields = []
 
     if len(request.args) > 0:
-        date_validation_errors, from_date, to_date, date_filters = (
-            validate_date_filters(request.args, browse_consignment=True)
-        )
+        if len([k for k, v in request.args.items() if "date" in k]):
```
anthonyhashemi, line 382: whats this new condition for?

pgandhizaizi: When the user already received a validation error on the consignment view and when they click on clear filters, the validation error remains , to avoid that a check has been made to see if there is no 'date' filter values, it should not perform date validations

app/main/util/date_filters_validator.py
pgandhizaizi: this is to return the actual value entered by user in date filter inputs, earlier they were returned as None and when user receive the date validation error, the inputs were getting blank values instead of values entered by user

app/main/util/date_validator.py
pgandhizaizi: "and (to_day or to_month or to_year)" was not executed as a conditional statement and returned True or False, if 'not date_to_errs' is suffice enough to check entered date values are correct and return the correct date

app/main/util/date_validator.py
pgandhizaizi: just using the existing input of "’Date from’" to turn the value in "date_from" , no new parameter has been passed, if you still insist , we need to pass another parameter to validate_date function, please confirm and I can do the changes

anthonyhashemi: my point is that it is simpler to just change the original we pass into the function as it is only used for the purpose of this error message and error field and we literally hardcode it in the function call above

anthonyhashemi: clarified on call, all good

app/main/util/date_filters_validator.py
anthonyhashemi: spoke about shifting this into validate_dates on call

app/main/routes.py
pgandhizaizi: as shown the demo on huddle, the code is required to work for clear filters

app/main/util/date_filters_validator.py
pgandhizaizi: updated the code as requested

anthonyhashemi (dismissed):
Looks good, thanks for addressing the comments

pgandhizaizi:
updated date_validator (removed browse_consignment reference) , moved to date_filters_validator function

anthonyhashemi (approved):

PR #292, opened 2024-03-22
Fix e2e test

anthonyhashemi:
## Changes in this PR

This PR changes the e2e tests so they work when running against our test instance db. Currently this repo doesnt hold a pg dump of that test data but we are planning on adding it.

We are aware that a lot of the e2e tests are still quite brittle and that we can potentially reduce the number of them but for now just getting them working in CI.

Still currently blocked on running them in CI (in a different repo) but hopefully we be ready soon.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-855

### After


pgandhizaizi (approved):
looks good

PR #293, opened 2024-03-22
AYR-841/ Non text content (accessibility) 

zaiziclebby:
## Changes in this PR

- As per accessibility report, task was to use the null alt attribute “” to hide the image from screen reader users.
- Did this across all relevant pages;  browse views, and for the footer. 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-841
### After


pgandhizaizi (approved):
looks good

PR #295, opened 2024-03-25
AYR-837 : Accessibility - reflow 

zaiziclebby:
… to normal and setting eh width to 100%

## Changes in this PR

-  Set the whitespace on the TNA header to 'normal' so that it doesn't overflow when magnified at 400%. 
- Set the width of the header container to 100% in desktop view so that the header doesn't wrap. 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-837
### After
Screenshot 2024-03-25 at 13 24 28


pgandhizaizi (approved):
looks good

PR #296, opened 2024-03-25
AYR 858-browse series mobile view styling

pgandhizaizi:
## Changes in this PR

1. fixed styling for browse series view in browse-series.html, 
2. added specific styling in _browse-series.scss and referenced it to main.scss, 
3. updated test cases in test_browse_series.py
4. updated e2e test timeout in conftest.py

## JIRA ticket

AYR - 858

### Before

browse series desktop view
image

browse series mobile view
image

### After
browse series desk to pview
image

browse series mobile view
image

### DIFF
--- a/app/static/src/scss/includes/_browse-series.scss
+++ b/app/static/src/scss/includes/_browse-series.scss
@@ -1,0 +1,31 @@
+.browse__series__mobile__header {
+  display: none;
+}
+
+.browse__series__desktop__header {
+  display: table-cell;
+
+  &--text-align {
+    text-align: right;
+  }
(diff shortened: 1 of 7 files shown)


anthonyhashemi:
for visibility, pasting the AC from the JIRA ticket comment:

"confirmed with Terry on design , the desktop view has consignment transferred date - right aligned , while, on mobile it should be left aligned

on desktop view consignment ref should be wide enough to show ‘TDR-2023-DXGE’ on one line instead of wrapped"

anthonyhashemi (approved):

PR #301, opened 2024-03-26
AYR-851/browse all remove breadcrumb

pgandhizaizi:
## Changes in this PR
1. remove bread crumbs from browse-all.html view and set it with <p> tag value
2. added css styling for <p> tag specific to browse-all.html view

## JIRA ticket

AYR - 851

### After

### DIFF
--- a/app/static/src/scss/includes/_browse-all.scss
+++ b/app/static/src/scss/includes/_browse-all.scss
@@ -159,11 +159,8 @@
     margin-right: 0.6rem;
   }
 
-  &--browse-all-breadcrumb {
-    font-weight: bold;
-    font-size: 1.5rem;
-    text-decoration: None;
-    line-height: 1.3rem;
+  &--browse-all-signpost {
+    @include signposting-text;
     margin-top: 15px;
   }
 }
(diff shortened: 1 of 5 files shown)


anthonyhashemi (dismissed):

app/static/src/scss/includes/_browse-all.scss
```diff
@@ -158,6 +158,14 @@
     margin-top: 2rem;
     margin-right: 0.6rem;
   }
+
+  &--browse-all-breadcrumb {
```
anthonyhashemi: looks like this is largely the same as the styling in

```
.govuk-breadcrumbs {
  &__list-item {
    &:last-child span 
      text-decoration: none;
      font-weight: bold;
      color: black;
      -webkit-text-fill-color: black;
      font-size: 1.5rem;
      line-height: 1.3rem;
      cursor: default;
```

since thats where this text used to be.

Can we create a higher level class that can be used by both instead of duplicating?

pgandhizaizi: the accessibility report want breadcrumb to be removed as it is only one text , it should not have <ol><li> use, due to that replace it with specific styling for <p> tag

anthonyhashemi: we discussed offline the ability to abstract the shared styles into a mixing that can be includes in both classes:

https://www.w3schools.com/sass/sass_mixin_include.php

anthonyhashemi: have updated

anthonyhashemi (approved):

PR #303, opened 2024-03-26
AYR - 870 - update cookies page and remove cookie banner

pgandhizaizi:
## Changes in this PR

1. updated existing cookies template to match with TDR one (change reference to AYR), added only one 'session' cookie in list of cookies table
2. removed cookie banner logic across the application
3. updated /cookies route to simply render cookies template
4. updated service name to use &ndash character instead of hypen (-)
5. updated test cases which are impacted with the changes
6. renamed 'cookies policy' label to 'cookies' 

## JIRA ticket

AYR - 870

### Before

image

### After

image


colinbowen:
Looks ok, coverage needs updating

e2e_tests/test_routes.py
```diff
@@ -0,0 +1,42 @@
+from playwright.sync_api import Page
+
+
+class TestRoutes:
+    def test_cookies_get(self, standard_user_page: Page):
```
anthonyhashemi, line 5: I noticed we have this e2e test and very similar integration test (using client.get). They both assert content in the html but slightly different assertions. I think we could do with just one of these, I dont think having both adds anything as long as we move the assertions from one to the other.

app/tests/test_routes.py
```diff
@@ -0,0 +1,55 @@
+from bs4 import BeautifulSoup
+from flask.testing import FlaskClient
+
+from app.tests.assertions import assert_contains_html
+
+
+class TestRoutes:
```
anthonyhashemi, line 41: is there a reason we didnt name this file and class with "cookies" not "routes" (same for the e2e test)

anthonyhashemi (dismissed):

app/tests/test_routes.py
pgandhizaizi: the idea is to define all the static pages test cases in 'TestRoutes' class

e2e_tests/test_routes.py
pgandhizaizi: The idea is to create e2e tests for all static content pages to have full testing coverage, this is just the start as it triggered coverage failure when I created a PR and it come as not covered

app/tests/test_routes.py
anthonyhashemi: okay, teststaticpages could be better asince its not all routes? Not a blocker though, so resolving this comment

e2e_tests/test_routes.py
anthonyhashemi: we're going to review our testing approach in retro and ways of working

anthonyhashemi (approved):

PR #311, opened 2024-03-28
AYR-884/date validation issues fix

pgandhizaizi:
## Changes in this PR
1. fixed issues with date filter validation for browse consignment view, 
2. updated logic in date_filters_validator.py, 
3. removed unwanted line from routes.py, 
4. updated test cases in test_date_filters_validator.py and e2e_tests/test_browse_consignment.py

## JIRA ticket

AYR - 884

### Before
date error even though a user has only changed a selection of record status and clicked on apply filter button

image

no validation when a user selected the date filter option and did not enter any date field values
image

### After
no date error, a user has only changed a selection of record status and clicked on the apply filter button

image

date validation error when a user has chosen the date filter field option and not entered any date field values

image


app/main/util/date_filters_validator.py
```diff
@@ -90,6 +84,32 @@ def validate_date_filters(args, browse_consignment=False):
...
+    # create dictionary for date fields only
+    date_fields = {k: v for k, v in args.items() if "date" in k}
+    if "date_filter_field" in date_fields:
+        date_fields.pop("date_filter_field", None)
+
+    if date_filter_field not in [
+        "date_last_modified",
+        "opening_date",
+    ]:
+        if len([k for k, v in date_fields.items() if len(v) > 0]):
```
anthonyhashemi: I think we can get rid of the outer len() since:

- empty list: [] -> len: 0 
boolean value of [] or 0 both False
- non-empty list: ["foo"] -> len: 1 
boolean value of ["foo"] or 1 both True

same thing for the repeated check below

app/main/util/date_filters_validator.py
```diff
@@ -90,6 +84,32 @@ def validate_date_filters(args, browse_consignment=False):
...
+            date_validation_errors.clear()
+            date_validation_errors["date_filter_field"] = (
+                "Select either ‘Date of record’ or ‘Record opening date’"
+            )
+    else:
+        if not len([k for k, v in date_fields.items() if len(v) > 0]):
```
anthonyhashemi: here as well

app/main/util/date_filters_validator.py
anthonyhashemi: would it make sense to just do 
`date_fields = {k: v for k, v in args.items() if "date" in k and v >0}`
and then not need these extra conditions?

app/main/util/date_filters_validator.py
```diff
@@ -18,18 +18,12 @@ def validate_date_filters(args, browse_consignment=False):
...
-    if browse_consignment and date_filter_field not in [
-        "date_last_modified",
-        "opening_date",
-    ]:
-        error_field = []
-        date_validation_errors.clear()
-        date_validation_errors["date_filter_field"] = (
-            "Select either ‘Date of record’ or ‘Record opening date’"
+    if browse_consignment:
+        date_validation_errors = _check_consignment_date_filters(
```
anthonyhashemi: Instead of needing to pass in date_validation_errors to  `_check_consignment_date_filters` and needing it potentially clear that object, could we separate concerns by calling `_check_consignment_date_filters` before `validate_dates` and only call `validate_dates` if _check_consignment_date_filters didnt result in an error?

app/main/util/date_filters_validator.py
pgandhizaizi: updated and removed unwanted conditions

app/main/util/date_filters_validator.py
pgandhizaizi: removed unwanted condition

app/main/util/date_filters_validator.py
pgandhizaizi: as discussed on the call, moved _check_consignment_date_filters first before validate dates and refactor the code base as requested

anthonyhashemi (approved):
Nice one

PR #313, opened 2024-04-02
Ayr 883/visual regression viewports

colinbowen:
## Changes in this PR
- Sets desktop viewport regression tests
- Sets mobile viewport regression tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-883

## Screenshots of UI changes
N/A

### After


e2e_tests/test_visual_regression.py
```diff
@@ -3,47 +3,145 @@
...
+):
+    """
+    Given a page in the AYR webapp
+    When an authenticated user navigates to it
+    Then it should look as expected on a desktop viewport
+
+    If any of these tests break due to intended changes to the design of the page,
+    run pytest with `--update-snapshots --headed` flags to update the stored screenshot
+    """
+    aau_user_page.set_viewport_size(viewport)
```
anthonyhashemi, line 67: so are we certain that setting the viewport size will fix the flakiness of these tests? I guess we'll see as we run them in the coming week on different machines etc

anthonyhashemi (approved):

e2e_tests/test_visual_regression.py
colinbowen: I think they should be better, it should at minimum remove any failures due to different dimensions. The maximum dimensions I could see was around 3500 height for static content pages so 4000 (the current desktop config) should never be reached.

anthonyhashemi: okay, did you figure out why they were being rendered with different dimensions if we werent specifying anything differently? Is that just some sort of inherent randomness of browsers?

And ah yeah I didnt even consider that, but if we're sure we shouldn't exceed that then seems fine

colinbowen: The default viewport for playwright is 1280x720 which it should have been using, whether it was due to the viewport or another issue I'm not 100% sure yet. If we still have issues we may need to configure the threshold for allowed diffs within screenshots.

This playwright discussion may also be useful:
https://github.com/microsoft/playwright/issues/8161

Threshold setting:
https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-2

colinbowen: Also setting viewport size was recommended for a different but slightly similar issue for issues with screenshots: https://stackoverflow.com/questions/77245555/how-to-ignore-eliminate-difference-playwright-screenshots-in-headed-and-headless

anthonyhashemi: Okay cool, yeah lets see how it goes. I know the threshold setting works if the size of the image is the same, if its different the pixelmatching function errors out earlier than the threshold comparison.

PR #314, opened 2024-04-03
Serve static assets from cloudfront

anthonyhashemi:
## Changes in this PR

Serve static assets from cloudfront

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-873


.gitignore
```diff
@@ -7,9 +7,9 @@ zappa_settings.json
 
 ### GOV.UK Frontend ###
 app/static/src/css*
-app/static/fonts*
```
anthonyhashemi, line 10: You will need to delete 
`app/static/fonts`
and
`app/static/images`
from your local repo and run `./build.sh` again once this is merged in

README.md
```diff
@@ -162,6 +162,10 @@ Properties configurable at runtime:
 - `SECRET_KEY`: Secret key used for Flask session and security.
 - `DEFAULT_PAGE_SIZE`: set value for no. of records to show on browse/search view.
 - `DEFAULT_DATE_FORMAT`: set value to show date in specific format cross the application. i.e. "DD/MM/YYYY"
+- `RECORD_BUCKET_NAME`: name of s3 bucket that holds all of the record objects themselves
```
anthonyhashemi, line 165: this was missing so added apart of this change

app/__init__.py
```diff
@@ -49,8 +51,7 @@ def create_app(config_class, database_uri=None):
 
     # Set content security policy
     csp = {
-        "default-src": "'self'",
-        "script-src": ["'self'"],
+        "default-src": f"'self' {app.config['FLASKS3_CDN_DOMAIN']}",
```
anthonyhashemi, line 54: Content-Security-Policy meta-tag allows you to reduce the risk of [XSS](http://en.wikipedia.org/wiki/Cross-site_scripting) attacks by allowing you to define where resources can be loaded from, preventing browsers from loading data from any other locations. This makes it harder for an attacker to inject malicious code into our site. 

However, now we will want to allow the flask webapp to allow resources from FLASKS3_CDN_DOMAIN to be loaded as well as from self (for when FLASKS3_ACTIVE is False)

app/templates/base.html
```diff
@@ -2,7 +2,7 @@
 {%- from 'govuk_frontend_jinja/components/error-summary/macro.html' import govukErrorSummary -%}
 {%- from 'govuk_frontend_jinja/components/notification-banner/macro.html' import govukNotificationBanner -%}
 {%- from 'govuk_frontend_jinja/components/phase-banner/macro.html' import govukPhaseBanner -%}
-{% set assetPath = url_for('static', filename='').rstrip('/') %}
+{% set assetPath = url_for('static', filename='assets').rstrip('/') %}
```
anthonyhashemi, line 5: Becaudse we moved the assets above, this had to change, although we could actually momit this now since the default for assetPath is `"assets"`: https://github.com/LandRegistry/govuk-frontend-jinja/blob/main/govuk_frontend_jinja/templates/template.html

colinbowen (approved):
Looks good, noticed we haven't specified a region which it will use the default instead

anthonyhashemi:
> Looks good, noticed we haven't specified a region which it will use the default instead

Yeah I think thats just used if you havent already created the bucket, so I think its fine to omit here

PR #325, opened 2024-04-09
Change from 93vw to 100% to remove overflow bug

ahosgood:
## Changes in this PR

Switch from 93vw to 100% width for main search box to stop the box overflowing the screen on smaller widths.

## JIRA ticket

[none]

### Before

test-one np ayr nationalarchives gov uk_browse

### After

test-one np ayr nationalarchives gov uk_browse (1)


colinbowen (approved):
Looks good thanks :)

PR #326, opened 2024-04-09
Remove display property that causes date fields to collapse on smaller screens

ahosgood:
## Changes in this PR

Removed the `display: inline-flex;` that causes the fields in the "date to" to get misaligned on smaller screens and then collapse on really small screens.

## JIRA ticket

[none]

### Before 1

test-one np ayr nationalarchives gov uk_browse (0)

### After 1

test-one np ayr nationalarchives gov uk_browse (1)

### Before 2

test-one np ayr nationalarchives gov uk_browse (2)

### After 2

test-one np ayr nationalarchives gov uk_browse (3)

### DIFF
--- a/app/static/src/scss/includes/_browse-all.scss
+++ b/app/static/src/scss/includes/_browse-all.scss
@@ -224,7 +224,6 @@
 .govuk-date-input {
   &--browse-all-date-to {
     margin-left: 0;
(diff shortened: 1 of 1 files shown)


colinbowen (approved):
looks good.

PR #327, opened 2024-04-09
Fix invalid address HTML in accessibility statement

ahosgood:
## Changes in this PR

`<p>` elements can only contain [phrasing content](https://developer.mozilla.org/en-US/docs/Web/HTML/Content_categories#phrasing_content) and `<address>` is flow content. It works the other way around though.

This also fixes the styles as Chrome was popping the `<address>` outside of the `.govuk-body` element to keep it valid.

## JIRA ticket

[none]

### Before

test-one np ayr nationalarchives gov uk_accessibility (1)

### After

test-one np ayr nationalarchives gov uk_accessibility


colinbowen (approved):
thanks AJ

PR #328, opened 2024-04-09
Update CSP scripts

ahosgood:
## Changes in this PR

Add JavaScript elements to the CSP, allowing for inline script execution.

## JIRA ticket

[none]

## Screenshots

image

image


anthonyhashemi:
FYI @colinbowen  The JS comes from the govuk html template. We do have the js from when we fetch the govuk assets but we dont currently utilise any of it. Had considered removing it from repo since I think its just for the cookies banner from the template app we started from which we dont use but this is good for now whilst we confirm what to do with that

anthonyhashemi (approved):
cheers

PR #332, opened 2024-04-10
AYR-854/screen-magnification-fix

pgandhizaizi:
## Changes in this PR

1. combined same css from _privacy.scss, _how-to-use-this-service.scss and _terms-of-use.scss into one new file _static-content-template.scss, 
2. added logic to make div position sticky or static and show scrollbar if the window is very small to show div content

## JIRA ticket

AYR - 854

### Before

image

### After
screen magnification : 200 % - content div not sticky
image

screen magnification : 100 % - content div sticky

image


app/static/src/scss/includes/_static-content-template.scss
```diff
@@ -1,9 +1,15 @@
-/* SCSS for How-to-use-this-service page */
+/* SCSS for static content template */
 
 .govuk-grid-column-one-third {
   &--ayr-content {
     position: sticky;
+    max-height: 100dvh;
```
anthonyhashemi, line 6: why do we need the extra bits? wasnt `sticky` sufficient?

app/static/src/scss/includes/_static-content-template.scss
```diff
@@ -1,9 +1,15 @@
...
+    display: flex;
+    flex-direction: column;
     top: 1rem;
+    @media screen and (width <=810px) {
```
anthonyhashemi, line 11: so when width is less than or equal to 810px we make position of this static, not sticky as per the AA report suggestion but whats the use of lines 6-8?

app/static/src/scss/includes/_static-content-template.scss
ahosgood:
```suggestion
    max-height: calc(100dvh - 1rem);
    overflow: auto;
```

ahosgood: The box is 1rem from the top so we have to subtract that from the height.

I've then added an overflow so that if it is shorter then you get scrollbars.

app/static/src/scss/includes/_static-content-template.scss
ahosgood: I don't think we want the box to be sticky unless the columns are side by side like on larger devices.

We shouldn't have sticky content on top of scrollable content else that is a WCAG 2.2 2.4.11 failure (https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html).

If the window is not as high/tall as the sidebar then you can never get to the last two links until you scroll to the bottom of the "sticky patch".

The lines _should_ mean you can scroll within the sidebar separately to the page.

app/static/src/scss/includes/_static-content-template.scss
ahosgood: I can't review this because I don't have a local dev env set up for AYR 😅

app/static/src/scss/includes/_static-content-template.scss
anthonyhashemi: okay, thanks for clarifying 👌

pgandhizaizi:
AJ's requested suggestion has been commited

colinbowen (approved):
Looks good, tested locally.

PR #348, opened 2024-04-15
AYR-614/config-var-not-set-error

pgandhizaizi:
## Changes in this PR

1. updated env_config.py to check if variable value set in .env file, if not raise an exception when app starts

## JIRA ticket

AYR - 614

### Before
image

### After
image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,3 +1,5 @@
+import inspect
+
 from flask import Flask, g
 from flask_compress import Compress
 from flask_s3 import FlaskS3
@@ -24,6 +26,8 @@ def null_to_dash(value):
 def create_app(config_class, database_uri=None):
     app = Flask(__name__, static_url_path="/assets")
     app.config.from_object(config_class())
+
+    inspect.getmembers(config_class())
 
     force_https = False if app.config["TESTING"] else True
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -25,9 +25,9 @@ def null_to_dash(value):
 
 def create_app(config_class, database_uri=None):
     app = Flask(__name__, static_url_path="/assets")
-    app.config.from_object(config_class())
-
-    inspect.getmembers(config_class())
+    config = config_class()
+    inspect.getmembers(config)
+    app.config.from_object(config)
 
     force_https = False if app.config["TESTING"] else True
 
--- a/app/tests/test_config.py
+++ b/app/tests/test_config.py
@@ -97,10 +97,6 @@ def test_local_env_config_variable_not_set_error(monkeypatch):
         inspect.getmembers(config)
 
     assert str(error.value) == "'DEFAULT_DATE_FORMAT'"
-    # assert (
-    #    str(error.value)
-    #    == "variable name : 'DEFAULT_DATE_FORMAT' has not been set in .env file"
-    # )
 
 
 @mock_aws
--- a/app/tests/test_config.py
+++ b/app/tests/test_config.py
@@ -93,13 +93,14 @@ def test_local_env_config_variable_not_set_error(monkeypatch):
 
     config = EnvConfig()
 
-    with pytest.raises(Exception) as error:
+    with pytest.raises(KeyError) as error:
         inspect.getmembers(config)
 
-    assert (
-        str(error.value)
-        == "variable name : 'DEFAULT_DATE_FORMAT' has not been set in .env file"
-    )
+    assert str(error.value) == "'DEFAULT_DATE_FORMAT'"
+    # assert (
+    #    str(error.value)
+    #    == "variable name : 'DEFAULT_DATE_FORMAT' has not been set in .env file"
+    # )
 
 
 @mock_aws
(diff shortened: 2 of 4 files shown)


configs/env_config.py
```diff
@@ -5,4 +5,17 @@
 
 class EnvConfig(BaseConfig):
```
anthonyhashemi, line 6: this will only work for the envconfig, we will need it for the aws config too, maybe we can do shared stuff at the BaseConfig level and anything unique at the Envconfig, awsconfig level

configs/env_config.py
```diff
@@ -5,4 +5,17 @@
...
     def _get_config_value(self, variable_name):
-        return os.getenv(variable_name)
+        if variable_name in (
+            "AWS_REGION",
```
anthonyhashemi: Isnt AWS_REGION needed for the s3 download when running locally? worth verifying

configs/env_config.py
```diff
@@ -5,4 +5,17 @@
...
+            "FLASKS3_ACTIVE",
```
anthonyhashemi: FLASKS3_ACTIVE needs to be set to False if we dont want to use local static assets

configs/env_config.py
pgandhizaizi: code has been extended to check for AWS config values

configs/env_config.py
pgandhizaizi: AWS_REGION has been set in .env file locally, removed logic which is no longer required

configs/env_config.py
pgandhizaizi: FLASKS3_ACTIVE has been set in .env file locally, removed logic which is no longer required

pgandhizaizi:
image

app/tests/test_config.py
```diff
@@ -59,6 +61,47 @@ def test_local_env_vars_config_initialized(monkeypatch):
...
+    monkeypatch.setenv("RECORD_BUCKET_NAME", "test_record_bucket_name")
+    monkeypatch.setenv("AWS_REGION", "test_region")
+    monkeypatch.setenv("FLASKS3_ACTIVE", "False")
+    monkeypatch.setenv("FLASKS3_CDN_DOMAIN", "test_flasks3_cdn_domain")
+    monkeypatch.setenv("FLASKS3_BUCKET_NAME", "test_flasks3_bucket_name")
+
+    config = EnvConfig()
+
+    with pytest.raises(Exception) as error:
+        inspect.getmembers(config)
```
anthonyhashemi, line 97: so should we place this in app/_init__.py so we error out on startup?

pgandhizaizi: line inspect.getmembers(config_class()) has been added in app/init__.py as requested to check config var values exist or not

app/tests/test_config.py
```diff
@@ -123,3 +166,52 @@ def test_aws_secrets_manager_config_initialized(monkeypatch):
...
+    monkeypatch.setenv(
+        "AWS_SM_CONFIG_SECRET_ID", "test_secret_id"
+    )  # pragma: allowlist secret
+
+    config = AWSSecretsManagerConfig()
+
+    with pytest.raises(KeyError) as error:
+        inspect.getmembers(config)
+
+    assert str(error.value) == "'DEFAULT_DATE_FORMAT'"
```
anthonyhashemi, line 213: should this be the error message?

app/__init__.py
```diff
@@ -25,6 +27,8 @@ def create_app(config_class, database_uri=None):
     app = Flask(__name__, static_url_path="/assets")
     app.config.from_object(config_class())
 
+    inspect.getmembers(config_class())
```
anthonyhashemi: I think it will be better to just create the class object once above:

```
config = config_class()
inspect.getmembers(config)
app.config.from_object(config)
```

app/tests/test_config.py
```diff
@@ -123,3 +166,52 @@ def test_aws_secrets_manager_config_initialized(monkeypatch):
...
+            "AWS_REGION": "test_aws_region",
+            "KEYCLOAK_BASE_URI": "test_keycloak_base_uri",
+            "KEYCLOAK_CLIENT_ID": "test_keycloak_client_id",
+            "KEYCLOAK_REALM_NAME": "test_keycloack_realm_name",
+            "KEYCLOAK_CLIENT_SECRET": "test_keycloak_client_secret",  # pragma: allowlist secret
+            "RECORD_BUCKET_NAME": "test_record_bucket_name",
+            "FLASKS3_ACTIVE": "False",
+            "FLASKS3_CDN_DOMAIN": "test_flasks3_cdn_domain",
+            "FLASKS3_BUCKET_NAME": "test_flasks3_bucket_name",
+            # "DEFAULT_DATE_FORMAT": "test_default_date_format",
```
anthonyhashemi: can remove this commented line

pgandhizaizi: line has been removed

app/tests/test_config.py
pgandhizaizi: for consistency checking the key for both .env and aws config values

app/__init__.py
pgandhizaizi: updated changes as requested above

anthonyhashemi (approved):

PR #349, opened 2024-04-16
Fixed up e2e test fixtures and added info about playwright browsers i…

anthonyhashemi:
## Changes in this PR

- Fixed up e2e test fixtures and added info about playwright browsers in readme

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-681

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -20,7 +20,12 @@ def page(page, request) -> Page:
 
 
 @pytest.fixture
-def create_user_page(page) -> Page:
+def create_user_page(
+    page, browser_name
+) -> (
+    Page
+):  # FIXME: browser_name specified until https://github.com/microsoft/playwright-pytest/issues/172 fixed
+    # so that multiple browser flags in cli are honoured
     def _create_user_page(username, password) -> Page:
         page.goto("/sign-in")
         page.get_by_label("Email address").fill(username)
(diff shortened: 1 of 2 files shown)


e2e_tests/conftest.py
```diff
@@ -1,26 +1,11 @@
 import os
 
 import pytest
-from playwright.sync_api import Page, Playwright
+from playwright.sync_api import Page
 
 
-@pytest.fixture(params=["firefox", "chromium", "webkit"])
-def page(request, playwright: Playwright, browser_context_args) -> Page:
-    browser_type = request.param
```
anthonyhashemi, line 9: spoke with @colinbowen , creating our own browser instance here  meant that any cli flags we passed to the pytest playwright cmd werent getting picked up, e.g. specifying headed didnt make a difference. 

https://playwright.dev/python/docs/browsers shows that we can just specify the browsers we want to run on on with the browser flag, so I added that info to the readme. 

Most times locally we wont want to run all the tests against all browsers, we can save that for ci

colinbowen (dismissed):

e2e_tests/conftest.py
pgandhizaizi: so do we have to pass parameter as browser every time we run the e2e test cases ?

anthonyhashemi: if you just want to run with the default chromium, no you dont need to pass a flag, only if you want multiple browsers or firefox or webkit

pgandhizaizi: ok

pgandhizaizi (approved):

PR #354, opened 2024-04-17
AYR-961/search screens content changes

pgandhizaizi:
## Changes in this PR
1. updated content in search-results-summary.html and search-transferring-body.html 
2. updated relevant test cases in test_search.py

## JIRA ticket
AYR - 961

### Before
search results summary screen - AAU user
image
search transferring body screen - standard user
image
image

### After
search results summary screen - AAU user
image
search transferring body screen - standard user
image
image

### DIFF
--- a/app/static/src/scss/includes/_search-transferring-body.scss
+++ b/app/static/src/scss/includes/_search-transferring-body.scss
@@ -191,7 +191,7 @@ a.govuk-link {
 .govuk-button {
   &__search-filters-form-apply-button {
     height: 35px;
-    width: 70px;
+    width: 110px;
     font-size: 1rem;
     margin: 0 26px 0 0;
   }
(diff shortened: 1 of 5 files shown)


pgandhizaizi:
executed all unit test cases
image
executed all e2e tests in all browsers
image

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -191,7 +191,7 @@ a.govuk-link {
 .govuk-button {
   &__search-filters-form-apply-button {
     height: 35px;
-    width: 70px;
+    width: 110px;
```
anthonyhashemi, line 194: why did this need changing?

pgandhizaizi: earlier the button text was 'Apply', now it has changed to 'Apply terms' as per the discussion with Terry on content changes, so has to accommodate the full text as per new changes

anthonyhashemi: cool 👍

anthonyhashemi (approved):

PR #356, opened 2024-04-17
AYR-962/ update legacy code of sqlalchemy 

pgandhizaizi:
## Changes in this PR
1. updated legacy query.get references for sqlalchemy in routes.py

## JIRA ticket

AYR - 962

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -3,6 +3,7 @@ import uuid
 import boto3
 from flask import (
     Response,
+    abort,
     current_app,
     redirect,
     render_template,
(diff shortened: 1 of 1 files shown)


pgandhizaizi:
executed all unit test cases
image

pgandhizaizi:
executed e2e tests for all browser (visual regression for chromium failing due to pixel issues)
image

colinbowen:
There looks to be an updated query that we can use for file = File.query.get_or_404(record_id) inside of routes.py similar to what you've done with the other queries e.g. file = db.session.get(File, record_id).

Will then have a check to just return a 404 if not found after.

app/main/routes.py
colinbowen, line 1: There looks to be one query left using legacy SQLAlchemy:
`file = File.query.get_or_404(record_id)`

colinbowen:
```

if file is None:
    abort(404)
```

pgandhizaizi: done the changes as requested and removed the use of get_or_404() function to remove sqlalchemy warning

colinbowen (approved):

PR #357, opened 2024-04-17
AYR-963/fix download record with file extension

pgandhizaizi:
## Changes in this PR
1. updated download route logic to add file extension when citeable_reference used in routes.py
2. updated related test cases in test_download.py and e2e_tests/test_record.py

## JIRA ticket

AYR - 963

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -637,11 +637,18 @@ def download_record(record_id: uuid.UUID):
 
     s3_file_object = s3.get_object(Bucket=bucket, Key=key)
 
+    download_filename = file.FileName
+
+    if file.CiteableReference:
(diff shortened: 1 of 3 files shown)


app/main/routes.py
```diff
@@ -637,11 +637,18 @@ def download_record(record_id: uuid.UUID):
 
     s3_file_object = s3.get_object(Bucket=bucket, Key=key)
 
+    download_filename = file.FileName
+
+    if file.CiteableReference:
+        if len(file.FileName.rsplit(".", 1)) > 1:
```
colinbowen, line 643: Not sure if this will be an issue but could the file name contain a dot?

colinbowen (approved):

PR #359, opened 2024-04-18
AYR 924/record screen content changes

pgandhizaizi:
## Changes in this PR

1. updated content as requested in record.html
2. added citeable reference metadata in record.html 
3. added logic to show downloaded file name reference in record.html when citeable reference exists
4. updated routes.py to retrieve download file name based on citeable reference 
5. updated test cases in test_record.py and e2e_tests/test_record.py
6. amended css in _record.scss 
7. updated conftest.py to add a file object without Citeable reference
8.  updated snapshots for record view for visual regression tests

## JIRA ticket

AYR - 924

### Before
image

### After
image

record with citable reference
image

record without citeable reference
image


app/main/routes.py
```diff
@@ -611,11 +611,19 @@ def record(record_id: uuid.UUID):
         6: {"file_name": file.FileName},
     }
 
+    download_filename = None
```
anthonyhashemi, line 614: could be worth moving this logic to its own unit, for example a property on the file class, then we can access it like

`file.download_filename` here and then test

```
def test_download_filename():
    file = FileFactory(# with specific attributes such as citeable ref, filename)
    assert file.download_filename
```

and then a similar test for when filename is empty

Instead of needing more tests at the high level just to cover this logic.

anthonyhashemi (approved):

app/main/routes.py
pgandhizaizi: as discussed over slack, it is not possible a value to a custom property when File class is derived from DB.model, so agreed not to do this changes

PR #366, opened 2024-04-23
fix: revise back to top link font

colinbowen:
## Changes in this PR
- Use govuk-body class for back to top link

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-931

### After


pgandhizaizi (approved):
looks good

PR #371, opened 2024-04-25
AYR-965 - Review Browse Page Tab order for navigation via keyboard

ltrasca:
## Changes in this PR

- changed the order of the browser and record pages HTML elements to include pagination right after table
- because of this change, styles and tabbing order across browse and search pages should now be consistent

*Why element order change and not tabIndex to change tabbing order?*

Using tabIndex to dictate the order of keyboard tabbing would make future development more tedious and harder to maintain, changing the order of elements should be a more organic approach that ensures future consistency

## JIRA ticket

AYR-965 -> https://national-archives.atlassian.net/browse/AYR-965

### Before

image

### After

image

### Comparison to Search page

image


colinbowen (approved):
Looks good 👍

anthonyhashemi (approved):
Nice one 🎉

pgandhizaizi (approved):
looks good

PR #373, opened 2024-04-26
Remove unused css, group classes, set colour vars

colinbowen:
## Changes in this PR
- removes unused css styles
- defined variables.scss for styles that are to be reused
- set container colours to the same colour (static pages / record / browse)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-874

### After


app/static/src/scss/includes/_breadcrumbs.scss
```diff
@@ -63,38 +63,38 @@
...
+        -webkit-text-fill-color: $colour-link-visited;
       }
     }
 
     &--consignment {
       white-space: nowrap;
-      color: #1d70b8;
+      color: $colour-link-visited;
 
       &:visited {
```
ltrasca, line 95: I see we mention visited and normal color for links quite a few times for different elements, perhaps we could have an overwrites file that can do this for us for links? Might save on quite a bit of CSS space and make these more readable

anthonyhashemi (approved):
Nice one, much cleaner 👌

Definitely a good example to point to in regards to rethinking the visual regression testing approach (due to the large number of files needed for this PR even though there is what looks like no practical visual change)

app/static/src/scss/includes/_variables.scss
```diff
@@ -0,0 +1,20 @@
+// breakpoints
+$breakpoint-small: 640px;
+$breakpoint-medium: 810px;
+$breakpoint-large: 1200px;
+
+// colour__component__type
+$colour-link-normal: #0b0c0c;
```
ltrasca, line 7: checking the GDS colour standards I believe normal links should have a colour of #1d70b8, then visited links #4c2c92 https://design-system.service.gov.uk/styles/colour

app/static/src/scss/includes/_accessibility.scss
```diff
@@ -26,23 +26,23 @@
...
 
   &--accessibility {
-    color: #1d70b8;
+    color: $colour-link-visited;
     font-size: 1rem;
   }
 
   &:visited {
-    color: #1d70b8;
+    color: $colour-link-visited;
```
ltrasca, line 34: $colour-link-visited is used as a main colour and visited colour on a few instances, I think these should be separate and updated with the GDS colors

ltrasca (commented):
just a few comments from me, everything else looks gud, great work!

app/static/src/scss/includes/_variables.scss
colinbowen: Yes we should stick to GDS as much as possible, however for these links we want them to permanently stay un-clicked due to AYR specific design decisions as keeping them part of the navigation.

PR #377, opened 2024-04-29
AYR-949/refactor db queries

pgandhizaizi:
## Changes in this PR

1. updated queries for each view to only return the required fields from the database
2. updated queries to use model joins instead of explicit join conditions
3. combined browse_all and browse_transferring_body query into one
4. updated unit test cases based on the refactor changes and made sure they passed as expected

## JIRA ticket

AYR - 949

### After


pgandhizaizi:
performed unit testing
image

performed e2e tests in all browser (note :  chromium visual regression test fails due to pixelation issue on individual machine)
firefox
image
webkit
image
chromium
image

colinbowen (dismissed):

app/main/db/queries.py
```diff
@@ -637,58 +556,51 @@ def _get_file_metadata_query(file_id: uuid.UUID):
...
+            Body.Name.label("transferring_body"),
+            Series.Name.label("series"),
+            Consignment.ConsignmentReference.label("consignment_reference"),
+        )
+        .join(File.consignment)
+        .join(Consignment.series)
+        .join(Series.body)
+    ).where(
+        sub_query.c.file_id == File.FileId
+    )  # using custom join with subquery
```
anthonyhashemi: I think this was just a leftover comment, since this isnt a custom join, its a filter.

So can remove?

app/tests/test_search.py
```diff
@@ -739,7 +739,7 @@ def test_search_transferring_body_with_table_data_links(
                 ],
             ),
             (
-                "query=docx&sort=consignment-reference-desc",
+                "query=docx&sort=consignment_reference-desc",
```
anthonyhashemi: why did the sort field change from hyphen to underscore?

app/main/db/queries.py
```diff
@@ -194,58 +193,10 @@ def build_browse_all_query(filters=None, sorting_orders=None):
...
-        )
-        .join(Consignment, Consignment.ConsignmentId == File.ConsignmentId)
-        .join(Series, Series.SeriesId == Consignment.SeriesId)
-        .join(Body, Body.BodyId == Series.BodyId)
-        .where(
-            (func.lower(File.FileType) == "file")
-            & (Body.BodyId == transferring_body_id)
+    if transferring_body_id:
+        query = query.filter(
+            sub_query.c.transferring_body_id == transferring_body_id
```
anthonyhashemi, line 195: why can't we just do `query.filter(Body.BodyId==transferring_body_id)`?

app/tests/test_search.py
pgandhizaizi: it was wrongly applied , please refer the image

image

app/main/db/queries.py
pgandhizaizi: as we are referencing the subquery columns, we need to compare transferring_body_id with the subquery field (sub_query.c.transferring_body_id)

app/main/db/queries.py
pgandhizaizi: comments removed from queries.py file

app/tests/test_search.py
anthonyhashemi: oh I see, this suggests that this test isnt testing anything useful if the assertions were passing before and continue to pass with this change. We should probably update it to be more useful

pgandhizaizi: removed series and consignment reference sorting order test cases as they are not performing anything

app/main/db/queries.py
anthonyhashemi: ok

anthonyhashemi (approved):
Nice one, a lot less code to manage 😄

PR #378, opened 2024-04-29
AYR-977 - 'search transferring body' view overlapping filter tray on table

ltrasca:
## Changes in this PR

There existed an edge case where file name links with no spaces were making the file name column oversized, thus leading to the results table overflowing out of its container, the change to fix this was making the file name anchor link word-break when too large.

note: During the CSS refactor the parent class of this anchor should be completely changed, given that its used and overwritten multiple times throughout the application - having it in one place and re-using it in the future should be much better.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-977

### Before
image
image

### After
image
image

### DIFF
--- a/app/static/src/scss/includes/_search-transferring-body.scss
+++ b/app/static/src/scss/includes/_search-transferring-body.scss
@@ -106,6 +106,11 @@
       &:visited {
         color: #1d70b8;
       }
+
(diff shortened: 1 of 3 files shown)


colinbowen (approved):
Looks ok. Making a ticket to do the refactoring needed would be a good idea.

PR #381, opened 2024-05-01
AYR-957 - Close sessions on e2e test teardown

ltrasca:
## Changes in this PR

For each of the "..._user_page" fixture files inside of conftest.py:

- changed the return of the page to a yield, essentially putting the rest of the function on standby until the function its returning a value to has a chance to run
- added a goto after the yield to redirect to the sign out page, effectively ending the session

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-957

## Screenshots of UI changes

No visual changes to the app

### Before

N/A

### After

N/A


anthonyhashemi (approved):
Verified this closes the sessions in keycloak

PR #382, opened 2024-05-01
AYR-984 - Update AYR webapp README for new starters

ltrasca:
## Changes in this PR

Added documentation regarding PostgreSQL and commit signing to the README

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-984

### Before

N/A

### After

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ in which you can run all of the following commands. Alternatively you can prefix
 
 ### Install PostgreSQL
 
+PostgrSQL is a dependency required for running unit tests, which are reliant on [testing.postgresql](https://github.com/tk0miya/testing.postgresql).
 [Download & Install PostgreSQL](https://www.postgresql.org/download/)
 
 To verify that PostgreSQL is installed on your machine you can use:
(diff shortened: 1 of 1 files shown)


colinbowen (dismissed):

README.md
```diff
@@ -27,6 +27,34 @@ poetry shell
 
 in which you can run all of the following commands. Alternatively you can prefix all of the following commands with `poetry run`.
 
+### Install PostgreSQL
```
anthonyhashemi, line 30: I'd suggest mentioning that this is a requirement for running the unit tests due to the dependency on testing.postgresql https://github.com/tk0miya/testing.postgresql

README.md
```diff
@@ -27,6 +27,34 @@ poetry shell
...
+psql --version
+```
+
+and
+
+```shell
+initdb --version
+```
+
+The latter command (initdb) is required for the application to function correctly.
```
anthonyhashemi: this is for running the uni tests as above. You could connect the app to any db, including an externally hosted postgres instance.

anthonyhashemi (approved):
Nice one, will be very useful for whoever next needs to get set up 👌

PR #384, opened 2024-05-02
AYR-779 - Static pages route tests

ltrasca:
## Changes in this PR

Added tests to make sure we get status 200 from static pages and routes

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-779

### Before

N/A

### After

N/A


ltrasca:
Unit tests now cover the routes that were previously shown as not covered

image

anthonyhashemi (approved):

PR #387, opened 2024-05-10
AYR-1004 - Refactor Search transferring body results page

ltrasca:
## Changes in this PR

- Changed the search transferring body and results summary to be consistent in styling with the rest of the pages (this included changes to the filter tray, the arrangement of elements in the page, gaps, column widths - with Terrys recommendations) 
- Changed missing dates from hyphens to dashes
- Changed unit tests for the new changes, including some refactoring to test_search_transferring_body_with_table_data_links in order to make it more class and style agnostic 
- To prevent merge conflicts with other tickets, added aria-label to the buttons used to remove filter terms

Note: the search results page now uses the same styles as the browse page in order to be consistent, there should be a separate ticket where we could make these styles generic and reusable

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1004

### Before
image

image

### After
image

image


e2e_tests/snapshots/test_visual_regression/test_css_no_visual_regression_mobile/firefox-search_results_summary_mobile.png
anthonyhashemi, line 1: I noticed that for firefox and webkit the top line in this table is not bold like the bottom one

whereas for chromium it looks bold.

Not sure if its an issue, just an observation of a difference

app/tests/test_search.py
```diff
@@ -575,98 +575,46 @@ def test_search_transferring_body_with_table_data_links(
...
-        </table>
-        """
+        table = BeautifulSoup(response.data, "html.parser").find("table")
+        anchors = table.find_all("a")
+        rows = table.find_all("td")
+
+        # these tests should probably be style and class agnostic, so intead of testing pure HTML,
+        # which is tedious, I've changed it to be a lot more generic
+
+        # we expect a number of anchors with certain hrefs and text
```
anthonyhashemi: the comment was useful for understanding the change in the PR (makes sense 👌 ) but I think we can remove the comment from the code, what do you say?

e2e_tests/snapshots/test_visual_regression/test_css_no_visual_regression_mobile/firefox-search_results_summary_mobile.png
ltrasca: I can see it as well, on Firefox and Webkit 🤔 seems to be something that happens to the browse page aswell - might be an older issue 
image
image

app/tests/test_search.py
ltrasca: definitely 🤝 will remove them

app/tests/test_search.py
```diff
@@ -575,98 +575,46 @@ def test_search_transferring_body_with_table_data_links(
...
-            html,
-            "table",
-            {"class": "govuk-table"},
+        # assert our anchors contain the right text and hrefs
+        assert [anchor["href"] for anchor in anchors] == expected_anchors_hrefs
+        assert [
+            anchor.get_text() for anchor in anchors
+        ] == expected_anchors_text
+        # testting the text for each row, stripping it of white spaces and new lines and asserting against expected rows
+        assert (
```
anthonyhashemi, line 602: isnt this already tested on line 576 with `verify_search_desktop_data_rows`?

e2e_tests/snapshots/test_visual_regression/test_css_no_visual_regression/chromium-search_results_summary.png
colinbowen, line 1: Looks like there has been a change to the borders border-top & border-bottom making them bolder here.

colinbowen: Looks like this may have been done on purpose. Lmk and I'll set this to be resolved

app/tests/test_search.py
ltrasca: @anthonyhashemi In part I believe so, but it seems to only test for desktop rows using classes to select the elements (not entirely sure why) - instead the tests I added should test that all expected rows exist in the DOM, we probably dont need both but wasn't sure which one to remove 😹

e2e_tests/snapshots/test_visual_regression/test_css_no_visual_regression/chromium-search_results_summary.png
ltrasca: > I can see it as well, on Firefox and Webkit 🤔 seems to be something that happens to the browse page aswell - might be an older issue image image

Seems that it was like this before as well - I'm guessing it went under the radar

app/tests/test_search.py
anthonyhashemi: I think I understand but might be worth hopping on a call just to confirm. Im free from 4.15, if you are. @colinbowen may want to join too?

e2e_tests/snapshots/test_visual_regression/test_css_no_visual_regression/chromium-search_results_summary.png
colinbowen: Discussed on call

anthonyhashemi (approved):

PR #394, opened 2024-05-17
AYR-958 - Make visual regression tests not flaky

ltrasca:
## Changes in this PR
- Added a Dockerfile to the e2e_tests directory that enables us to create a Docker image where the e2e visual regression testing environment is stable
- Added scripts to make the build/run process of the Docker container less complicated
- Modified README to reflect changes and mention E2E visual regression testing

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-958

### Before
N/A

### After
N/A


e2e_tests/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
+pytest-playwright-visual = "^2.1.2"
+pytest-playwright = "^0.5.0"
+PyJWT = "^2.8.0"
+itsdangerous = "^2.2.0"
+Flask = "^3.0.3"
```
anthonyhashemi: why do we need flask?

e2e_tests/dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM python:3.12
+ARG parent_dir
+# this environment variable specifies where playwright should download browsers and ffmpeg, unused for now
+# ENV PLAYWRIGHT_BROWSERS_PATH=/app/pw-browsers
+WORKDIR /e2e_tests
+
+# seems like poetry doesn't pick up pyproject.toml for some unapparent reason - same with the readme, might break if removed
+COPY pyproject.toml pyproject.toml
+RUN touch README.md
```
anthonyhashemi: why do you need a blank README in here?

e2e_tests/dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+RUN pip install poetry
+RUN poetry install
+RUN poetry run playwright install --with-deps
+
+# runs the e2e visual regression tests for webkit, firefox and chromium
+CMD poetry run pytest test_visual_regression.py --base-url=https://host.docker.internal:5000/ --browser webkit --update-snapshots; \
```
anthonyhashemi: You can run across multiple browsers with 1 command, just by adding extra flags:
`--browser webkit --browser firefox --browser chromium`

e2e_tests/pyproject.toml
ltrasca: Can't remember the reason why I added it but don't think its required to run the tests, I've removed it

e2e_tests/dockerfile
ltrasca: was left over from an older version of the file, will remove 🤝

e2e_tests/dockerfile
ltrasca: got it, this is quite ideal, I previously wrote it down as that because of the comment below, should I remove the comment?
image

e2e_tests/dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3.12
+ARG parent_dir
```
colinbowen: This looks to not be used anywhere, can it be removed?

colinbowen: ```arg parent_dir```

ltrasca: Removed it 🤝

e2e_tests/dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+# seems like poetry doesn't pick up pyproject.toml for some unapparent reason - same with the readme, might break if removed
+COPY pyproject.toml pyproject.toml
+
+RUN pip install poetry
+RUN poetry install
+RUN poetry run playwright install --with-deps
+
+# runs the e2e visual regression tests for webkit, firefox and chromium
+CMD poetry run pytest test_visual_regression.py --base-url=https://host.docker.internal:5000/ --browser webkit --browser firefox --browser chromium --update-snapshots
```
colinbowen, line 14: Running this with --update-snapshots all the time won't allow us to see diffs locally. It should only update snapshots that have differences in them so this may be ok

e2e_tests/dockerfile
anthonyhashemi: This comment is still needed, it refers to the fact we pass browser_name into here and we will continue to need to do this until that github issue is fixed.

colinbowen (commented):
There may still be some work to do to ensure the "flakiness" is reduced. I had a test fail with fresh snapshots created from the docker container. PR looks ok though.

e2e_tests/dockerfile
ltrasca: Added it back

anthonyhashemi (approved):

PR #395, opened 2024-05-17
AYR-1006 - Fix for hover states

ltrasca:
## Changes in this PR

- Changed the colour variable "$colour-link-visited", which represented the normal blue color of a link, to "$colour-link-default", and added some more link color states (hover, visited)
- Changed some classes to be more BEM-like and descriptive, e.g. changed "__ayr-content" to "__1-rem-mb-font" (because it is used to change margin bottom and font to be 1 rem)
- Modified some classes where elements could be overwritten by multiple different files at the same time, and changed behaviour to be more along the lines of the modifier classes pattern (this was most likely causing elements to sometimes be inconsistent in their hover state) - e.g. instead of overwriting &:hover inside .govuk-link, which carries the risk of affecting links in other pages across the app, there is now a modifier of "__no-visited-color" that makes sure links don't have a visited colour.
- Links on the left should now not have any visited link colour but links on the right do (as requested by Terry)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1006

### Before
image
image

### After
image
image

### DIFF
--- a/app/static/src/scss/includes/_accessibility.scss
+++ b/app/static/src/scss/includes/_accessibility.scss
@@ -26,23 +26,27 @@
   }
 
   &--accessibility {
-    color: $colour-link-visited;
+    color: $colour-link-default;
     font-size: 1rem;
   }
 
   &:visited {
-    color: $colour-link-visited;
+    color: $colour-link-default;
+  }
+
+  &:hover {
+    color: $colour-link-hover;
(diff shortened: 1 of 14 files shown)


anthonyhashemi (approved):
Looks great, thanks for the info in the PR description covering the what and why of these changes

PR #400, opened 2024-05-21
AYR-1010 - Setup Pa11y CI as a custom pre-commit hook

ltrasca:
## Changes in this PR
- Added a configuration file for pa11y-ci inside configs/ containing the URLs that should be tested
- Added a bash script that calls pa11y-ci via npx
- Included an entry inside .pre-commit-config.yaml that calls the bash script mentioned above as a custom pre-commit hook

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1010

### Before
N/A

### After
N/A

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,7 +65,7 @@ repos:
         language: system
         pass_filenames: false
         verbose: true
-        stages: [push, manual]
+        stages: [commit, manual]
 
 - repo: https://github.com/thibaudcolas/pre-commit-stylelint
   rev: v16.2.1
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -65,7 +65,6 @@ repos:
         language: system
         pass_filenames: false
         verbose: true
-        stages: [merge-commit]
 
 - repo: https://github.com/thibaudcolas/pre-commit-stylelint
   rev: v16.2.1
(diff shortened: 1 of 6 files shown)


custom_hooks/pa11y_ci_precommit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set +e
+
+export $(grep -v '^#' .env.e2e_tests | xargs)
+
+# Install pa11y-ci if not installed
+npm list | grep pa11y-ci || npm install --save-dev pa11y-ci --no-shrinkwrap
```
colinbowen:
```suggestion
OUTPUT=$(npx pa11y-ci --config configs/pa11y_ci_precommit.js)
echo "${OUTPUT}"
```

colinbowen: Might be able to remove the cmd from package.json if you use npx

anthonyhashemi (approved):
Nice one 👌

.pre-commit-config.yaml
```diff
@@ -57,6 +57,15 @@ repos:
         language: system
         pass_filenames: false
 
+-   repo: local
+    hooks:
```
anthonyhashemi, line 61: Not for now, but like we discussed if we decide we want to have it just as a pre-push hook if we find it takes too long on each commit, we can specify stages https://pre-commit.com/#config-stages

PR #402, opened 2024-05-22
AYR-1036 - Long TB name causing bad styling on Series Browse view

ltrasca:
## Changes in this PR
- removed the no-break class from the transferring_body and added a class for word-break
- fixed a small bug where the pagination buttons had a visited link color
- modified the width of transferring_body to be one quarter using the GDS override styles

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1036

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_browse-series.scss
+++ b/app/static/src/scss/includes/_browse-series.scss
@@ -20,6 +20,10 @@
   &--consignment-reference {
     white-space: nowrap;
   }
+
+  &.word-break {
+    word-break: break-word;
+  }
 }
 
 @media screen and (width <= $breakpoint-medium) {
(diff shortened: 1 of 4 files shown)


Terry-Price:
Looks good @ltrasca .
Does it work like this is the transferring body if multiple words in the same way. (no just one long word)???

ltrasca:
> Looks good @ltrasca . Does it work like this is the transferring body if multiple words in the same way. (no just one long word)???

Hey @Terry-Price yes it works in the same way, I've attached a screenshot of an example below

image

colinbowen (approved):
Looks good, ok to merge once approved by Terry.

Terry-Price:
All good then @ltrasca . thank you.

PR #403, opened 2024-05-23
AYR-1037 & AYR-1038 - Broken pagination

ltrasca:
## Changes in this PR
- Added 'id' as a None property to the functions that render the Browse and Search Results Summary templates -> the pagination template was checking if this property was None (ideally it would have been None by default if it doesn't exist but alas, Jinja) and because it didn't exist it was introducing a '_id' query into the next URL which in turn was breaking pagination.
- Deleted the govuk-pagination__link class in our SCSS (which had inconsistent performance across pages) and replaced it with the previously created .govuk-link **__no-visited-color** modifier, which should ensure links always have a blue color. Please note that the govuk-pagination__link class still exists inside the HTML itself, this is because a lot of unit tests still rely on it, which is not ideal but can be fixed in a separate ticket.
- Moved the pagination component inside the Search Results Summary template to be inside the same container as the table, ensuring that its always placed under it as opposed to next to it.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1037
https://national-archives.atlassian.net/browse/AYR-1038

### Before
- https://127.0.0.1:5000/browse?_id=&page=2
image

- https://127.0.0.1:5000/search_results_summary?_id=&page=2&query=a
image

- https://127.0.0.1:5000/search_results_summary
image

### After
- https://127.0.0.1:5000/browse?page=3
image

- https://127.0.0.1:5000/search_results_summary?page=3&query=a
image

### DIFF
--- a/app/templates/main/pagination.html
+++ b/app/templates/main/pagination.html
@@ -4,7 +4,7 @@
          aria-label="Pagination">
         {% if results.has_prev == True %}
             <div class="govuk-pagination__prev">
-                <a class="govuk-link govuk-pagination__link"
+                <a class="govuk-link govuk-link__no-visited-color govuk-pagination__link"
                    href="{% if id != None %} {{ url_for(view_name, _id=id, page=current_page-1, **query_string_parameters) }} {% else %} {{ url_for(view_name, page=current_page-1, **query_string_parameters) }} {% endif %}"
                    rel="prev">
                     <svg class="govuk-pagination__icon govuk-pagination__icon--prev"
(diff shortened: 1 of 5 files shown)


anthonyhashemi (approved):
Nice one 🎉

anthonyhashemi:
Could you create a tech debt ticket for the backlog for removing those legacy classes in the html and fixing the tests

app/templates/main/pagination.html
```diff
@@ -4,7 +4,7 @@
          aria-label="Pagination">
         {% if results.has_prev == True %}
             <div class="govuk-pagination__prev">
-                <a class="govuk-link govuk-pagination__link"
+                <a class="govuk-link govuk-link__no-visited-color govuk-pagination__link"
```
anthonyhashemi, line 7: Could you create a tech debt ticket for the backlog for removing the `govuk-pagination__link` class in the html and fixing the tests

colinbowen (approved):
Looks good, thanks for the db setup script 👍

app/templates/main/pagination.html
ltrasca: @anthonyhashemi created a ticket for this https://national-archives.atlassian.net/browse/AYR-1055

PR #408, opened 2024-05-31
Spoof e2e tests user agent

anthonyhashemi:
## Changes in this PR

Spoof e2e tests user agent to get past bot control rules when testing against deployed instances.

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -69,6 +69,11 @@ def browser_context_args(browser_context_args):
     """
     return {
         **browser_context_args,
+        "user_agent": (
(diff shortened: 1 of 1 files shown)


e2e_tests/conftest.py
colinbowen, line 1: Looks ok, is there a reason why Chrome & Safari are part of the same string?

colinbowen (approved):

e2e_tests/conftest.py
colinbowen: Just a visual preference for code maintainability so I've approved 👍

anthonyhashemi: nah no reason, just chunked in by rough line size as precommit complained about long lines

PR #410, opened 2024-06-04
Upgrade Jinja2 

ltrasca:
## Changes in this PR

- Upgraded jinja2 to 3.1.4
- Added vulnerability reported by Safety CLI to ignore list for now

## JIRA ticket
N/A

### Before
N/A

### After
N/A


custom_hooks/safety_check.sh
```diff
@@ -13,4 +13,4 @@ if ! command -v safety &> /dev/null; then
     pip install safety
 fi
 
-poetry export --without-hashes -f requirements.txt | safety check --full-report --stdin
+poetry export --without-hashes -f requirements.txt | safety check --full-report --stdin --ignore=70612
```
anthonyhashemi, line 16: could we add a comment just explaining why we are ignoring 70612 just so we arent confused down the road

anthonyhashemi: cheers

anthonyhashemi (approved):

PR #411, opened 2024-06-04
Deploy to test-one and not to stg on merge to main

anthonyhashemi:
## Changes in this PR

Deploy to test-one and not to stg on merge to main since we want to run e2e tests continuously on code from main and to keep in line with tdr stg should be more stable. This may be revisited in the future. Currently stg has no data anyway so doesnt matter too much


ltrasca (approved):
looks good to me 👍

PR #419, opened 2024-06-10
AYR-1055 - Remove 'govuk-pagination__link' from tests

ltrasca:
## Changes in this PR

- replaced pagination tests to locate elements using testid instead of classes inside of unit tests and e2e tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1055

### Before
N/A

### After
N/A

### DIFF
--- a/app/templates/main/pagination.html
+++ b/app/templates/main/pagination.html
@@ -4,7 +4,8 @@
          aria-label="Pagination">
         {% if results.has_prev == True %}
             <div class="govuk-pagination__prev">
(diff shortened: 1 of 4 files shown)


anthonyhashemi:
does this fix the failing e2e test?

ltrasca:
@anthonyhashemi this refers to something different (for pagination), I'm working on that ticket in parallel.

anthonyhashemi (approved):
Cool, this looks good then 😄

PR #422, opened 2024-06-11
Add err checking for get & read file

colinbowen:
## Changes in this PR
- change to use send_file (https://flask.palletsprojects.com/en/3.0.x/api/#flask.send_file) as this can try to detect the mimetype it from the file name.
- Default to octet-stream content type if there isn't a file type. See: https://www.rfc-editor.org/rfc/rfc9110.html#name-content-type. Our files stored in S3 don't have this set and could be causing an issue.
- Add 404 error on getting file from S3 in download_record
- Add 500 error on reading file from S3 in download_record
- Add tests

## JIRA ticket
- https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-1031

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -643,10 +643,13 @@ def download_record(record_id: uuid.UUID):
 
     s3 = boto3.client("s3")
     bucket = current_app.config["RECORD_BUCKET_NAME"]
-
     key = f"{file.consignment.ConsignmentReference}/{file.FileId}"
 
-    s3_file_object = s3.get_object(Bucket=bucket, Key=key)
+    try:
+        s3_file_object = s3.get_object(Bucket=bucket, Key=key)
+    except Exception as e:
+        current_app.logger.error(f"S3 error: {e}")
+        abort(404)
 
     download_filename = file.FileName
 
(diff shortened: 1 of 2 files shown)


app/main/routes.py
```diff
@@ -643,10 +643,13 @@ def download_record(record_id: uuid.UUID):
...
     bucket = current_app.config["RECORD_BUCKET_NAME"]
-
     key = f"{file.consignment.ConsignmentReference}/{file.FileId}"
 
-    s3_file_object = s3.get_object(Bucket=bucket, Key=key)
+    try:
+        s3_file_object = s3.get_object(Bucket=bucket, Key=key)
+    except Exception as e:
+        current_app.logger.error(f"S3 error: {e}")
+        abort(404)
```
anthonyhashemi, line 653: can we add a test for this

app/main/routes.py
```diff
@@ -656,10 +659,16 @@ def download_record(record_id: uuid.UUID):
                 file.CiteableReference + "." + file.FileName.rsplit(".", 1)[1]
             )
 
+    try:
+        file_content = s3_file_object["Body"].read()
+    except Exception as e:
+        current_app.logger.error(f"Error reading S3 file content: {e}")
```
anthonyhashemi, line 666: and this case

colinbowen: Added tests with mocks for this

app/tests/test_download.py
```diff
@@ -158,6 +159,66 @@ def test_download_record_standard_user_without_citable_reference(
...
+
+        bucket_name = "test_bucket"
+        file = FileFactory(
+            FileType="file", FileName="testimage.png", CiteableReference=None
+        )
+        create_mock_s3_bucket_with_object(bucket_name, file)
+        app.config["RECORD_BUCKET_NAME"] = bucket_name
+
+        # Mock the S3 client and its get_object method
+        mock_s3_client = MagicMock()
```
anthonyhashemi, line 204: If youre using moto I dont think you need to create magic mocks for s3. You could use the moto api to put an object: https://docs.getmoto.org/en/latest/docs/getting_started.html#moto-usage

But happy to leave as is if this works

anthonyhashemi (approved):

PR #423, opened 2024-06-11
AYR-1079 - Fix e2e tests after changes to tables

ltrasca:
## Changes in this PR

- Added a fixture for utilities that includes standardised functions for getting table headers and rows as array of strings or, in the case of the rows, array of arrays of strings - these replace all of the manually written queries made inside each one of the e2e tests affected by the table refactor changes for better readability & maintainability.
- Made other minor changes to all of the e2e tests affected by the table refactor changes, such as modifying assertions for headers at the top of the files in order to be compatible with the output of the functions described above.
- The utils fixture is now passed to functions that require it - in the future we could use this to add other functions to replace logic that is overly repeated.

Ignoring the visual regression tests (because they are still flaky) all e2e tests now pass
image

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1079

### Before
N/A

### After
N/A

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -77,3 +77,28 @@ def browser_context_args(browser_context_args):
         "ignore_https_errors": True,
         "java_script_enabled": False,
     }
+
+
+class Utils:
+    @staticmethod
+    def get_page_table_headers(page: Page):
+        return page.locator(
+            "th:not(.govuk-table--invisible-on-desktop)"
+        ).evaluate_all("""els => els.map(e => e.innerText.trim())""")
+
+    @staticmethod
+    def get_page_table_rows(page: Page):
+        return page.get_by_role("row").evaluate_all(
+            """els => {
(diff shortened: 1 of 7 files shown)


e2e_tests/conftest.py
```diff
@@ -77,3 +77,28 @@ def browser_context_args(browser_context_args):
         "ignore_https_errors": True,
         "java_script_enabled": False,
     }
+
+
+class Utils:
+    @staticmethod
+    def get_page_table_headers(page: Page):
+        return page.locator(
+            "th:not(.govuk-table--invisible-on-desktop)"
```
anthonyhashemi, line 86: is it worth adding the info about these helper functions being for the desktop version, not for mobile, to the names of these functions or class.

ltrasca: @anthonyhashemi pushed a change for this

anthonyhashemi (approved):
Nice one 😄

PR #435, opened 2024-06-24
AYR-898 - Replace favicon with TNA logo

ltrasca:
## Changes in this PR
- Added a new favicon ico/svg and apple-touch-icons to the static/image/ folder
- Changed some of the paths inside ```template.html``` to use this path instead of /images/ - this is because 1) the /images/ folder is ignored by git and we can't make changes to it for the icon directly and 2) it would be overwritten by the build.sh script with new icons from the gov-uk repo each time the application is re-built
### Before
image

### After
image


app/templates/template.html
```diff
@@ -18,17 +18,17 @@
...
-                  href="{{ assetPath | default('/assets', true) }}/images/favicon.svg"
+                  href="{{ assetPath | default('/assets', true) }}/image/favicon.svg"
                   type="image/svg+xml">
             <link rel="mask-icon"
                   href="{{ assetPath | default('/assets', true) }}/images/govuk-icon-mask.svg"
                   color="{{ themeColor | default('#0b0c0c') }}">
             {#- Hardcoded value of $govuk-black #}
             <link rel="apple-touch-icon"
-                  href="{{ assetPath | default('/assets', true) }}/images/govuk-icon-180.png">
+                  href="{{ assetPath | default('/assets', true) }}/image/govuk-icon-180.png">
```
anthonyhashemi: These new icons arent the govuk-icon anymore so having them named as such may be misleading. Maybe name it something to do with tna.

On that note, I did notice we already had `image/the-national-archives-logo.svg` that we use in the webapp pages themselves, is that usuable for any of these links?

app/templates/template.html
```diff
@@ -18,17 +18,17 @@
...
                   sizes="48x48"
-                  href="{{ assetPath | default('/assets', true) }}/images/favicon.ico">
+                  href="{{ assetPath | default('/assets', true) }}/image/favicon.ico">
             <link rel="icon"
                   sizes="any"
-                  href="{{ assetPath | default('/assets', true) }}/images/favicon.svg"
+                  href="{{ assetPath | default('/assets', true) }}/image/favicon.svg"
                   type="image/svg+xml">
             <link rel="mask-icon"
                   href="{{ assetPath | default('/assets', true) }}/images/govuk-icon-mask.svg"
```
anthonyhashemi: Does this path not also need changing?

app/templates/template.html
ltrasca: I've replaced the govuk prefix for the files with just "tna", the ```image/the-national-archives-logo.svg```file looks like this, a little different to the ico
image

app/templates/template.html
ltrasca: @anthonyhashemi theoretically yes, although wasn't 100% sure what i should replace the mask-icon with, couldn't find a mask-icon on the official TNA website, I could ask Terry and see if we could perhaps reuse the ```favicon.svg``` for this purpose

app/templates/template.html
anthonyhashemi: ah okay, thanks 🙏

app/templates/template.html
anthonyhashemi: oh I see, yeah see what Terry thinks

ltrasca: @anthonyhashemi should be sorted now 🤝

anthonyhashemi (approved):

PR #443, opened 2024-06-28
AYR-998 - Reading Keycloak client secret from different specified SM store

ltrasca:
## Changes in this PR
- Inside ```AWSSecretsManagerConfig``` the function ```_get_secrets_manager_config_dict``` now takes a ```secret_id``` as an argument, allowing us to use it in multiple locations as a generic function and with multiple ids
- The ```KEYCLOAK_CLIENT_SECRET``` property inside the same class is now being overwritten in order to use the above mentioned function to retrieve the SM store for specifically the Keycloak client secret, in this case the ```secret_id``` is stored as an environment variable as ```AWS_SM_KEYCLOAK_CLIENT_SECRET_ID```
- Made changes to fix broken unit tests after the changes above
### Before
N/A
### After
N/A

### DIFF
--- a/configs/aws_secrets_manager_config.py
+++ b/configs/aws_secrets_manager_config.py
@@ -15,9 +15,11 @@ if os.environ.get("DEFAULT_AWS_PROFILE"):
 class AWSSecretsManagerConfig(BaseConfig):
     def __init__(self) -> None:
         super().__init__()
-        self.secrets_dict = self._get_secrets_manager_config_dict()
+        self.secrets_dict = self._get_secrets_manager_config_dict(
+            os.getenv("AWS_SM_CONFIG_SECRET_ID")
+        )
 
-    def _get_secrets_manager_config_dict(self):
+    def _get_secrets_manager_config_dict(self, secret_id):
         """
         Get string value of `secret_name` in Secrets Manager.
         :param key: Name of key whose value will be returned.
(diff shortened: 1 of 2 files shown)


configs/aws_secrets_manager_config.py
```diff
@@ -15,9 +15,11 @@
...
         super().__init__()
-        self.secrets_dict = self._get_secrets_manager_config_dict()
+        self.secrets_dict = self._get_secrets_manager_config_dict(
+            os.getenv("AWS_SM_CONFIG_SECRET_ID")
+        )
 
-    def _get_secrets_manager_config_dict(self):
+    def _get_secrets_manager_config_dict(self, secret_id):
         """
         Get string value of `secret_name` in Secrets Manager.
```
anthonyhashemi: worth updating this docstring

configs/aws_secrets_manager_config.py
```diff
@@ -27,12 +29,18 @@ def _get_secrets_manager_config_dict(self):
...
-        )["SecretString"]
+        secret_value_json_string = client.get_secret_value(SecretId=secret_id)[
+            "SecretString"
+        ]
         secrets_dict = json.loads(secret_value_json_string)
 
         return secrets_dict
 
     def _get_config_value(self, variable_name):
         return self.secrets_dict[variable_name]
```
anthonyhashemi, line 45: looks good. let's test it out locally in the webap and see if it works

anthonyhashemi (approved):

PR #466, opened 2024-07-12
Simpliy e2e tests

anthonyhashemi:
## Changes in this PR

- Simplify e2e tests
- Add 2 user journey navigation e2e tests


e2e_tests/test_static_links.py
```diff
@@ -0,0 +1,61 @@
...
+    Given an unauthenticated user
+    When they click a header link
+    Then the user is redirected to the associated page
+    """
+    page.goto("/")
+    page.click(f'text="{link_text}"')
+    expect(page).to_have_url(expected_url)
+
+
+SERVICE_NAME = "AYR – Access Your Records – GOV.UK"
```
colinbowen: Does this need to be defined here?

anthonyhashemi: nope was used for some old assertions I removed. good catch

colinbowen (approved):

PR #480, opened 2024-07-26
Ayr 1142/non js render

colinbowen:
## Changes in this PR
- Add Support for rendering PDF & Image (PNG/JPG) on browsers that have JS disabled

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1141

### After
Screenshot 2024-07-26 at 11 51 30

Screenshot 2024-07-26 at 11 51 45

Screenshot 2024-07-26 at 11 54 59
Screenshot 2024-07-26 at 11 54 46


app/__init__.py
```diff
@@ -48,12 +48,48 @@ def create_app(config_class, database_uri=None):
...
     csp = {
-        "default-src": f"'self' {app.config['FLASKS3_CDN_DOMAIN']}",
-        "script-src": f"'self' {app.config['FLASKS3_CDN_DOMAIN']}"
-        + " 'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='"  # pragma: allowlist secret
-        + " 'sha256-l1eTVSK8DTnK8+yloud7wZUqFrI0atVo6VlC6PJvYaQ='",  # pragma: allowlist secret
+        "default-src": f"'self' {app.config['FLASKS3_CDN_DOMAIN']} https://cdn.jsdelivr.net 'unsafe-inline' ",
+        "script-src": (
+            [
+                SELF,
+                "'unsafe-eval'",
```
ltrasca: if there is no options around using these it would be good to figure out which ones we really need and which unsafe flags we can get rid of, also if we allow unsafe eval/inline scripts we should maybe look into sanitising any query strings we use - this could maybe be in a different ticket

app/static/src/scss/includes/_record.scss
```diff
@@ -257,6 +257,95 @@
...
+}
+
+.uv .mainPanel .leftPanel {
+  border-right: 2px solid white;
+  // border-top: 2px solid white;
+  background-color: #0b0c0c !important;
+  // width: 15rem !important;
+}
+
+.uv .mainPanel {
```
ltrasca: any way we can have some of these classes under the same .uv class namespace? as in .uv { .mainPanel ... } just for the sake of it being slightly more organized

package.json
```diff
@@ -15,6 +15,7 @@
     "stylelint-selector-bem-pattern": "^3.0.1"
   },
   "dependencies": {
-    "npm": "^10.5.2"
+    "npm": "^10.5.2",
+    "universalviewer": "^4.0.25"
```
ltrasca: do we need uv as an npm dependency if we import it from a cdn?

ltrasca (commented):
just a few minor comments from me, all else looks gud

package.json
colinbowen: Yes that's right, I've removed this.

app/static/src/scss/includes/_record.scss
colinbowen: Updated to move classes under .uv class

app/__init__.py
```diff
@@ -48,12 +48,48 @@ def create_app(config_class, database_uri=None):
...
+                f"{app.config['FLASKS3_CDN_DOMAIN']}",
+                "https://cdn.jsdelivr.net",
+                "https://cdnjs.cloudflare.com",
```
anthonyhashemi: is this allowing any scripts from these domains? Is it possible to lock it down to specific urls of those we need, or do they change dynamically?

app/__init__.py
```diff
@@ -65,7 +101,16 @@ def create_app(config_class, database_uri=None):
     db.init_app(app)
     s3.init_app(app)
     compress.init_app(app)
-    talisman.init_app(app, content_security_policy=csp, force_https=force_https)
+    talisman.init_app(
+        app,
+        content_security_policy=csp,
+        force_https=force_https,
+        content_security_policy_nonce_in=[
```
anthonyhashemi, line 122: what is the purpose of this additional param. documentation, is it so we ignore worker-src and img-src? Either way, why did we need to treat them differently to the rest?

app/static/src/scss/includes/_record.scss
```diff
@@ -1,5 +1,16 @@
 /* SCSS for Record page */
 
+.govuk-tabs__list-item {
```
anthonyhashemi, line 3: I have skipped over this file as well as other css files, so worth getting someone else to review.

app/main/routes.py
```diff
@@ -708,3 +711,179 @@ def terms_of_use():
 @bp.app_errorhandler(HTTPException)
 def http_exception(error):
     return render_template(f"{error.code}.html"), error.code
+
+
+@bp.route("/generate_manifest/<uuid:record_id>")
+@access_token_sign_in_required
+def generate_manifest(record_id: uuid.UUID):
```
anthonyhashemi, line 736: does this need to be a an endpoint if we only need the response in the `record` route? If not, we could just make it a function?

app/main/routes.py
```diff
@@ -708,3 +711,179 @@ def terms_of_use():
...
+                ],
+            }
+        ],
+    }
+
+    return jsonify(manifest)
+
+
+@bp.route("/files/<record_id>")
+def get_file(record_id=None):
```
anthonyhashemi, line 865: same with this, do we have a need for calling this as an api endpoint? If not, we can just make it a function

app/templates/main/record.html
```diff
@@ -59,49 +59,111 @@
...
+                                    <h2 class="govuk-heading-m govuk-heading-m__record-header">Record arrangement</h2>
+                                    <ol class="record-arrangement-list">
+                                        {% for file_level in record["file_path"].split("/") %}
+                                            <li class="govuk-body__record-arrangement-list">{{ file_level }}</li>
+                                        {% endfor %}
+                                    </ol>
+                                </div>
+                            </div>
+                        </div>
+                        <!-- Record HTML View -->
```
anthonyhashemi: as discussed, we're going to park html view for now, so this block can be removed

app/templates/main/record.html
```diff
@@ -59,49 +59,111 @@
...
-                <div class="govuk-grid-column-one-third govuk-grid-column-one-third--record mobile-record-layout-record-arrangement">
-                    <div class="record-container">
-                        <h2 class="govuk-heading-m govuk-heading-m__record-header">Record arrangement</h2>
-                        <ol class="record-arrangement-list">
-                            {% for file_level in record["file_path"].split("/") %}
-                                <li class="govuk-body__record-arrangement-list">{{ file_level }}</li>
-                            {% endfor %}
-                        </ol>
+                        <!-- Record Details -->
+                        <div class="govuk-tabs__panel govuk-tabs__panel--hidden govuk-tabs__panel__record-details"
```
anthonyhashemi: could be worth extracting record details and record  view blocks into their own individual templates which we can include here

app/static/init.uv.js
```diff
@@ -0,0 +1,86 @@
...
+
+document.querySelectorAll(".govuk-tabs__tab").forEach((tab) => {
+  tab.addEventListener("click", function (event) {
+    if (this.getAttribute("href") === "#record-view") {
+      setTimeout(initUniversalViewer, 0);
+    }
+  });
+});
+
+// Function to remove the .attribution element
```
anthonyhashemi, line 48: why do we need to remove this element?

app/static/init.uv.js
```diff
@@ -0,0 +1,86 @@
...
+      },
+    });
+  });
+}
+
+document.addEventListener("DOMContentLoaded", function () {
+  initUniversalViewer();
+});
+
+document.querySelectorAll(".govuk-tabs__tab").forEach((tab) => {
```
anthonyhashemi, line 40: could we decouple the tab switching logic from the javascript and instead just do form submissions on clicking of the tabs? If we do that, we should then be able to reuse the tabs for when javascript is disabled also like we discussed in the demo

app/main/routes.py
anthonyhashemi: ah Im just piecing it all together now I think. Do you have documentation that explains the manifest and file url generation?

app/static/accessibility_audit_report.pdf
anthonyhashemi, line 1: what changed in the AA report pdf?

app/main/routes.py
anthonyhashemi: I think it might be useful to chat through this, as Im wondering whether we should have a separate api for this manifest

app/main/routes.py
```diff
@@ -708,3 +711,179 @@ def terms_of_use():
...
+    key = f"{file.consignment.ConsignmentReference}/{file.FileId}"
+
+    file_type = filename.split(".")[-1].lower()
+
+    try:
+        s3_response_object = s3.get_object(Bucket=bucket, Key=key)
+        file_content = s3_response_object["Body"].read()
+        file_type = filename.split(".")[-1].lower()
+
+        return send_file(
```
anthonyhashemi: is this different from the download route?

app/main/routes.py
anthonyhashemi, line 736: if we do need to keep the endpoint, would it be better to define the url as something like /record/id/manifest, like we have record/id/download etc as its not creating a new entry in db or anything like that, just buulding up a response for that record which will be consistent

app/static/init.uv.js
colinbowen: Discussed on call. Also providing a screenshot of the attribution element which is removed. (Bottom left corner toolbar which contains the filename).

Screenshot 2024-08-05 at 12 05 04

app/static/src/scss/includes/_record.scss
colinbowen: Will request another review from Lucian.

app/__init__.py
colinbowen: Removed unsafe eval and limited script-src & style-src to specific universal viewer sources

app/__init__.py
colinbowen: Reviewed on call and updated while pairing.

app/static/accessibility_audit_report.pdf
colinbowen: Added pdf back. This should not have been deleted.

app/__init__.py
colinbowen: Reviewed on call and updated while pairing. Awaiting a review from AJ

app/main/routes.py
colinbowen: This URL has been renamed to record/id/manifest. It is protected and does require authentication to be accessed.

app/main/routes.py
colinbowen: Changed the download route to include the changes for render. An request arg is used to differentiate between needing to download the file as an attachment or just return the file for render

app/templates/main/record.html
colinbowen: Removed any references to Record HTML view

app/templates/main/record-view.html
```diff
@@ -0,0 +1,33 @@
+{% if file_type == "iiif" %}
+    <div class="govuk-tabs__panel" id="record-view">
+        <div class="record-view-header">
+            <a href="{{ url_for('main.download_record', record_id=record['file_id']) }}"
+               class="govuk-button govuk-button__download--record record-view-download-btn"
+               data-module="govuk-button">Download record</a>
+        </div>
+        <div class="universal-viewer" id="viewer">
+            <script nonce="{{ csp_nonce() }}"
+                    src="https://cdn.jsdelivr.net/npm/universalviewer@4.0.25/dist/umd/UV.min.js"></script>
```
ahosgood: Would it be better to include a SRI hash in this resource and not rely on jsdelivr's auto optimisation?

```suggestion
            <script nonce="{{ csp_nonce() }}" src="https://cdn.jsdelivr.net/npm/universalviewer@4.0.25/dist/umd/UV.js" integrity="sha256-Fdp0oui1Jul9OK4Kuwm7OSbhzuuqBcgD5wi8A2tXTns=" crossorigin="anonymous"></script>
```

There is some advice on this: https://nationalarchives.github.io/engineering-handbook/third-party/cdn/

app/static/init.uv.js
```diff
@@ -0,0 +1,86 @@
...
+
+document.addEventListener("DOMContentLoaded", function () {
+  var uvElement = document.getElementById("uv");
+  if (uvElement) {
+    uvElement.style.width = "100%";
+    uvElement.style.height = "70vh";
+    uvElement.style.maxHeight = "65%";
+
+    // Apply media query for small devices
+    if (window.matchMedia("(max-width: 810px)").matches) {
```
ahosgood, line 79: What happens if the window is resized? Does this need to be reevaluated?

app/__init__.py
```diff
@@ -48,12 +48,48 @@ def create_app(config_class, database_uri=None):
...
+                "https://cdn.jsdelivr.net/npm/universalviewer@4.0.25/dist/umd/4864.b0b319b4f29542847e0e.js",
+                "https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.11.174/pdf.min.js",
+                "https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.11.174/pdf.worker.min.js",
+                "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",  # pragma: allowlist secret
+                "'sha256-l1eTVSK8DTnK8+yloud7wZUqFrI0atVo6VlC6PJvYaQ='",  # pragma: allowlist secret
+            ]
+        ),
+        "style-src": [
+            SELF,
+            "'unsafe-hashes'",
```
ahosgood: Is there any way to avoid using `unsafe-hashes` as it devalues the rest of the CSP? What events or code is being run that needs this?

app/static/init.uv.js
colinbowen: Yes this will need to be updated for resized windows

app/__init__.py
colinbowen: UniversalViewer looks to pull some CSS which requires this, currently investigating

app/templates/main/record-view.html
anthonyhashemi: I understand the benefit of adding out of interest but @ahosgood what do you mean by " not rely on jsdelivr's auto optimisation"?

app/__init__.py
anthonyhashemi: Can we just specify script-src since we only refer to `csp_nonce` in the script tags?

app/main/routes.py
```diff
@@ -629,13 +629,18 @@ def record(record_id: uuid.UUID):
...
+        file_extension=file_extension,
     )
 
 
 @bp.route("/download/<uuid:record_id>")
 @access_token_sign_in_required
 def download_record(record_id: uuid.UUID):
     file = db.session.get(File, record_id)
+    render = request.args.get("render", False)
+    print(request.args)
```
anthonyhashemi: I get this was useful for development but do we want to keep this print?

app/templates/main/record.html
```diff
@@ -59,49 +59,29 @@
...
+                                <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
+                                    <a class="govuk-tabs__tab" href="#record-view">Record View</a>
+                                </li>
+                            {% endif %}
+                            <li class="govuk-tabs__list-item">
+                                <a class="govuk-tabs__tab" href="#record-details">Record Details</a>
+                            </li>
+                        </ul>
+                        <!-- Record View -->
+                        {% include "record-view.html" %}
```
anthonyhashemi: nice, much cleaner now these are extracted out 👌

app/main/routes.py
colinbowen: Yep will remove this

app/templates/main/record-html-view.html
```diff
@@ -0,0 +1,5 @@
+<!-- Record HTML View -->
```
anthonyhashemi: can get rid of this file as we dont need to support html view

app/static/uv.min.css
```diff
@@ -0,0 +1,19 @@
+/* stylelint-disable */
```
anthonyhashemi: Ok so we went with serving the uv css file ourselves but leaving the js to coem from the external cdn.

Remind me, what benefit do we get fro serving this file ourself

app/templates/main/record-view.html
ahosgood: Universal Viewer contains the "main" file `UV.js`.

jsDelivr has a feature that automatically minifies non-minified resources. By requesting `UV.min.js`, you are asking for the version minified by the CDN rather than the original.

SRI doesn't work with files that have been optimised by jsDelivr.

image

image

app/main/routes.py
anthonyhashemi: for visibility: here https://iiif.io/guides/using_iiif_resources/ is an explanation of the need for a manifest and uri.

A manifest is the crucial resource for the iiif standard.

We are choosing to use universal viewer to render iiif resources in our webapp but by exposing this authenticated manifest url, we are giving our users the ability to go and render the manifest in another viewer / look at the raw details available in the manifest. Initially I was concerned that there could be some data here we would not want to expose but I believe that it is ok; the viewer is just a convenience; all the data is in the webapage already.

Would be worth sitting down and verifying this is the case and letting the team know that this endpoint will be available. Whether we want to advertise this endpoint is an interesting question.

app/main/routes.py
```diff
@@ -708,3 +722,31 @@ def terms_of_use():
 @bp.app_errorhandler(HTTPException)
 def http_exception(error):
     return render_template(f"{error.code}.html"), error.code
+
+
+@bp.route("/record/<uuid:record_id>/manifest")
+@access_token_sign_in_required
+def generate_manifest(record_id: uuid.UUID):
```
anthonyhashemi, line 736: would be really good to add 1 or 2 high level tests for this new endpoint and then any edge casey ones fdor manifest creation at the generate_pdf_manifest and generate_image_manifest function levels. will act as a good example of what manifests should look like

app/templates/main/record-view.html
```diff
@@ -0,0 +1,32 @@
+{% if file_type == "iiif" %}
```
anthonyhashemi, line 1: when is this file_type not iiif?

app/__init__.py
colinbowen: Set nonces to be used in script-src only 👍

app/templates/main/record-view.html
colinbowen: File type will only be iiif for files that can be rendered by iiif & UV, see:

```
def get_file_details(file):
    """Retrieve file metadata and determine file type and extension."""
    file_metadata = get_file_metadata(file.FileId)
    file_extension = file.FileName.split(".")[-1].lower()

    if file_extension in ["pdf", "png", "jpg", "jpeg"]:
        file_type = "iiif"
    else:
        file_type = None

    return file_metadata, file_type, file_extension
```

If a render isn't available, we default to loading the record details view

colinbowen: Just wanted to check that having this conditional here was ok?

app/static/init.uv.js
colinbowen: This is just to ensure that UV is init correctly on either tab switch or the initial load when viewing the record page not for any form submissions. I think we spoke about this on the call but might be worth clarifying this one.

app/__init__.py
```diff
@@ -48,12 +48,62 @@ def create_app(config_class, database_uri=None):
...
+                "https://cdn.jsdelivr.net/npm/universalviewer@4.0.25/dist/umd/4864.b0b319b4f29542847e0e.js",
+                "https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.11.174/pdf.min.js",
+                "https://cdnjs.cloudflare.com/ajax/libs/pdf.js/3.11.174/pdf.worker.min.js",
+                "'sha256-GUQ5ad8JK5KmEWmROf3LZd9ge94daqNvd8xy9YS1iDw='",  # pragma: allowlist secret
+                "'sha256-l1eTVSK8DTnK8+yloud7wZUqFrI0atVo6VlC6PJvYaQ='",  # pragma: allowlist secret
+            ]
+        ),
+        "style-src": [
+            SELF,
+            "'sha256-aqNNdDLnnrDOnTNdkJpYlAxKVJtLt9CtFLklmInuUAE='",  # pragma: allowlist secret
```
ahosgood, line 73: How easy are these going to be to maintain?

Is this relevant or have we tested it? [Firefox does not support hash-based CSP](https://bugzilla.mozilla.org/show_bug.cgi?id=1409200)

colinbowen: These hashes should be the same and shouldn't really change going forward as they are the hash of each style required by UV. We're also not using strict-dynamic in our CSP and I've done some testing in Firefox where the PDF is loaded correctly.

app/static/uv.min.css
colinbowen: There's no difference to serving this file ourself as it was previously set to a specific version from the CDN. This changed during testing / development to try and debug the issues with the CSP. For the sake of simplicity both files are being pulled from the cdn.

ltrasca:
The tabs container sticks out a little bit on each side compared to the elements above it (e.g. breadcrumbs and header elements), any way this could be fixed in this pr?

image

colinbowen:
> The tabs container sticks out a little bit on each side compared to the elements above it (e.g. breadcrumbs and header elements), any way this could be fixed in this pr?
> 
> image

Moved the tabs-content to a govuk width container

app/__init__.py
ahosgood: So everything is good as long as we continue to use the same version of UV?

ltrasca (dismissed):

app/__init__.py
colinbowen: Yes that's correct. So the CSP hashes in either script-src or style-src are to allow certain styles or inline scripts. See: https://content-security-policy.com/hash/

app/templates/main/record-view.html
anthonyhashemi: sorry for the late response, I get it now. I assume @ltrasca expands this in his work to show the we don't support your document for render ticket?

app/static/init.uv.js
anthonyhashemi: Its been a while since the call so yeah maybe wortth catching up again but my thoughts are, if we support the tab switching with non javascript then we'd need to do it via form submissions and thus wouldnt need javascript to do it.

colinbowen: Check with Terry that UI side of changes is ok

app/static/init.uv.js
colinbowen: Will ensure for full responsiveness after this has been merged. Ticket to track: https://national-archives.atlassian.net/browse/AYR-1198

app/main/routes.py
colinbowen: Discussed with AYR, ok to leave this protected route in.

colinbowen: Discussed on call on why this route is required for Universal Viewer. 
- UV requires a URL to access the manifest file from in JSON format
- Manifest data can be created and is behind a protected route

app/main/util/render_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    file_name = file.FileName
+    file_url = url_for(
+        "main.download_record", record_id=record_id, _external=True
+    )
+
+    manifest = {
+        "@context": [
+            "http://iiif.io/api/presentation/3/context.json",
+        ],
+        "id": f"{url_for('main.download_record', record_id=record_id, _external=True, render=True)}",
```
anthonyhashemi: I just realised that id should be the url of the manifest online, not the url of the file. https://iiif.io/api/cookbook/recipe/0001-mvm-image/

`The id property identifies this manifest with the URL at which it is available online.`

app/main/util/render_utils.py
```diff
@@ -0,0 +1,220 @@
...
+    image = Image.open(io.BytesIO(file_content))
+    width, height = image.size
+
+    file_url = url_for(
+        "main.download_record", record_id=record_id, _external=True, render=True
+    )
+
+    manifest = {
+        "@context": "http://iiif.io/api/presentation/2/context.json",
+        "@id": f"{url_for('main.download_record', record_id=record_id, _external=True)}",
```
anthonyhashemi: same here

app/main/routes.py
colinbowen: Added PDF generation & image generation manifest tests to ensure correct manifest is produced.

anthonyhashemi (dismissed):
Spoke on call. Approving this PR to unblock accesibility testing but currently we cant allow a user without download permission to view a render. We will work on fixing this in a separate PR ASAP.

anthonyhashemi (approved):

PR #497, opened 2024-08-09
AYR-1121 - Add support for download group permissions 

ltrasca:
## Changes in this PR

- Added a new property inside ayr_user.py that checks if a user is in the appropriate group to download records
- Inside the record template we only render the download button section when the user has passes the condition outlined above
- Made some small changes to existing unit tests for the record endpoint in order to reduce repetition of code & strings
- Along side the tests for checking a standard user can see the download record button there are now tests for all access users as well
- Added tests to check that standard and all access users without the download group are not able to view the download record button
- Inside of the download record endpoint, users that are not in the download group now get a 403 response (forbidden)
- Added tests for the change above to the download tests
- Added 403 page template (Flask throws a fit when this doesnt exist and we do abort(403))
- Changed e2e tests slightly so users get created with the download group by default in order to not break any existing tests

### Before
image

### After
image

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -593,6 +593,8 @@ def record(record_id: uuid.UUID):
     """
     form = SearchForm()
     file = db.session.get(File, record_id)
+    ayr_user = AYRUser(session.get("user_groups"))
+    can_download_records = ayr_user.can_download_records
 
     if file is None:
         abort(404)
@@ -628,6 +630,7 @@ def record(record_id: uuid.UUID):
         record=file_metadata,
         breadcrumb_values=breadcrumb_values,
         download_filename=download_filename,
+        can_download_records=can_download_records,
         filters={},
     )
 
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -643,7 +643,7 @@ def download_record(record_id: uuid.UUID):
     can_download_records = ayr_user.can_download_records
 
     if can_download_records is False:
-        abort(401)
+        abort(403)
 
     if file is None:
         abort(404)
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -639,6 +639,11 @@ def record(record_id: uuid.UUID):
 @access_token_sign_in_required
 def download_record(record_id: uuid.UUID):
     file = db.session.get(File, record_id)
+    ayr_user = AYRUser(session.get("user_groups"))
+    can_download_records = ayr_user.can_download_records
+
+    if can_download_records is False:
+        abort(401)
 
     if file is None:
         abort(404)
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -642,7 +642,7 @@ def download_record(record_id: uuid.UUID):
     ayr_user = AYRUser(session.get("user_groups"))
     can_download_records = ayr_user.can_download_records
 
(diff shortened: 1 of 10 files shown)


app/main/routes.py
```diff
@@ -636,6 +639,11 @@ def record(record_id: uuid.UUID):
 @access_token_sign_in_required
 def download_record(record_id: uuid.UUID):
     file = db.session.get(File, record_id)
+    ayr_user = AYRUser(session.get("user_groups"))
+    can_download_records = ayr_user.can_download_records
+
+    if can_download_records is False:
+        abort(401)
```
anthonyhashemi: I think technically this should be a 403 as the user is authenticated but just is forbidden from this.

ltrasca: yup I agree should be 403, will change

app/main/routes.py
```diff
@@ -636,6 +639,11 @@ def record(record_id: uuid.UUID):
 @access_token_sign_in_required
 def download_record(record_id: uuid.UUID):
     file = db.session.get(File, record_id)
+    ayr_user = AYRUser(session.get("user_groups"))
+    can_download_records = ayr_user.can_download_records
+
+    if can_download_records is False:
```
anthonyhashemi: I think `if not can_download_records` or even `if can_download_records is not True` might be safer, just in case it comes is null or otherwise for whatever reason

e2e_tests/conftest.py
```diff
@@ -127,6 +127,7 @@ def create_aau_keycloak_user(keycloak_admin, create_keycloak_user):
 def create_standard_keycloak_user(keycloak_admin, create_keycloak_user):
     user_groups = [
         "/ayr_user_type/view_dept",
+        "/ayr_user_type/download",
```
anthonyhashemi, line 163: I think its worth us adding  end to end test for a happy path and unhappy path for the button and download route to show that we have configured thiings correctly e2e, since the other tests mock keycloak groups

app/tests/test_record.py
```diff
@@ -232,26 +295,89 @@ def test_record_download_record_with_citeable_reference(
...
+
+        expected_download_html = (
+            expected_download_html_without_citeable_reference(file.FileId)
+        )
 
         assert_contains_html(
             expected_download_html, html, "div", {"class": "rights-container"}
         )
 
+    def test_record_aau_user_with_perms_can_download_record_with_citeable_reference(
```
anthonyhashemi: not sure if we need this extra test at this level, seems quite edge casey, as its just about the citable ref and we have tests for that already

e2e_tests/test_record.py
```diff
@@ -97,7 +97,81 @@ def test_record_metadata_with_closed_record_status(
...
+        Then the download button is NOT visible
+        """
+        record_id = "100251bb-5b93-48a9-953f-ad5bd9abfbdc"
+        standard_user_page.goto(f"{self.route_url}/{record_id}")
+
+        button = standard_user_page.get_by_role("link", name="Download record")
+
+        expect(button).to_be_hidden()
+
+    def test_record_download_record(
```
anthonyhashemi, line 172: can we get one test where user without download gets the 403 response

anthonyhashemi (approved):
nice one 👌

PR #515, opened 2024-08-21
AYR-1183 - Alert banner component & unsupported file handle

ltrasca:
## Changes in this PR

- Added a new alert banner component inside a new directory called "components", along with styling
- Modified some record view logic to display the banner whenever a user requests an unsupported file type
- Added unit tests to check the changes mentioned above

### Before
image

### After
image


app/static/src/scss/includes/components/_alert-banner.scss
```diff
@@ -0,0 +1,49 @@
...
+  padding: 13px 21px;
+  box-sizing: border-box;
+  text-align: left;
+  font-family: "GDS Transport", sans-serif;
+
+  // default bg and border (info level)
+  background-color: #dbeff9;
+  border-left: 10px solid #2b8cc4;
+
+  // todo: make a dedicated file for mixins
```
anthonyhashemi, line 19: should we do this now?

app/tests/conftest.py
```diff
@@ -1098,11 +1098,78 @@ def record_files():
         ),
     }
 
+    # file with format thats not supported by IIIF
```
anthonyhashemi: I feel like the tests in general (not just the ones in this PR) might be simplified by using more specific fixtures or direct setup in the tests themselves so it is easier to show that the setup is correct. I think we see the issue with having this abstract bulk record_files fixture show itself in the tests where we need to add comments to the code to explain that nth file has particular properties.

Not necessarily for this PR, but just so we keep it in mind going forward

app/tests/test_record.py
```diff
@@ -251,6 +251,66 @@ def test_record_record_arrangement(
...
+        """
+        Given a File in the database
+        When a standard user with request to view the record details page
+        Then the response status code should be 200
+        If the extension of the File requested is not compatible with IIIF
+        Then the alert banner responsible with alerting the user should be visible
+        """
+
+        # index 6 is .foobar
+        file = record_files[6]["file_object"]
```
anthonyhashemi: for example here we just need one file, so dont necessarily need to reference record_files

app/tests/test_record.py
```diff
@@ -251,6 +251,66 @@ def test_record_record_arrangement(
...
+        """
+        Given a File in the database
+        When a standard user with request to view the record details page
+        Then the response status code should be 200
+        If the extension of the File requested is compatible with IIIF
+        Then the alert banner responsible with alerting the user should NOT be visible
+        """
+
+        # index 1 is .pdf
+        file = record_files[1]["file_object"]
```
anthonyhashemi: same here

e2e_tests/test_record.py
```diff
@@ -169,6 +169,44 @@ def test_record_standard_users_without_perms_cant_see_download_button(
 
         expect(button).to_be_hidden()
 
+    def test_record_alert_banner_is_visible_with_unsupported_file_extension(
```
anthonyhashemi: I've been thinking about how we can keep e2e tests minimal to just critical paths and for things that rely upon external api which we mock at the lower level.

I know we are mocking the s3 bucket in the lower level tests but we do have other e2e tests which show that the integration of s3 works correctly for download, for example, so I feel like the logic we have added here is sufficiently tested at the lower level.

What do you think?

ltrasca: I agree 👍

app/tests/conftest.py
ltrasca: definitely, I feel like this issue could be solved by not returning the files as an array from the fixture, maybe key value pairs where the key is descriptive about what the file is or what its purpose it - for this ticket I'll make some fixtures that have descriptive names but I'll make a ticket to look into this further

anthonyhashemi (approved):

PR #538, opened 2024-09-06
AYR-1001/performance-test-setup

colinbowen:
## Changes in this PR
- Init locust for performance tests
- Modify access token sign in decorator to allow for sign in via an access token generated via API
- Add PERF_TEST env variable

## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-1001

## Screenshots of UI changes
N/A

### After


README.md
```diff
@@ -192,6 +192,7 @@ Properties configurable at runtime:
 - `FLASKS3_ACTIVE`: whether to fetch static assets from s3/Cloudfront rather than the usual `url_for`.
 - `FLASKS3_CDN_DOMAIN`: CDN domain to fetch assets from if `FLASKS3_ACTIVE` is set to `True`
 - `FLASKS3_BUCKET_NAME`: S3 bucket assets are uploaded to and served to Cloudfront from.
+- `PERF_TEST`: Enable to allow access tokens generated via API to be accepted for performance testing.
```
ltrasca, line 195: could this env be added to the .env template?

performance_tests/locustfile.py
```diff
@@ -0,0 +1,106 @@
...
+    wait_time = between(1, 5)
+
+    def on_start(self):
+        self.token, self.refresh_token = self.authenticate()
+
+    def authenticate(self):
+        token_url = KEYCLOAK_AUTH_URL
+        client_id = KEYCLOAK_CLIENT_ID
+        client_secret = KEYCLOAK_CLIENT_SECRET
+        username = AYR_AAU_USER_USERNAME
```
ltrasca, line 54: ideally we should create username/password on the fly similarly to the e2e tests and just use those credentials in order to get the token via password grant type, otherwise would we need a specific test account or our developer accounts not to have 2fa enabled?

(see e2e_tests/conftest.py line 91)

performance_tests/locustfile.py
```diff
@@ -0,0 +1,106 @@
+import secrets
+
+from locust import HttpUser, between, task
+
+CERT_PATH = False
+AYR_AAU_USER_USERNAME = ""
+AYR_AAU_USER_PASSWORD = ""
+KEYCLOAK_AUTH_URL = ""
```
ltrasca: would these be picked up from env variables as opposed to declared locally?

app/main/authorize/access_token_sign_in_required.py
```diff
@@ -82,8 +107,13 @@ def decorated_view(*args, **kwargs):
 def _validate_or_refresh_tokens(access_token, refresh_token):
     tokens_are_refreshed = False
 
-    if not (access_token and refresh_token):
-        raise InvalidAccessToken
+    # PERFORMANCE TEST MODE
+    if os.getenv("PERF_TEST", "False") == "True":
```
ltrasca, line 100: An idea: we could have a generic 'test mode' where endpoints accept an authorization header in the form of the token, this way we don't have to embed logic or new env variables for each test mode (e.g. accessibility tests in the future). Overall, this is a really good piece of functionality, its gonna make our lives much easier testing pages.

performance_tests/locustfile.py
```diff
@@ -0,0 +1,106 @@
...
+            )
+            return None
+
+    @task
+    def browse_pages(self):
+        if self.token:
+            headers = {"Authorization": f"Bearer {self.token}"}
+
+            urls = [
+                "/browse/transferring_body/c3e3fd83-4d52-4638-a085-1f4e4e4dfa50",
```
ltrasca, line 84: I see these are hardcorded urls, which would mean we depend on objects we currently have in our database for the tests (I think this has been a slight problem in our unit and e2e tests at times), these might get deleted, modified, replaced in the future, would there be any way we can mock these?

ltrasca (commented):
few comments from me, overall good work!

README.md
colinbowen: Yes good point, will add it.

performance_tests/locustfile.py
colinbowen: This is worth looking into. As part of this we could generate some test data which then used to benchmark the site using locust

performance_tests/locustfile.py
colinbowen: Yes this does require a specific account with 2FA disabled. There is a test user which could be kept specifically for test purposes or created on the fly without 2FA.

app/main/authorize/access_token_sign_in_required.py
colinbowen: Created: https://national-archives.atlassian.net/browse/AYR-1223

performance_tests/locustfile.py
colinbowen: Moved to be read from ENV vars

ltrasca (approved):

PR #552, opened 2024-09-18
Use presigned URLs for render fallback

colinbowen:
## Changes in this PR
- Revise render functionality to use presigned URLs for render fallback

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1208

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -1,5 +1,6 @@
 import io
 import json
+import os
 import uuid
 
 import boto3
@@ -10,6 +11,7 @@ from flask import (
     render_template,
     request,
     send_file,
+    send_from_directory,
     session,
     url_for,
 )
@@ -48,7 +50,7 @@ from app.main.util.render_utils import (
     get_download_filename,
     get_file_details,
     get_file_mimetype,
-    manage_static_file,
+    write_temporary_file,
 )
 
 from .forms import SearchForm
@@ -620,7 +622,7 @@ def record(record_id: uuid.UUID):
     )
 
     try:
-        manage_static_file(file, file_extension)
+        write_temporary_file(file, file_extension)
     except Exception as e:
         current_app.logger.info(f"Error with file IO: {e}")
 
@@ -757,3 +759,14 @@ def generate_manifest(record_id: uuid.UUID):
         return generate_image_manifest(s3_file_object, record_id)
     else:
         return http_exception(BadRequest())
+
+
+@access_token_sign_in_required
+@bp.route("/files/<filename>")
+def serve_file(filename):
+    temp_directory = session.get("temp_dir")
+    files_directory = os.path.join(temp_directory, "files")
+    if temp_directory:
+        return send_from_directory(files_directory, filename)
+    else:
+        abort(404)
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -44,13 +44,13 @@ from app.main.util.filter_sort_builder import (
     build_sorting_orders,
 )
 from app.main.util.render_utils import (
+    create_presigned_url,
     generate_breadcrumb_values,
     generate_image_manifest,
     generate_pdf_manifest,
     get_download_filename,
     get_file_details,
     get_file_mimetype,
-    write_temporary_file,
 )
 
 from .forms import SearchForm
@@ -605,6 +605,7 @@ def record(record_id: uuid.UUID):
     file = db.session.get(File, record_id)
     ayr_user = AYRUser(session.get("user_groups"))
     can_download_records = ayr_user.can_download_records
+    presigned_url = None
 
     if file is None:
         abort(404)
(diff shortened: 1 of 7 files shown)


app/main/util/render_utils.py
```diff
@@ -57,36 +58,37 @@ def get_download_filename(file):
 
 
 def manage_static_file(file, file_extension):
-    """Manage the file in the static directory."""
+    """Manage the file in a temporary directory."""
```
anthonyhashemi: It took me a while to figure out the purpose of the manage_static function but It seems like the function is writing the file stream from s3 object to a temp file on the lambda. Im assuming so the file can be served by the html template / universal viewer which assues the file exists there?

Could we use the returned path and pass it to the render_teplate which we can pass down to record.html so that we dont have to hardcode

                {% if file_extension == "pdf" %}
                    <object data="{{ url_for('static', filename='files/temp_file.pdf') }}"
                            width="100%"
                            height="700px">
                        <p>Please enable JS to view this record.</p>
                    </object>
                {% elif file_extension == "png" or file_extension == "jpg" %}
                    <object data="{{ url_for('static', filename='files/temp_file.png') }}"

that way we dont couple this route to the html

Also currently Id assume the template needs to change anyway as it is still hardcoded to static, not temp?

Could we also rename the function to something more transparent like write_temporary_file or similar

app/main/routes.py
```diff
@@ -620,7 +621,7 @@ def record(record_id: uuid.UUID):
     )
 
     try:
-        manage_static_file(file, file_extension)
+        presigned_url = create_presigned_url(file)
     except Exception as e:
```
anthonyhashemi, line 625: I think it would be worth making the error message more specific,something like "Failed to create presigned url for document render non-javascript fallback" and also logging the original error message/stacktrace

app/main/util/render_utils.py
```diff
@@ -56,37 +54,21 @@ def get_download_filename(file):
...
-    file_content = s3_file_object["Body"].read()
+    try:
+        presigned_url = s3.generate_presigned_url(
+            "get_object", Params={"Bucket": bucket, "Key": key}, ExpiresIn=10
+        )
 
-    with open(static_file_path, "wb") as static_file:
-        static_file.write(file_content)
+    except Exception as e:
+        current_app.logger.error(f"S3 error: {e}")
```
anthonyhashemi: oh I see we abort now rather than letting it continue. 

In that case the try catch I commented on where we call this function is redundant

colinbowen: I think changing this error message to the one you recommended would be useful

app/main/util/render_utils.py
```diff
@@ -56,37 +54,21 @@ def get_download_filename(file):
     return None
 
 
-def manage_static_file(file, file_extension):
-    """Manage the file in the static directory."""
+def create_presigned_url(file):
```
anthonyhashemi, line 57: since we are not using this url if extension not pdf or jpg, at the moment, we can lock down files even more by only creating urls for those. We should probably define a constant for the supported extension types which we can pass through to this function and the templates

app/main/util/render_utils.py
```diff
@@ -56,37 +54,28 @@ def get_download_filename(file):
...
-    file_content = s3_file_object["Body"].read()
+    try:
+        presigned_url = s3.generate_presigned_url(
+            "get_object", Params={"Bucket": bucket, "Key": key}, ExpiresIn=10
+        )
 
-    with open(static_file_path, "wb") as static_file:
-        static_file.write(file_content)
+    except Exception as e:
+        current_app.logger.info(f"Failed to create presigned url. {e}")
```
anthonyhashemi: since we have the catch at the higher level, do we need this here? the messages seem to be pretty similar, so having 2 different messages may be confusing?

And if any of the rest of this function fails we currently error silently (with the exception at the higher level), whereas if `s3.generate_presigned_url(` errors, we abort  with 404

colinbowen: Changed to only keep the try except at the routes.py level

anthonyhashemi (dismissed):

app/main/util/render_utils.py
colinbowen: Added SUPPORTED_RENDER_EXTENSIONS to config 👍

app/main/util/render_utils.py
```diff
@@ -97,7 +81,7 @@ def generate_pdf_manifest(record_id):
 
     file_name = file.FileName
     file_url = url_for(
-        "main.download_record", record_id=record_id, _external=True
+        "main.download_record", record_id=record_id, _external=True, render=True
```
anthonyhashemi, line 84: whats this change for? to address https://national-archives.atlassian.net/browse/AYR-1212?

colinbowen: Yes, the manifest should have had these arguments. generate_image_manifest already had this arg set to true.

anthonyhashemi: 👌

anthonyhashemi (approved):

PR #554, opened 2024-09-18
AYR-1200 - Enhance search with OpenSearch

ltrasca:
## Changes in this PR
- /search/transferring_body/<uuid:_id> endpoint enhanced to use OpenSearch by @anthonyhashemi 
- /local_services/docker-compose.yml used to start up a Docker container of OpenSearch added by @anthonyhashemi 
- rewrote the pagination into a new component to be compatible with utilities that are separate to Alchemy
- modified the search summary to include an implementation for OpenSearch
- added new pagination component along with its corresponding logic to the search summary page
- modified unit tests for the search endpoints and added unit tests for the new pagination utilities

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1200

## Screenshots of UI changes

N/A

- [X] Requires env variable(s) to be updated
Env variables

OPEN_SEARCH_HOST=foo
OPEN_SEARCH_USERNAME=bar
OPEN_SEARCH_PASSWORD=example

are now required in order for the search endpoint to use OpenSearch


app/main/routes.py
```diff
@@ -556,17 +559,42 @@ def search_transferring_body(_id: uuid.UUID):
...
+                "bool": {
+                    "must": [
+                        {
+                            "multi_match": {
+                                "query": query,
+                                "fields": ["*"],
+                            }
+                        }
+                    ],
+                    # "filter": [
```
anthonyhashemi: we do need this uncommented and pass in the transferring_body_id since this route is searching within a transferring body

app/main/routes.py
```diff
@@ -30,7 +31,6 @@
     build_browse_query,
     build_browse_series_query,
     build_fuzzy_search_summary_query,
-    build_fuzzy_search_transferring_body_query,
```
anthonyhashemi, line 33: this function itself can be deleted now alongside its tests

configs/base_config.py
```diff
@@ -93,6 +93,17 @@ def FLASKS3_CDN_DOMAIN(self):
     def FLASKS3_BUCKET_NAME(self):
         return self._get_config_value("FLASKS3_BUCKET_NAME")
 
+    @property
+    def OPEN_SEARCH_HOST(self):
+        return self._get_config_value("OPEN_SEARCH_HOST")
+
+    @property
+    def OPEN_SEARCH_HTTP_AUTH(self):
```
anthonyhashemi, line 101: this is fine for local dev but to get this working with AWS we will need to use iam so will need to override this method in 
`configs/aws_secrets_manager_config.py`

so that it uses something like

```
    from requests_aws4auth import AWS4Auth

    sts_client = boto3.client('sts')
    assumed_role = sts_client.assume_role(
        RoleArn=self._get_config_value("OPEN_SEARCH_MASTER_ROLE_ARN"),
        RoleSessionName="AYRWebappLambdaSession"
    )
    credentials = assumed_role['Credentials']
    open_search_http_auth = AWS4Auth(
        credentials['AccessKeyId'],
        credentials['SecretAccessKey'],
        self.AWS_REGION,
        'es',
        session_token=credentials['SessionToken']
    )
```

app/main/util/pagination.py
```diff
@@ -0,0 +1,63 @@
+def get_pagination(current_page: int, total_pages: int) -> dict | None:
```
anthonyhashemi, line 1: looks good at a glance. I think this function would benefit from some unit tests to cover all the main cases:

- only 1 page
- more than 1 but less than 4
- no previous page
- no next page
- elipses before, no elipses after
- elises before and after
- no elipses before, elipses after

and maybe some others that would be useful

app/templates/main/search-transferring-body.html
```diff
@@ -90,42 +90,47 @@ <h2 class="govuk-body browse__body">You are viewing</h2>
...
+                                                    {% endif %}
+                                                {% endfor %}
+                                            {% endif %}
                                         </div>
                                     </tbody>
                                 </table>
                                 <!-- PAGINATION -->
                                 {% with view_name='main.search_transferring_body', id = request.view_args['_id'] %}
-                                    {% include "pagination.html" %}
+                                    {% include "pagination-new.html" %}
```
anthonyhashemi, line 133: can we use the new pagination for all the routes and justget rid of the old one? It can wait for a separate PR but just checking that the logic should be universal right?

app/tests/test_search.py
```diff
@@ -516,259 +586,33 @@ def test_search_transferring_body_with_table_data_links(
...
-        assert [anchor["href"] for anchor in anchors] == expected_anchors_hrefs
-        assert [
-            anchor.get_text() for anchor in anchors
-        ] == expected_anchors_text
-        assert (
-            list(filter(None, [row.get_text(strip=True) for row in rows]))
-            == expected_all_rows_text
-        )
-
-    @pytest.mark.parametrize(
```
anthonyhashemi, line 549: I remember we had lots of potentially redundant/repetitive tests due to testing pagination logic at a high level. Im assuming thats why weve removed soe of these tests, but I think it'll be worth jumping on a call later to align

app/main/routes.py
anthonyhashemi, line 1: as discussed offline, we'll also need to update the search_results_sumary route to use opensearch too. The main difference wit that query would be we need to aggregate over transferring bodies. Will need to look into opensearch query documentations on how to do this.

Could be done in a separate PR though to keep this one simpler and just focused on the search_transf_body route

configs/base_config.py
anthonyhashemi: we will test this on an aws deployment to mmake sure its working

app/tests/test_pagination.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def test_no_next_when_last_page(self):
+        """
+        Test that when current page is the last previous is correct while next is None
+        """
+        result = get_pagination(100, 100)
+        assert result["previous"] == 99
+        assert result["next"] is None
+
+    @pytest.mark.parametrize(
```
anthonyhashemi, line 83: nice 👍

app/tests/test_config.py
```diff
@@ -161,31 +201,43 @@ def test_aws_secrets_manager_config_initialized(monkeypatch):
...
+        )
+        assert config.DEFAULT_PAGE_SIZE == "test_default_page_size"
+        assert config.DEFAULT_DATE_FORMAT == "test_default_date_format"
+
+        assert config.RECORD_BUCKET_NAME == "test_record_bucket_name"
+        assert config.FLASKS3_ACTIVE is False
+        assert config.FLASKS3_CDN_DOMAIN == "test_flasks3_cdn_domain"
+        assert config.FLASKS3_BUCKET_NAME == "test_flasks3_bucket_name"
+        assert config.PERF_TEST is False
+        assert isinstance(aws_auth, AWS4Auth)
```
anthonyhashemi, line 229: 😎

app/main/routes.py
```diff
@@ -474,18 +473,60 @@ def search_results_summary():
...
+        dsl_query = {
+            "query": {
+                "bool": {
+                    "must": [
+                        {"multi_match": {"query": query, "fields": ["*"]}}
+                    ],
+                },
+            },
+            "aggs": {
+                "aggregate_by_transferring_body": {
```
anthonyhashemi, line 513: nice one 👌

app/main/routes.py
```diff
@@ -474,18 +473,60 @@ def search_results_summary():
...
+                        }
+                    },
+                }
+            },
+        }
+
+        size = per_page
+        page_number = page
+        from_ = size * (page_number - 1)
+        search_results = open_search.search(dsl_query, from_=from_, size=size)
```
anthonyhashemi, line 520: I think we need to adjust/ add tests for this route too now

app/tests/test_search.py
```diff
@@ -211,29 +231,81 @@ def test_search_results_summary_top_search(
...
     def test_search_results_summary_no_query(
-        self, client: FlaskClient, mock_all_access_user
+        self, mock_search_client, client: FlaskClient, mock_all_access_user
     ):
         """
         Given an all_access_user accessing the search results summary page
         When they make a GET request without a query
         Then they should not see any results on the page.
         """
+        mock_search_client.return_value = MockOpenSearch(
```
anthonyhashemi: I think this mock is not necessary for this test, as this is the scenario where there is no query and so opensearch.query is not called.

app/tests/test_config.py
```diff
@@ -136,6 +172,10 @@ def test_aws_secrets_manager_config_initialized(monkeypatch):
             "DB_NAME": "test_db_name",
             "DB_SSL_ROOT_CERTIFICATE": "test_db_ssl_root_certificate",
             "DEFAULT_PAGE_SIZE": "test_default_page_size",
+            "OPEN_SEARCH_MASTER_ROLE_ARN": "test_master_role_arn",
+            "OPEN_SEARCH_HOST": "test_os_host",  # pragma: allowlist secret
+            "OPEN_SEARCH_USERNAME": "test_os_username",
```
anthonyhashemi: username and password dont need to be set in the secret as we build the AWS4Auth object for aws

anthonyhashemi (dismissed):

configs/base_config.py
anthonyhashemi: tested and works in AWS

app/tests/test_config.py
```diff
@@ -161,31 +194,43 @@ def test_aws_secrets_manager_config_initialized(monkeypatch):
...
+
+        assert config.RECORD_BUCKET_NAME == "test_record_bucket_name"
+        assert config.FLASKS3_ACTIVE is False
+        assert config.FLASKS3_CDN_DOMAIN == "test_flasks3_cdn_domain"
+        assert config.FLASKS3_BUCKET_NAME == "test_flasks3_bucket_name"
+        assert config.PERF_TEST is False
+        assert isinstance(aws_auth, AWS4Auth)
+        assert aws_auth.access_id == "test_access_key"
+        assert aws_auth.region == "test_aws_region"
+        assert aws_auth.service == "es"
```
anthonyhashemi, line 232: found we can add 

```
            assert (
                aws_auth.signing_key.secret_key
                == "test_secret_key"  # pragma: allowlist secret
            )
```

app/templates/main/search-transferring-body.html
```diff
@@ -11,16 +11,16 @@
     {%- set _ = breadcrumbs.update({'search_terms': breadcrumb_values[3]["search_terms"]}) -%}
 {% endif %}
 {% set sorting_list = {
-    "Series reference (A to Z)": "series-asc",
-    "Series reference (Z to A)": "series-desc",
+    "Series reference (A to Z)": "series_id-asc",
```
anthonyhashemi, line 14: does this change mean sorting is working with opensearch in this PR?

app/main/util/filter_sort_builder.py
```diff
@@ -38,6 +38,23 @@ def build_sorting_orders(args):
     return sorting_orders
 
 
+def build_sorting_orders_open_search(args):
```
anthonyhashemi, line 41: is this functioning then?

app/main/routes.py
```diff
@@ -556,17 +615,42 @@ def search_transferring_body(_id: uuid.UUID):
...
+                                "fields": ["*"],
+                                "query": query,
+                                "fuzziness": 1,
+                            }
+                        }
+                    ],
+                    "filter": [{"term": {"transferring_body_id.keyword": _id}}],
+                }
+            },
+            # "sort": sorting_orders,
```
anthonyhashemi, line 641: oh ok so we are building the orders but still need to work on it a bit before applying it

app/main/routes.py
```diff
@@ -473,18 +478,69 @@ def search_results_summary():
...
+
+        size = per_page
+        from_ = size * (page - 1)
+        search_results = open_search.search(dsl_query, from_=from_, size=size)
+        results = search_results["aggregations"][
+            "aggregate_by_transferring_body"
+        ]["buckets"]
+
+        total_records = 0
+        for bucket in results:
```
anthonyhashemi, line 537: do we have tests covering this?

app/templates/main/search-transferring-body.html
ltrasca: was just me tinkering around with it haha

app/main/util/filter_sort_builder.py
ltrasca: the function itself and the query works, the terms at times differ (e.g. it sometimes needs to be foobar.keyword, sometimes metadata.date because its nested in metadata), which makes it slightly more complicated after that point and I'd like to make it slightly cleaner, there is also some logic for passing the template the current filter value thats broken

app/main/routes.py
ltrasca: we didnt before but I've added one

app/tests/test_search.py
```diff
@@ -288,6 +288,67 @@ def test_search_results_summary_with_no_results(
             {"class": "govuk-list govuk-list--bullet"},
         )
 
+    @patch("app.main.routes.OpenSearch")
```
anthonyhashemi, line 291: noice

anthonyhashemi (approved):
Looks good, as discussed offline, feature parity of sorting of search results will be complete in a followup PR

PR #556, opened 2024-09-20
Add text extract and index lambda handler

anthonyhashemi:
## Changes in this PR

- Added `lambda_handler` for extracting data from postgres db and text content fro s3 bucket and indexing it in opensearch
- Added `index_file_content_and_metadata_in_opensearch` which is used by the lambda_handler that can be used independently of aws. 
- Dockerfile and requirements.txt to run the lambda in

## JIRA ticket

[AYR-1199](https://national-archives.atlassian.net/browse/AYR-1199?atlOrigin=eyJpIjoiZWZhMzgyNThhZDhjNDFjODgxNmVhMzU4YTk5N2M3ZjIiLCJwIjoiaiJ9)

[AYR-1199]: https://national-archives.atlassian.net/browse/AYR-1199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


data_management/opensearch/tests/test_process_and_index_file.py
```diff
@@ -0,0 +1,197 @@
...
+from sqlalchemy.orm import declarative_base, relationship, sessionmaker
+
+from data_management.opensearch.index_file_content_and_metadata_in_opensearch import (
+    index_file_content_and_metadata_in_opensearch,
+)
+
+Base = declarative_base()
+
+
+class Body(Base):
```
anthonyhashemi: I'd love to avoid redefining this here as we have apppp/models.py but we use flask-sqlalchemy in there. We will need to do some work to make models.py look like this , i.e not flask specific but that might cause some issues in our queries in the webapp. Can create a techdebt ticket for the backlog.

data_management/opensearch/index_file_content_and_metadata_in_opensearch.py
```diff
@@ -0,0 +1,132 @@
...
+) -> None:
+    open_search = OpenSearch(
+        open_search_host_url,
+        http_auth=open_search_http_auth,
+        use_ssl=True,
+        verify_certs=False,
+        connection_class=RequestsHttpConnection,
+    )
+    response = open_search.index(index="documents", id=file_id, body=document)
+    logger.info(response)
```
anthonyhashemi: double check this doesnt log any sensitive data

anthonyhashemi: wasnt sensitive but removed anyway as unnecessary as we already get logs from the opensearch client saying it was a successful request or not

data_management/opensearch/index_file_content_and_metadata_in_opensearch.py
```diff
@@ -0,0 +1,131 @@
...
+        "file_path": result[0].file_path,
+        "citeable_reference": result[0].citeable_reference,
+        "series_id": result[0].series_id,
+        "series_name": result[0].series_name,
+        "transferring_body": result[0].transferring_body,
+        "transferring_body_id": result[0].transferring_body_id,
+        "transferring_body_description": result[
+            0
+        ].transferring_body_description,
+        "consignment_reference": result[0].consignment_reference,
```
ltrasca: would useful to have consignment_id here

data_management/opensearch/tests/test_process_and_index_file.py
```diff
@@ -0,0 +1,197 @@
...
+    __tablename__ = "Body"
+    BodyId = Column(UUID(as_uuid=True), primary_key=True)
+    Name = Column(Text)
+    Description = Column(Text)
+
+
+class Series(Base):
```
ltrasca: might be useful to have these in a dedicated file (perhaps models.py) as opposed to the tests, also I agree to the previous comment you left about these

data_management/opensearch/tests/test_lambda_handler.py
```diff
@@ -0,0 +1,128 @@
+import json
```
ltrasca: this is a bit of a question but would we want to extend the unit_test workflow to include running tests for additional modules like this?

ltrasca (commented):
just a few comments from me, very good work 🤝

data_management/opensearch/tests/test_process_and_index_file.py
anthonyhashemi: yeah for sure. I going to make a ticket to look into decoupling the model in models.py fro flask so we can use here too

data_management/opensearch/tests/test_lambda_handler.py
anthonyhashemi: great point, yes! I will adjust it now

data_management/opensearch/index_file_content_and_metadata_in_opensearch.py
anthonyhashemi: We're not using in the search at the moment for whatever reason so I hadnt, but yeah definitely makes sense to, will add

data_management/opensearch/index_file_content_and_metadata_in_opensearch.py
```diff
@@ -0,0 +1,131 @@
...
+    file_id: str,
+    document: Dict[str, Any],
+    open_search_host_url: str,
+    open_search_http_auth: AWS4Auth,
+) -> None:
+    open_search = OpenSearch(
+        open_search_host_url,
+        http_auth=open_search_http_auth,
+        use_ssl=True,
+        verify_certs=False,
```
anthonyhashemi: ideally we would verify certs. Might make a followup ticket to deal with this, as we also need to handle for the webapp when we search.

data_management/opensearch/tests/test_process_and_index_file.py
anthonyhashemi: https://national-archives.atlassian.net/browse/AYR-1245

data_management/opensearch/tests/test_process_and_index_file.py
anthonyhashemi: https://national-archives.atlassian.net/browse/AYR-1245

ltrasca (dismissed):

data_management/opensearch/index_file_content_and_metadata_in_opensearch.py
anthonyhashemi: just switched to True now.

ltrasca (approved):

PR #559, opened 2024-09-23
AYR-1232 - Adjust permissions for search TB & summary pages

ltrasca:
## Changes in this PR

- Added check inside the transferring body search endpoint to only allow standard users with permissions (else, we return 404)
- Added check inside the summary_search_page to return 403 to non all-access users
- Adjusted tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1232

## Screenshots of UI changes
N/A


anthonyhashemi (approved):
Nice one 👌

PR #563, opened 2024-09-25
AYR-1238 - Fixes for OpenSearch sorting query

ltrasca:
## Changes in this PR

- Modified existing function that built the OpenSearch sort query to map our search terms to value names (e.g. closure_type is mapped to metadata.closure_type.keyword) - this has a secondary function of allowing us to check that a sort term should be added to the query or not, random terms should not be added to it
- Fixed dropdown component not being passed a correct object with sort terms & subsequently not showing the currently selected term
- Added unit tests to handle some edge cases and check for dropdown values when a sort term is selected

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1238

## Screenshots of UI changes

N/A

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -588,7 +588,7 @@ def search_transferring_body(_id: uuid.UUID):
 
     filters = {"query": query}
 
-    sort_query = build_sorting_orders_open_search(request.args)
+    sorting_query = build_sorting_orders_open_search(request.args)
     sorting_orders = build_sorting_orders(request.args)
 
     breadcrumb_values = {
@@ -649,7 +649,7 @@ def search_transferring_body(_id: uuid.UUID):
                     "filter": [{"term": {"transferring_body_id.keyword": _id}}],
                 }
             },
-            "sort": sort_query,
+            "sort": sorting_query,
         }
         size = per_page
         page_number = page
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -587,7 +587,9 @@ def search_transferring_body(_id: uuid.UUID):
     page = int(request.args.get("page", 1))
 
     filters = {"query": query}
-    sorting_orders = build_sorting_orders_open_search(request.args)
+
+    sort_query = build_sorting_orders_open_search(request.args)
+    sorting_orders = build_sorting_orders(request.args)
 
     breadcrumb_values = {
         0: {"query": ""},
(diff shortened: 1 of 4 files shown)


app/main/routes.py
```diff
@@ -587,7 +587,9 @@ def search_transferring_body(_id: uuid.UUID):
     page = int(request.args.get("page", 1))
 
     filters = {"query": query}
-    sorting_orders = build_sorting_orders_open_search(request.args)
+
+    sorting_query = build_sorting_orders_open_search(request.args)
+    sorting_orders = build_sorting_orders(request.args)
```
anthonyhashemi, line 592: `sorting_orders` isnt being used in this function anymore so I think this line can be removed

app/tests/test_search.py
```diff
@@ -1406,6 +1483,15 @@ def test_search_transferring_body_with_search_filter_full_test(
...
         assert response.status_code == 200
         soup = BeautifulSoup(response.data, "html.parser")
+        select = soup.find("select", {"id": "sort"})
+        option = select.find("option", selected=True)
+        option_text = "series_id-asc"
+
+        if option:
+            option_text = option.get("value")
+
+        assert select
```
anthonyhashemi, line 1537: what are we asserting here?

app/tests/test_search.py
```diff
@@ -1406,6 +1483,15 @@ def test_search_transferring_body_with_search_filter_full_test(
 
         assert response.status_code == 200
         soup = BeautifulSoup(response.data, "html.parser")
+        select = soup.find("select", {"id": "sort"})
+        option = select.find("option", selected=True)
+        option_text = "series_id-asc"
+
+        if option:
```
anthonyhashemi: Can we break this out into 2 tests, one where option is expected and another where it is not, so that each test is testing a distinct scenario so it is easier to debug and reason with

app/tests/test_search.py
ltrasca: just that the select exists

app/tests/test_search.py
anthonyhashemi: I saw the updated commit but looks like we still have that condition

ltrasca: @anthonyhashemi pushed a commit with the changes

anthonyhashemi (approved):

PR #572, opened 2024-10-01
AYR-1235 - Replace pagination with new component

ltrasca:
## Changes in this PR
- Deleted older pagination component and replaced it with the new one under the same file name
- Added utility to endpoints to calculate pagination and pass it to its respective template
- Removed an extra mention of pagination on *app/templates/main/browse-transferring-body.html* (we never had more than one transferring_body on this page so guessing it went under the radar)

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1235

## Screenshots of UI changes

N/A


app/templates/main/browse-transferring-body.html
```diff
@@ -118,14 +118,5 @@ <h3 class="govuk-heading-s govuk-heading-s--browse-all-filter">Transfer date</h3
                 </div>
             </div>
         </div>
-        <!-- PAGINATION -->
-        {% if num_records_found > 0 %}
-            <div class="govuk-grid-column-two-thirds browse-grid--two-thirds mobile-pagination">
-                {% with view_name='main.browse_transferring_body', id = request.view_args['_id'] %}
```
anthonyhashemi, line 124: why have we removed this?

ltrasca: the page already has pagination under the table at line 76 (alike the other pages) this was an extra mention of it that wasn't noticed

image

anthonyhashemi: ah and we never saw the paginartion widget for this view as we never had that any bodies in our dev/test data, ok

anthonyhashemi (approved):
Nice one

PR #574, opened 2024-10-02
AYR-1277 - Query term bug

ltrasca:
## Changes in this PR

- Fixed a bug where if the query param in the URL was edited to be empty the breadcrumbs would break

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1277

### Before
image

### After
image

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -508,7 +508,7 @@ def search_results_summary():
             hosts=current_app.config.get("OPEN_SEARCH_HOST"),
             http_auth=current_app.config.get("OPEN_SEARCH_HTTP_AUTH"),
             use_ssl=True,
-            verify_certs=False,
+            verify_certs=True,
             connection_class=RequestsHttpConnection,
         )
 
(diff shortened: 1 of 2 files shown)


app/main/routes.py
```diff
@@ -508,7 +508,7 @@ def search_results_summary():
             hosts=current_app.config.get("OPEN_SEARCH_HOST"),
             http_auth=current_app.config.get("OPEN_SEARCH_HTTP_AUTH"),
             use_ssl=True,
-            verify_certs=True,
+            verify_certs=False,
```
anthonyhashemi: I know this is needed currently for local dev but needs reverting here

app/main/routes.py
```diff
@@ -642,7 +643,7 @@ def search_transferring_body(_id: uuid.UUID):
...
```
anthonyhashemi: and here

app/main/routes.py
```diff
@@ -607,6 +607,7 @@ def search_transferring_body(_id: uuid.UUID):
         0: {"query": ""},
         1: {"transferring_body_id": _id},
         2: {"transferring_body": db.session.get(Body, _id).Name},
+        3: {"search_terms": "‘’"},
```
anthonyhashemi, line 610: could put in a test for this scenario

ltrasca: added a test for this

anthonyhashemi (approved):

PR #577, opened 2024-10-03
ensure opensearch dates retrieved are in required format

colinbowen:
## Changes in this PR
- Format dates in results retrieved from OpenSearch to be in required AYR DD/MM/YYY format

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1233

### After


anthonyhashemi (approved):
Nice one 👌

PR #582, opened 2024-10-07
Dev local services docker compose

anthonyhashemi:
### After


README.md
```diff
@@ -148,6 +148,133 @@ You should now have the app running on <https://localhost:5000/>
...
+export OPEN_SEARCH_CA_CERTS=local_services/opensearch_certs/root-ca.pem
+
+export DB_SSL_ROOT_CERTIFICATE=local_services/webapp_postgres_certs/root-ca.pem
+```
+
+Finally you can populate the opensearch cluster with the corresponding data stored in snapshot 1 in `local_services/snapshots/` by running `source .env && local_services/opensearch-entrypoint.sh`.
+
+Then you can run the flask server with `flask run`
+
+## Local dev without docker
```
anthonyhashemi, line 202: ignore this section, I just moved this from below and put it under "Local dev without docker" in case anyone would want to continue doing that

colinbowen (approved):

PR #587, opened 2024-10-08
Fix textract process

anthonyhashemi:
## Changes in this PR

Fix textract process which was failing on pdfs or docxs before when not specifying a file type on the tmp file we pass to textract,

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1199


data_management/opensearch_indexer/tests/test_process_and_index_file.py
```diff
@@ -196,3 +198,40 @@ def test_index_file_content_and_metadata_in_opensearch(
...
+            == "This is line 1\nThis is line 2\nThis is line 3\nThis is line 4, the final line.\n"
+        )
+
+    def test_docx_file(self):
+        path = Path(__file__).parent / "multiline.docx"
+        with open(path, "rb") as file:
+            file_stream = file.read()
+        file_type = "docx"
+
+        assert extract_text(file_stream, file_type) == (
```
anthonyhashemi, line 220: for both pdf and docx, textract seems to pull out whitespace characters like these. Probably not much we can do apart from potentially do our own trimming. Maybe lets see if this causes issues with opensearch. WIll leave to start with in case we remove characters that should have been there

ltrasca (approved):

PR #594, opened 2024-10-11
AYR-1285 -  Change default font to Arial

ltrasca:
## Changes in this PR
- Added a CSS rule that applies the Arial font family to all elements in the page - it uses the !important directive which is not ideal, the reasoning behind this is that its impossible for us achieve the same thing by overriding SCSS variables (the proper way to do it) - we use a minified css file to import govuk styles and not the npm package and thus the styles are already built.

There are quite a lot of sideeffects that come along with changing the font, will catch up with @Terry-Price to review.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1285

### DIFF
--- a/app/static/src/scss/includes/_overrides.scss
+++ b/app/static/src/scss/includes/_overrides.scss
@@ -1,3 +1,8 @@
 legend {
   padding-inline: 0;
 }
+
+// use of !important is not ideal but fits this purpose
+* {
+  font-family: Arial, Helvetica, sans-serif !important;
+}


app/static/src/scss/includes/_overrides.scss
```diff
@@ -1,3 +1,8 @@
 legend {
   padding-inline: 0;
 }
+
+// use of !important is not ideal but fits this purpose
```
anthonyhashemi, line 5: Having read your description and then information on !important, I see why you say its not ideal; it could override other fonts unintentionally if there was a specific different font needed somewhere else on the site.

Given the reason for needing to use this, do you think it it worth us creating a tech debt ticket for us to switch to using the npm package rather than the prebuilt minified css? I imagine there could be other benefits?

anthonyhashemi (commented):
What are the side effects you mention? Will hold off approving until you have had your discussion with Terry

app/static/src/scss/includes/_overrides.scss
ltrasca: @anthonyhashemi yes definitely, I don't think we use any other fonts at the moment but its still a side effect to be aware of and that we don't want. Having a ticket to have the npm package instead would be good.

PR #603, opened 2024-10-15
AYR-1203/1288 - Update TB search page to current design 

ltrasca:
## Changes in this PR
- Changes to table on the search TB page in order to bring it up to speed with Terry's designs
- Added accordions to the rows of the table (works with or without JS)
- Text highlighting returned by OpenSearch
- The expand/collapse feature via checkbox
- Addition of an inner table that contains metadata about the record

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1203
https://national-archives.atlassian.net/browse/AYR-1288

### Before

image

### After

image
image
image


app/tests/test_search.py
```diff
@@ -108,9 +117,10 @@ def verify_search_desktop_transferring_body_header_row(data):
 
     expected_row = (
         [
-            "Series reference",
-            "Consignment reference",
-            "File name",
+            "Found within",
```
anthonyhashemi: I was going to ask about this function and mention it seems odd to have found within and search results alongside these others but its no longer being used, so can be removed entirely!

However, it does highlight that we have no tests or assertions in existing tests covering the search Found within and search results part of the new results screen

app/tests/utils.py
```diff
@@ -20,23 +29,22 @@ def decompose_mobile_invisible_elements(soup):
...
-def evaluate_table_body_rows(soup, expected_results):
+def get_table_rows_cell_values(table):
     """
-    Gets all rows and cell values for a table inside a page after decomposing invisible items then asserts the result
-    :param soup: BeautifulSoup
-    :param espected_results: string[][]
+    Returns all rows and cell values for a table that has been found using BeautifulSoup
+    :param table: BeautifulSoup object
+    Returns:
+    string[][], e.g. [["row_1_value_1", "row_1_value_2"], ["row_2_value_1", "row_2_value_2"]]
```
anthonyhashemi: not familiar with this return type syntax, is it for python? I am familiar with the mypy List[List[str]]

app/tests/test_search.py
```diff
@@ -1655,16 +1730,108 @@ def test_search_transferring_body_with_search_term_edge_case_path(
...
+                    ],
+                ],
+            ),
+            (
+                "&query=!!!!!",
+                os_mock_return_tb,
+                [
+                    [
+                        "Test field 1 +1",
+                        "test1 and ... this is just a sentence with a mark element in it",
```
anthonyhashemi: where is the mark element?

app/tests/test_search.py
anthonyhashemi: this still seems funky, cant we differentiate the theads we are looking at.

The new 2 are for an outer t head,

the rest are for an inner thead

and should not be one expected row, from my understanding?

app/tests/utils.py
ltrasca: im used to typescript which is why I wrote it like this, but yes its List[List[str]]

app/tests/test_search.py
ltrasca: @anthonyhashemi the way BeautifulSoup strips the text & the tags inside certain elements makes keeping the tags difficult, which is why I've opted to test raw text like this (which is the output we get from BeautifulSoup) and then add the tests to check if the text inside the mark elements is correct further on in the file

app/tests/test_search.py
ltrasca: @anthonyhashemi will add two separate tests to differentiate between the two

app/tests/test_search.py
anthonyhashemi: Oh I see, what elements does it strip and is it a certain method we call on the original soup object that does that? It would be nice to have some test that shows that we persist the marks around the matching text, this test wouldnt catch a bug if we accidentally stripped the tags some how.

app/tests/utils.py
anthonyhashemi: nitpic: could we omit or change to the python syntax

app/tests/test_search.py
```diff
@@ -1755,3 +1922,74 @@ def test_search_transferring_body_returns_correct_date_format(
...
+
+        form_data = {"query": "test"}
+        response = client.get(
+            f"{self.route_url}/{transferring_body_id}", data=form_data
+        )
+
+        assert response.status_code == 200
+        soup = BeautifulSoup(response.data, "html.parser")
+
+        mark_elements = soup.find_all("mark")
```
anthonyhashemi: ah I see, you have this test here. Could we not just have this in one test as essentially we just care that given a search with some query, we get results with matches where we have highlights corresponding to the query

ltrasca: @anthonyhashemi could merge the two together, the test with raw text also tests that we concat strings together with ellipses " ... "

app/tests/test_search.py
ltrasca: @anthonyhashemi we gather td elements from the table and for each we call get_text(strip=True), which essentially strips all tags, new lines, etc. and returns the raw text of all elements inside - this is done because some fields may not just contain text, some (like the first cell in the table that also contains the `details` element for triggering the accordion) have other tags inside, these could also be span, divs, etc., perhaps there is a way to retain these tags but at that point we'd also be testing against what is essentially just HTML, which is not ideal because it may change

will try to figure out a way to keep just the `mark` tags

app/tests/utils.py
ltrasca: @anthonyhashemi yes should be changed now

anthonyhashemi (approved):
Nice one, a lot of good work put into this 🦾

PR #618, opened 2024-10-23
Use presigned url in manifest generation

colinbowen:
## Changes in this PR
- Uses presigned url to generate file paths in iiif manifest
- Requires CORS headers to be sent from S3

## JIRA ticket

https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-1212

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/main/util/render_utils.py
+++ b/app/main/util/render_utils.py
@@ -84,6 +84,16 @@ def generate_pdf_manifest(record_id):
         "main.download_record", record_id=record_id, _external=True, render=True
     )
 
+    presigned_url = None
+    try:
+        presigned_url = create_presigned_url(file)
+    except Exception as e:
+        current_app.logger.info(
+            f"Failed to create presigned url for document render non-javascript fallback {e}"
+        )
+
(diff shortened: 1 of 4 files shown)


app/main/util/render_utils.py
```diff
@@ -84,6 +84,16 @@ def generate_pdf_manifest(record_id):
         "main.download_record", record_id=record_id, _external=True, render=True
     )
 
+    presigned_url = None
+    try:
+        presigned_url = create_presigned_url(file)
```
anthonyhashemi, line 89: I searched on main and noticed there's no test for `create_presigned_url`, could we add one for it too

anthonyhashemi:
Looks good, just would be good to add test for create_presigned_url itself

app/main/util/render_utils.py
colinbowen: Sure will add

anthonyhashemi (approved):

PR #627, opened 2024-10-29
AYR-1282 - Modifying top search component to support search areas

ltrasca:
## Changes in this PR
- Refactored some of the search summary and search TB routes code
- Added search utilities that are reused by both endpoints
- Modified the search query to adjust fields depending on search area and use "AUTO" fuzziness
- The indexer now provides OpenSearch with a flat structure for record fields
- Modified the top search component to contain radios and be consistent with the design
- Added unit tests for the newly added search utilities and the radio buttons
- Modified "search" endpoint to pass along the search area along with the query
- Fixed tabs not being switchable on record page

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1281

### Before
image

### After
image


app/main/routes.py
```diff
@@ -457,9 +462,13 @@ def browse_consignment(_id: uuid.UUID):
 @bp.route("/search", methods=["GET"])
 @access_token_sign_in_required
 def search():
-    query = request.form.get("query", "") or request.args.get("query", "")
+    form_data = request.form.to_dict()
+    args_data = request.args.to_dict()
+
+    # merge both dictionaries (args takes precedence over form if there are overlapping keys)
```
anthonyhashemi, line 470: could we add a test for this edge case?

data_management/opensearch_indexer/opensearch_indexer/index_file_content_and_metadata_in_opensearch.py
```diff
@@ -87,6 +87,11 @@ def _fetch_file_data(
     session.close()
 
     file_data = {
+        **{
```
anthonyhashemi, line 90: I think it is worth us documenting this in a new ADR as it is a fundamental data management change, which is different to the initial decision in ADR 17

app/tests/test_search.py
```diff
@@ -581,22 +568,65 @@ def test_search_transferring_body_top_search(
...
         assert button is not None
-        assert text is not None
-        assert (
-            text_content
-            == "Search by file name, series or consignment reference."
+        assert radio_1 and radio_2 and radio_3
+        # radio for search area "everywhere" should be checked by default
+        assert "checked" in radio_1.attrs
+
+    @pytest.mark.parametrize(
```
anthonyhashemi, line 583: nice

data_management/opensearch_indexer/opensearch_indexer/index_file_content_and_metadata_in_opensearch.py
ltrasca: @anthonyhashemi added a "Modification" section at the bottom of the document explaining the change

anthonyhashemi: cheers, I will take a look

app/main/routes.py
```diff
@@ -590,132 +543,66 @@ def search_results_summary():
 
 @bp.route("/search/transferring_body/<uuid:_id>", methods=["GET"])
 @access_token_sign_in_required
-# C901 (function too complex) being ignored until AYR-1307
```
anthonyhashemi, line 593: we can probably mark AYR-1307 as done / wont fix when this is in

app/main/routes.py
```diff
@@ -590,132 +543,66 @@ def search_results_summary():
...
-            )
-        except opensearchpy.exceptions.ConnectionTimeout:
-            abort(504)
+        open_search = setup_opensearch()
+        dsl_query = build_dsl_query(query, search_area, open_search, _id)
 
+        search_results = execute_search(open_search, dsl_query, page, per_page)
         results = search_results["hits"]["hits"]
 
         for result in results:
```
anthonyhashemi: can we extract this for loop into a format_opensearch_results function as we're doing a refactor?

app/main/routes.py
```diff
@@ -727,10 +614,11 @@ def search_transferring_body(_id: uuid.UUID):  # noqa: C901
         num_records_found=num_records_found,
         sorting_orders=sorting_orders,
         search_terms=search_terms,
+        search_area=search_area,
         pagination=pagination,
         open_all=open_all,
         query_string_parameters={
-            k: v for k, v in request.args.items() if k not in "page"
+            k: v for k, v in request.args.items() if k != "page"
```
anthonyhashemi, line 631: Ah yeah, good spot, we dont want to filter out any query param apart from page as that's already passed in, substrings should still be allowed (even if unlikely)!

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+
+def get_param(param):
+    return request.form.get(param, "") or request.args.get(param, "")
+
+
+def get_query_and_search_area():
+    """Fetch query and search_area from form or request args"""
+    query = get_param("query")
+    search_area = get_param("search_area")
+    return query.strip(), search_area
```
anthonyhashemi, line 37: Is this abstraction adding anything if we just build out 2 calls for two values which we then have to return and split out in the route anyway?

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        use_ssl=True,
+        verify_certs=True,
+        ca_certs=current_app.config.get("OPEN_SEARCH_CA_CERTS"),
+        connection_class=RequestsHttpConnection,
+    )
+
+
+def execute_search(open_search, dsl_query, page, per_page):
+    """Execute the search query using OpenSearch"""
+    size = per_page
```
anthonyhashemi: I guess we can just define size as the argument, or just pass per_page

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+def get_all_fields_excluding(open_search, index_name, exclude_fields=None):
+    """Retrieve all fields from the index and exclude certain fields"""
+    mappings = open_search.indices.get_mapping(index=index_name)
+    all_fields = list(mappings[index_name]["mappings"]["properties"].keys())
+    filtered_fields = get_filtered_list(all_fields, exclude_fields)
+
+    return filtered_fields
+
+
+def get_query_aggregations():
```
anthonyhashemi: nitpick but I'd suggest renaming all these query definition functions from `get_*` to `build_*` as we aren't retrieving any data

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+        "sort": sorting_query,
+    }
+
+
+def get_query_source_rules():
+    """Returns the _source segment of the DSL query"""
+    return {"_source": {"exclude": ["*.keyword"]}}
+
+
+def build_dsl_query(query, search_area, open_search, _id=None):
```
anthonyhashemi: can we rename _id to transferring_body_id as there is noway to know it's meant to be that, and some context explaining that this function can be used to search across all bodies, or just the one if body is provided

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+                            "_source": ["transferring_body"],
+                        }
+                    }
+                },
+            }
+        },
+    }
+
+
+def get_query_multi_match(query, search_area, open_search, _id):
```
anthonyhashemi: same comment about transferring body id here

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+    all_fields = list(mappings[index_name]["mappings"]["properties"].keys())
+    filtered_fields = get_filtered_list(all_fields, exclude_fields)
+
+    return filtered_fields
+
+
+def get_query_aggregations():
+    """Returns the aggregations segment of the DSL query"""
+    return {
+        "aggs": {
```
anthonyhashemi, line 121: so this is getting used by both search results summary and search transferring body. Can we skip this for the search transferring body?

app/tests/test_routes.py
```diff
@@ -244,3 +246,94 @@ def test_route_generate_image_manifest(
...
+        client: FlaskClient,
+        user_role,
+        form_data,
+        args_data,
+        expected_redirect_route,
+        expected_params,
+        mock_standard_user,
+        mock_all_access_user,
+    ):
+        if user_role == "standard_user":
```
anthonyhashemi: can we split this test function into tests that test one scenario each. Having conditionals in the tests makes them harder to follow as documentation pieces and harder to debug due to all the diverging paths

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,167 @@
...
+    query, search_area = get_query_and_search_area()
+    assert query == "test_query"
+    assert search_area == "test_area"
+
+
+@patch("app.main.util.search_utils.OpenSearch", MockOpenSearch)
+def test_setup_opensearch(app):
+    with app.app_context():
+        client = setup_opensearch()
+        assert isinstance(client, MockOpenSearch)
```
anthonyhashemi: I think we can improve this by not mocking OpenSearch and changing to

`assert isinstance(client, OpenSearch)`
and can even assert the properties too

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@patch("app.main.util.search_utils.OpenSearch")
+def test_execute_search(mock_open_search, app):
+    with app.app_context():
+        dsl_query = {"query": {"match_all": {}}}
+        execute_search(mock_open_search, dsl_query, page=1, per_page=10)
+        mock_open_search.search.assert_called_once()
```
anthonyhashemi: can improve by asserting what search is called with with
`mock_open_search.search.assert_called_once_with(a=foo,b=bar,etc.....)`

app/main/util/search_utils.py
```diff
@@ -0,0 +1,174 @@
...
+                },
+            }
+        },
+    }
+
+
+def get_query_multi_match(query, search_area, open_search, _id):
+    """Returns the multi-match segment of the DSL query"""
+
+    fields_record = ["file_name", "file_path", "content"]
```
anthonyhashemi: thinking we can move this block from 93 to 102, defining the `fields` variable outside this function and then pass it in as an argument, that way we don't need to pass open_search instance to the build_dsl_query at all, then it will just be an easily tested function without any mocking.

Then we just need to do extract this out, maybe to its own function, something like `get_open_search_fields_to_search_on` and call it before we call this function in the routes.

Then the test for the new function is the only test that needs to mock opensearch (apart from the route test)

Then you could just do:

`build_dsl_search_query(query, search_fields)` for search summary results

and 

`build_dsl_search_query(query, search_fields, transferring_body_id)` for search transferring_body

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_get_query_source_rules():
+    source_rules = get_query_source_rules()
+    assert source_rules == {"_source": {"exclude": ["*.keyword"]}}
+
+
+@patch(
+    "app.main.util.search_utils.get_query_source_rules",
+    return_value={"_source": {"exclude": ["*.keyword"]}},
```
anthonyhashemi: seems unnecessary to mock a lot of these functions which are just hardcoded values., e.g. `get_query_source_rules` `get_query_highlighting`,`get_query_aggregations` etc.

Tbh you could even do away with the functions as they only get used here and are pretty simple/ could potentially say its easier to digest if the query is in one place (if each part was doing lots of calculations I might have thought differently though):

```
def build_dsl_search_query(query, search_fields, sorting_orders, transferring_body_id=None):
    filter_clauses = []
    aggregation_query =   {"aggs":
         "aggregate_by_transferring_body": {
                "terms": {"field": "transferring_body_id.keyword"},
                "aggs": {
                    "top_transferring_body_hits": {
                        "top_hits": {
                            "size": 1,
                            "_source": ["transferring_body"],
                        }
                    }
                },
            }
        },
    }
    if transferring_body_id:
        filter_clauses = (
            [{"term": {"transferring_body_id.keyword": _id}}]
        )
        aggregation_query = None

    dsl_query = {
        "query": {
            "bool": {
                "must": [
                     {
                        "multi_match": {
                             "query": query,
                            "fields": fields,
                             "operator": "AND",
                             "fuzziness": "AUTO",
                             "lenient": True,
                        }
                     }
                 ],
                "filter": filter_clauses,
            }
        },
        "sort": sorting_query,
        "_source": {"exclude": ["*.keyword"]},
        "highlight": {
            "pre_tags": ["<mark>"],
            "post_tags": ["</mark>"],
            "fields": {
                "*": {},
            },
        },
    }
    if aggregations:
        dsl_query.extend(aggregations)

    return dsl_query
```

in routes:
```
search_fields = get_open_search_fields_to_search_on()
sorting_orders = build_sorting_orders_open_search(request.args)
search_query = build_dsl_search_query(query, search_fields, sorting_orders, transferring_body_id) # (or just search_query = build_dsl_search_query(query, search_fields, sorting_orders) for the summary results one)
````

then all three of these functions can be unit tested , and now `build_dsl_search_query` is really simple where it only has the logic at the top based off passing in transferring_body_id or not.

you could even take it a step further and have 

```
def build_dsl_search_query(query, search_fields, sorting_orders, filter_clauses)
    return {
        "query": {
            "bool": {
                "must": [
                     {
                        "multi_match": {
                             "query": query,
                            "fields": fields,
                             "operator": "AND",
                             "fuzziness": "AUTO",
                             "lenient": True,
                        }
                     }
                 ],
                "filter": filter_clauses,
            }
        },
        "sort": sorting_query,
        "_source": {"exclude": ["*.keyword"]},
        "highlight": {
            "pre_tags": ["<mark>"],
            "post_tags": ["</mark>"],
            "fields": {
                "*": {},
            },
        },
    }
```

```
def build_search_results_summary_query(query, search_fields, sorting_orders):
    filter_clauses = []
    dsl_query = build_dsl_search_query(query, search_fields, sorting_orders, filter_clauses)
    aggregation_query =   {"aggs":
         "aggregate_by_transferring_body": {
                "terms": {"field": "transferring_body_id.keyword"},
                "aggs": {
                    "top_transferring_body_hits": {
                        "top_hits": {
                            "size": 1,
                            "_source": ["transferring_body"],
                        }
                    }
                },
            }
        },
    }

    dsl_query.extend(aggregations)

    return dsl_query
````


```
def build_search_transferring_body_query(query, search_fields, sorting_orders, transferring_body_id):
    filter_clauses = [{"term": {"transferring_body_id.keyword": _id}}]
    dsl_query = build_dsl_search_query(query, search_fields, sorting_orders, filter_clauses)
    return dsl_query
````
which I think is pretty clean. What do you think?

File: app/main/util/search_utils.py (hunk 7 above)
Comment (ltrasca) [reply app/main/util/search_utils.py:37]: both endpoint use get_query_and_search_area so it slightly declutters things and make them easier to read

File: app/main/util/search_utils.py (hunk 8 above)
Comment (ltrasca) [reply app/main/util/search_utils.py]: 100%, I've just passed per_page so naming stays consistent

File: app/main/util/search_utils.py (hunk 12 above)
Comment (ltrasca) [reply app/main/util/search_utils.py:121]: Made it so if the ID is none this segment of the query doesn't get built (in the summary endpoint the transferring body ID doesnt exist)

**Review by anthonyhashemi — APPROVED** [review]

**Review by anthonyhashemi — CHANGES_REQUESTED** [review]
test hhas conditionals

File: app/tests/test_search_utils.py [hunk 18]
```diff
@@ -0,0 +1,311 @@
...
+def test_setup_opensearch(app):
+    app.config["OPEN_SEARCH_HOST"] = "localhost"
+    with app.app_context():
+        client = setup_opensearch()
+        assert isinstance(client, OpenSearch)
+
+
+@patch("app.main.util.search_utils.OpenSearch")
+def test_execute_search(mock_open_search, app):
+    with app.app_context():
```
anthonyhashemi: why do we need the app context?

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,311 @@
...
+    query, search_area = get_query_and_search_area()
+    assert query == "test_query"
+    assert search_area == "test_area"
+
+
+def test_setup_opensearch(app):
+    app.config["OPEN_SEARCH_HOST"] = "localhost"
+    with app.app_context():
+        client = setup_opensearch()
+        assert isinstance(client, OpenSearch)
```
anthonyhashemi: can we assert the properties of the opensearch based on the app.config

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,311 @@
...
+
+        mock_form_get.return_value = ""
+        assert get_param("param") == "test_arg_value"
+
+
+@patch(
+    "app.main.util.search_utils.get_param",
+    side_effect=["test_query", "test_area"],
+)
+def test_get_query_and_search_area(mock_get_param):
```
anthonyhashemi: we should not need to mock here, just make request a argument to `def get_param` and `get_query_and_search_area`

app/tests/test_search_utils.py
```diff
@@ -0,0 +1,311 @@
...
+        "metadata_field_3",
+    ]
+    actual_fields = get_open_search_fields_to_search_on(
+        mock_opensearch, search_area
+    )
+    assert actual_fields == expected_fields
+
+
+def test_get_param(app):
+    with patch(
```
anthonyhashemi: If we make this functional like the rest of the functions, we can pass request in and then tetsing we can pass in an object with the same interface as request

app/tests/test_routes.py
```diff
@@ -244,3 +246,121 @@ def test_route_generate_image_manifest(
...
+
+        query_string = "&".join(
+            [f"{key}={value}" for key, value in args_data.items()]
+        )
+        url = url_for("main.search") + "?" + query_string
+
+        response = client.get(url, data=form_data)
+        assert response.status_code == 302
+
+        if expected_redirect_route == "main.search_transferring_body":
```
anthonyhashemi: we still have conditionals in this test, can we split them out into individual test scenarios

app/tests/test_routes.py
```diff
@@ -244,3 +246,121 @@ def test_route_generate_image_manifest(
...
```
anthonyhashemi: same here, lets remove conditionals by splitting out permutations that require conditional test logic into their own tests

app/main/util/search_utils.py
```diff
@@ -0,0 +1,160 @@
...
+def format_opensearch_results(results):
+    results_clone = results
+    for result in results_clone:
+        for key, value in result["_source"].items():
+            if "date" in key:
+                result["_source"][key] = format_opensearch_date(value or "")
+    return results_clone
+
+
+def get_open_search_fields_to_search_on(open_search, search_area):
```
anthonyhashemi, line 18: could simplify

```
def get_open_search_fields_to_search_on(open_search, search_area):
    """Retrieve a list of fields depending on the search area (all fields, metadata, record, etc.)"""
    if search_area == "metadata":
        return get_all_fields_excluding(
        open_search, "documents", fields_record
    )
    elif search_area == "record":
        return ["file_name", "file_path", "content"]
    return ["*"]
```

app/tests/test_routes.py
ltrasca: @anthonyhashemi this is just a reminant from the older test, because we've split the tests into all access and standard user the conditional shouldnt have been there at all

app/tests/test_search_utils.py
ltrasca: We dont cannot remember why thats there, think I came across some technicality where pytest was complaining about working outside of app context but seems fine now so I've removed it

app/tests/test_routes.py
anthonyhashemi: ah okay, coool

anthonyhashemi (approved):

PR #630, opened 2024-10-30
Fix for the tabs

ltrasca:
## Changes in this PR

- Added fix for govuk-frontend js not being initialized properly

### After

### DIFF
--- a/app/templates/base.html
+++ b/app/templates/base.html
@@ -111,8 +111,6 @@
 {% endblock %}
 {% block bodyEnd %}
     <!--[if gt IE 8]><!-->
-    <script type="module"
-            src="{{ url_for('static', filename='govuk-frontend.min.js') }}"></script>
     <script type="module">
         import { initAll } from "{{ url_for('static', filename='govuk-frontend.min.js') }}"
         initAll()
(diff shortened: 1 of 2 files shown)


app/templates/base.html
```diff
@@ -111,7 +111,9 @@ <h2 class="govuk-visually-hidden">Support links</h2>
 {% endblock %}
 {% block bodyEnd %}
     <!--[if gt IE 8]><!-->
-    <script type="module" src="{{ url_for('static', filename='govuk-frontend.min.js') }}">
+    <script type="module"
+            src="{{ url_for('static', filename='govuk-frontend.min.js') }}"></script>
```
anthonyhashemi: why do we need this separate script block? Does the first one import everything from that js file and the second just initalAll and call it? How do we use the stuff from the first one and how does it fix the bug?

ltrasca: @anthonyhashemi they had something like this in the documentation which is why I've added that line but I've removed it and it seems to work just fine
image

anthonyhashemi: interesting, okay, if it works without sounds good

anthonyhashemi (approved):

PR #640, opened 2024-11-05
AYR-1203/1322 - Show file name & hide keyword fields

ltrasca:
## Changes in this PR
- Changed template to always show the file name in some fashion - if the highlighting has file_name as the first result, then the behaviour is as before, if not, the file name is always shown as the 2nd row in the table and is always visible
- If the file name field is not the first in the highlight results BUT it exists in the highlight results, it gets shown as the 2nd row, if not we show the file name from inside _source - this allows us to show highlighted results if appropriate
- Fields that contain ".keyword" will now be filtered out of the highlight results, a positive consequence of this is that fields such as "transferring_body_id.keyword" will now not be shown to the user
- Added escaping for highlighted content in the form of filtering for HTML tags (which was previously found to be possible), this is in part to reduce the risk of script tags being introduced into the content or other tags breaking page layout
- Added tests for all the above, including the Jinja template filters
- When there is just one highlight result there was a bug where the first row title contained "+0", in this instance this segment of the cell value is not shown

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1203
https://national-archives.atlassian.net/browse/AYR-1322

### Before
image

### After
image


app/main/util/search_utils.py
```diff
@@ -7,21 +7,41 @@
...
     return results_clone
 
 
+def filter_opensearch_highlight_results(results):
+    """Filter highlight results for fields that are not needed"""
+    results_clone = results.copy()
+    for result in results_clone:
+        if result.get("highlight", None):
+            keys_to_remove = [
+                key for key in result["highlight"].keys() if ".keyword" in key
```
anthonyhashemi, line 25: what keys might we be removing here? i.e what keys use .keyword? I understand we donyt want transferring body but would we not want some of them?

app/templates/main/search-transferring-body.html
```diff
@@ -22,6 +22,7 @@
     "Record status (closed)": "closure_type-asc",
     "Record status (open)": "closure_type-desc"
 } %}
+{% set anchor_fields = {"file_name": "file_id", "file_path": "file_id"} %}
```
anthonyhashemi: why is file_path in this anchor_fields dict?

app/main/util/search_utils.py
```diff
@@ -107,8 +127,9 @@ def build_dsl_search_query(
                 "filter": filter_clauses,
             }
         },
-        "sort": sorting_orders,
-        "_source": {"exclude": ["*.keyword"]},
+        # set as {} until sorting ticket is in done
```
anthonyhashemi, line 130: makes sense

app/templates/main/search-transferring-body.html
```diff
@@ -90,36 +91,63 @@ <h2 class="govuk-body browse__body">You are viewing</h2>
                                                 {% for record in results %}
                                                     {% set main_loop_index = loop.index %}
                                                     <tr class="govuk-table__row govuk-table__row--primary">
+                                                        {% set first_key = record['highlight'].keys() | list | first %}
```
anthonyhashemi: what's important about the first key and value?

Could we name them to something more specific, what actually are the key and value representing?

app/__init__.py
```diff
@@ -16,13 +17,19 @@
...
+    """Filter that converts string values that are "null" or Nones to dash"""
     if value == "null":
         return "-"
     if value is None:
         return "-"
     return value
 
 
+def clean_tags(text):
+    """Filter that removes ALL HTML tags that are not <mark>"""
```
anthonyhashemi: Annoying that opensearch doesnt sanitize results like elastcsearch does: https://www.elastic.co/guide/en/app-search/current/sanitization-guide.html


[Maybe?/might not be worth] It could be worth, making "mark" a variable so we can pass the sam variable to this function and to the function that builds up the highlighting?

anthonyhashemi: also, wha happens if our docs contain mark tags?

anthonyhashemi: I think we'd need to use the variable approach, specify the highlight marker to be a uuid then

`.replace(f{"<{highlight_marker_uuid}>", "<mark>").replace(f"</{highlight_marker_uuid}>", "</mark>")`
when using n the template.

app/templates/main/search-transferring-body.html
```diff
@@ -90,36 +91,63 @@ <h2 class="govuk-body browse__body">You are viewing</h2>
...
                                                             <div class="govuk-table__cell__data">
                                                                 <details class="details-toggle-{{ main_loop_index }}"
                                                                          {% if open_all %}open{% endif %}>
                                                                     <summary />
                                                                 </details>
                                                                 {% if record['highlight'] %}
-                                                                    {% set key = record['highlight'].keys() | list | first | replace('_', ' ') | capitalize %}
-                                                                    {{ key }} +{{ record['highlight'].keys() | length - 1 }}
+                                                                    {% set count = record['highlight'].keys() | length - 1 %}
+                                                                    {% set field = first_key | replace('_', ' ') | capitalize %}
```
anthonyhashemi: There is a lot of logic in these jinja templates now. I think we should start unit testing them directly rather than only when testing the routes as the routes themselves have a lot of logic.

I will create a ticket(or tickets) for us to start doing this across our current templates. We should also split out a lot of them as they are quite large.

something like


```
def test_template_a():
    context = { 
        'test_var': 'test_value'
    }
    path = 'path/to/template/dir'
    filename = 'template_to_test.html'
    rendered = jinja2.Environment(
        loader=jinja2.FileSystemLoader(path)
    ).get_template(filename).render(context)
    expected = #......
    assert rendered == expected
```
and we could maybe even use something like: https://github.com/MrSenko/coverage-jinja-plugin to show we are covering all edge cases in these templates

app/tests/test_jinja_filters.py
```diff
@@ -0,0 +1,73 @@
...
+        ),
+        # EXTREMELY IMPORTANT: script tags with attributes
+        (
+            "<script src='myscript.js'>This is a script tag with attrs</script>",
+            "This is a script tag with attrs",
+        ),
+    ],
+)
+def test_clean_tags(input_text, expected_output):
+    assert clean_tags(input_text) == expected_output
```
anthonyhashemi: I'd lean on the side of using bleach: https://github.com/mozilla/bleach (even though it is deprecated it is still being uspported indefinitely) as it is a tried and tested, robust sanitization lib:

but tbh I have thought through various scenarios and I can't think of any others where would have issues, apart from when there is mark tags already in the document and what I suggested above with UUIDs and replacing should handle that

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.2 and should not be changed by hand.
```
anthonyhashemi, line 1: I guess this file didnt need updating but doesnt look like anything changed apart from markup 3.01 to 3.02

app/tests/test_search.py
```diff
@@ -2009,3 +2023,147 @@ def test_search_transferring_body_all_accordions_open_with_open_all_query(
...
+        soup = BeautifulSoup(response.data, "html.parser")
+        anchors = soup.find_all("a")
+        anchors_text = []
+        for anchor in anchors:
+            anchors_text.append(anchor.get_text(strip=True))
+
+        table_body = soup.find("tbody")
+        table_rows_cell_values = get_table_rows_cell_values(table_body)
+        assert table_rows_cell_values[1] == ["File name", "fifth_file.doc"]
+        assert "fifth_file.doc" in anchors_text
```
anthonyhashemi: for this test and the one below could we put the anchors_text assertion right after we build them up so that it is easier to follow?

also could we make a variable called `expected_file_name = "fifth_file.doc"` and use it in both the anchors assertions and the table_rows_cell_values assertions so it is easy to show the relationship

app/tests/test_search.py
```diff
@@ -2009,3 +2023,147 @@ def test_search_transferring_body_all_accordions_open_with_open_all_query(
...
+                    },
+                    "hits": [
+                        {
+                            "_source": {
+                                "file_name": "fifth_file.doc",
+                            },
+                            "highlight": {
+                                "foo": ["bar"],
+                                "marco": ["polo"],
+                                "file_name": ["<mark>test_file.pdf</mark>"],
```
anthonyhashemi: Is this an actual scenario we'd expect? I.e is this a valid output from opensearch? When would this happen?

app/main/util/search_utils.py
ltrasca: @anthonyhashemi did some more reading up and seems like they might be important, though in our case we get duplicate results like this, I'm not entirely sure if we want this or not
image

app/templates/main/search-transferring-body.html
ltrasca: @anthonyhashemi removed it 🤝

app/templates/main/search-transferring-body.html
ltrasca: 100%, this would be a much better approach

app/tests/test_search.py
ltrasca: yes its pretty much identical to what we'd get from OS, it would happen when getting highlighting for the file_name field where its anything besides the 1st or 2nd result, here its the 3rd but we need to always show it as 2nd

app/tests/test_jinja_filters.py
ltrasca: will try this lib out, should be more comprehensive by default from what I can tell, e.g. we might still want to show tags and < or > symbols (or replacements of them) in the contents, at the moment with the filter I wrote they are entirely replaced

ltrasca: definitely much better with this lib 
image

app/main/util/search_utils.py
anthonyhashemi: let's hop on a call to investigate this together

app/tests/test_search.py
anthonyhashemi: sorry I should have been more specific. Why is the file_name different in _source and highlight?

app/templates/main/search-transferring-body.html
```diff
@@ -22,6 +22,7 @@
     "Record status (closed)": "closure_type-asc",
     "Record status (open)": "closure_type-desc"
 } %}
+{% set anchor_fields = {"file_name": "file_id"} %}
```
anthonyhashemi: I think we can get rid of this dict and simplify your logic below to

`if first_result_key == "file_name`


`<a href="{{ url_for('main.record', record_id=record['_source']["file_id"]) }}">{{ first_result_value }}</a>`

app/templates/main/search-transferring-body.html
```diff
@@ -90,36 +91,62 @@ <h2 class="govuk-body browse__body">You are viewing</h2>
...
                                                             <div class="govuk-table__cell__data">
                                                                 <details class="details-toggle-{{ main_loop_index }}"
                                                                          {% if open_all %}open{% endif %}>
                                                                     <summary />
                                                                 </details>
                                                                 {% if record['highlight'] %}
-                                                                    {% set key = record['highlight'].keys() | list | first | replace('_', ' ') | capitalize %}
-                                                                    {{ key }} +{{ record['highlight'].keys() | length - 1 }}
+                                                                    {% if field_count > 0 %}
+                                                                        {{ first_result_key }} +{{ field_count }}
```
anthonyhashemi: for consistency could we rename

`first_result_key` -> `first_field`


and change the logic so that filed_count doesnt include the minus 1, and then 
`{% if field_count > 1 %}
and then here you do 
`{{ first_field }} + {{ field_count - 1 }}`

could maybe rename field_count to `total_fields` or similar as field_count implies it is a variable that would be incremented

app/tests/test_search.py
```diff
@@ -101,6 +103,8 @@
     }
 }
 
+expected_file_name = "fifth_file.doc"
```
anthonyhashemi, line 107: can we just put this in each test that needs it? I feel like in tests unless it was unwieldy duplication its nicer to have all the context in there to understand the test

app/__init__.py
```diff
@@ -16,13 +17,20 @@
...
         return "-"
     if value is None:
         return "-"
     return value
 
 
+def clean_tags(text):
+    """Sanitizes ALL HTML tags that are not <mark>"""
+    allowed_tags = ["mark"]
+    return bleach.clean(text, tags=allowed_tags)
```
anthonyhashemi: this still is dangerous if someone has put malicious code in a "mark" tag,
e.g. `<mark onclick="alert('XSS')">Click here</mark>`

So I think the suggestion of using a uuid as original tag that we want opensearch to wrap our results in, then doing this clean with no allowed_tags and then str replacing it with mark afterwards, would be the only way to be 100% safe.

Let's chat through it though

app/__init__.py
```diff
@@ -25,10 +25,11 @@ def null_to_dash(value):
     return value
 
 
-def clean_tags(text):
+def clean_tags(text, highlight_tag):
```
anthonyhashemi: docstring definitely needs updating, now mark has been generalised to highlight_tag and also add info about how it then replaces with mark.

function_name should probably be updated too
maybe
`clean_tags_and_replace_highlight_tag_with_mark`

app/templates/main/search-transferring-body.html
```diff
@@ -91,20 +90,20 @@ <h2 class="govuk-body browse__body">You are viewing</h2>
...
                                                             <div class="govuk-table__cell__data">
                                                                 <details class="details-toggle-{{ main_loop_index }}"
                                                                          {% if open_all %}open{% endif %}>
                                                                     <summary />
                                                                 </details>
                                                                 {% if record['highlight'] %}
-                                                                    {% if field_count > 0 %}
-                                                                        {{ first_result_key }} +{{ field_count }}
+                                                                    {% if total_fields > 1 %}
+                                                                        {{ first_field }} +{{ total_fields }}
```
anthonyhashemi: this needs to be `total_fields - 1` now

the tests updated in this commit have incorrect expectation (off by 1)

app/main/routes.py
```diff
@@ -606,6 +606,7 @@ def search_transferring_body(_id: uuid.UUID):
         )
 
         search_results = execute_search(open_search, dsl_query, page, per_page)
+        print(json.dumps(search_results))
```
anthonyhashemi: need to remove this, as data could be sensitive so better not to log

app/__init__.py
```diff
@@ -16,13 +17,21 @@
...
+    """Filter that converts string values that are "null" or Nones to dash"""
     if value == "null":
         return "-"
     if value is None:
         return "-"
     return value
 
 
+def clean_tags_and_replace_highlight_tag(text, highlight_tag):
+    """Sanitizes ALL HTML tags that are not <mark>"""
```
anthonyhashemi: docstring still needs updating

app/main/util/search_utils.py
anthonyhashemi: called and decided its fine to remove all of the .keywords for now as always will have a text equivalent (field without the .keyword mapping) match but have created a ticket to investigate whether we can utuilise keyword for optimizations or remove it from index.

anthonyhashemi (dismissed):

app/tests/test_search.py
```diff
@@ -2009,3 +2029,152 @@ def test_search_transferring_body_all_accordions_open_with_open_all_query(
...
+        mock_search_client.return_value = MockOpenSearch(
+            search_return_value={
+                "hits": {
+                    "total": {
+                        "value": 1000,
+                    },
+                    "hits": [
+                        {
+                            "_source": {
+                                "file_name": "fifth_file.doc",
```
anthonyhashemi, line 2099: noticed this test still has the odd mocked data where the filename is different in source and highlight. Can we updated this and the assertions as discussed

anthonyhashemi (approved):

PR #642, opened 2024-11-07
Update search query to handle multiple query type searching within opensearch

colinbowen:
## Changes in this PR
- AND behaviour for search terms: Each term in the search query now acts as a filter, requiring that all terms be present in the results.
- Phrase Matching with Quotes:
Terms in quotes (e.g., "everybody book") are treated as exact phrases. Records must contain the exact phrase with words in the specified order
- Individual words not in quotes (e.g., best or during) are treated as single terms with fuzzy matching, which allows minor variations (e.g. "best" could match "test")

Examples:
- Quoted Phrase ("everybody book") - Matches documents containing the exact phrase "everybody book".
- Multiple Words without Quotes (best, during) - Matches documents containing both terms with flexibility for minor spelling errors.
- Mixed Query ("everybody book", best, during) - Combines exact phrase matching for "everybody book" with flexible term matching for best and during.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1278

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -588,22 +588,15 @@ def search_transferring_body(_id: uuid.UUID):
     if query:
         quoted_phrases, single_terms = extract_search_terms(query)
 
+        if search_filter:
+            if search_filter.startswith('"') and search_filter.endswith('"'):
+                quoted_phrases.append(search_filter[1:-1])
+            else:
+                single_terms.append(search_filter.strip())
         search_terms = quoted_phrases + single_terms
 
-        if query:
-            quoted_phrases, single_terms = extract_search_terms(query)
-
-            if search_filter:
-                if search_filter.startswith('"') and search_filter.endswith(
-                    '"'
-                ):
-                    quoted_phrases.append(search_filter[1:-1])
-                else:
-                    single_terms.append(search_filter.strip())
-            search_terms = quoted_phrases + single_terms
-
-            query = f"{query},{search_filter}" if search_filter else query
-            filters["query"] = query
+        query = f"{query},{search_filter}" if search_filter else query
+        filters["query"] = query
 
         breadcrumb_values[0] = {"query": query}
         display_terms = " + ".join(
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -1,5 +1,6 @@
 import io
 import json
+import re
 import uuid
 
 import boto3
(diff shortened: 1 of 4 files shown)


app/main/util/search_utils.py
```diff
@@ -90,27 +90,50 @@ def get_all_fields_excluding(open_search, index_name, exclude_fields=None):
 def build_dsl_search_query(
     query, search_fields, sorting_orders, filter_clauses
 ):
-    """Constructs the base DSL query for OpenSearch"""
-    return {
+    """Constructs the base DSL query for OpenSearch with AND behavior for multiple terms"""
+
+    search_terms = [term.strip() for term in query.split(",") if term]
```
ltrasca: would be good to have this segment dealing with building the must_clauses in its own utility - I think we also use the logic at line 95 quite a lot in other places & endpoints, could maybe also be split off from this and reused later on

app/tests/test_search_utils.py
```diff
@@ -335,6 +335,54 @@ def test_build_dsl_search_query():
     assert dsl_query == expected_base_dsl_search_query
 
 
+def test_build_dsl_search_query_and_exact_fuzzy_search():
+    query = '"exact match", fuzzy, search'
```
anthonyhashemi, line 344: I think we are missing a phrase in quotes that includes a comma to show that that doesn't split the phrase

app/main/util/search_utils.py
anthonyhashemi: Yeah, if we could build this out in a function that we call in the route then we can pass in must_clauses to this function (and the function that calls it above), like we do for the other args in this function.

app/main/util/search_utils.py
```diff
@@ -107,23 +109,54 @@ def get_all_fields_excluding(open_search, index_name, exclude_fields=None):
     return filtered_fields
 
 
+def build_must_clauses(query, search_fields):
```
anthonyhashemi: we should pass `single_terms` and `quoted_phrases` into here via `build_dsl_search_query` (or pass must_clauses` to `build_dsl_search_query` from the route itself) as we already build these for `search_terms = quoted_phrases + single_terms` in the route itself

app/main/routes.py
```diff
@@ -588,12 +588,21 @@ def search_transferring_body(_id: uuid.UUID):
...
         quoted_phrases, single_terms = extract_search_terms(query)
 
-        # Combine quoted phrases and single terms into final search_terms list
         search_terms = quoted_phrases + single_terms
 
-        if search_filter:
-            search_terms.append(search_filter)
-            query += f",{search_filter}" if search_terms else search_filter
+        if query:
+            quoted_phrases, single_terms = extract_search_terms(query)
```
anthonyhashemi: we are nesting the same if condtion and first line, we already did it above ^

colinbowen: Removed thanks

anthonyhashemi (approved):

app/main/util/search_utils.py
colinbowen: Updated

app/main/util/search_utils.py
colinbowen: Updated to build clauses in it's own func

app/tests/test_search_utils.py
colinbowen: Added tests and also created this ticket to handle special characters: https://national-archives.atlassian.net/browse/AYR-1372

PR #645, opened 2024-11-11
AYR-1327 - Format metadata fields

ltrasca:
## Changes in this PR
- Added Jinja filter utility to format the names of fields 
- Modified styles so the field count element is not visible when the accordion is open
- Added tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1327

### Before
image

### After
image

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -10,7 +10,7 @@ from jinja2 import ChoiceLoader, PackageLoader, PrefixLoader
 
 from app.logger_config import setup_logging
 from app.main.db.models import db
-from app.main.util.search_utils import opensearch_field_name_map
+from app.main.util.search_utils import OPENSEARCH_FIELD_NAME_MAP
 
 compress = Compress()
 talisman = Talisman()
@@ -35,9 +35,7 @@ def clean_tags_and_replace_highlight_tag(text, highlight_tag):
 
 def format_opensearch_field_name(field):
     """Format the name of an OpenSearch field using a map or dynamically"""
-    return opensearch_field_name_map.get(
-        field, field.replace("_", " ").capitalize()
-    )
+    return OPENSEARCH_FIELD_NAME_MAP.get(field)
 
 
 def create_app(config_class, database_uri=None):
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -10,23 +10,11 @@ from jinja2 import ChoiceLoader, PackageLoader, PrefixLoader
 
 from app.logger_config import setup_logging
 from app.main.db.models import db
+from app.main.util.search_utils import opensearch_field_name_map
 
 compress = Compress()
 talisman = Talisman()
 s3 = FlaskS3()
-
-
-opensearch_field_name_map = {
-    "file_name": "File name",
-    "description": "Description",
-    "Source-Organisation": "Transferring body",
-    "foi_exemption_code": "FOI code",
-    "content": "Content",
-    "closure_start_date": "Closure start date",
-    "end_date": "Record date",
-    "date_last_modified": "Record date",
-    "closure_start_date": "Closure start date",
-}
 
 
 def null_to_dash(value):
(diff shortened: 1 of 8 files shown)


app/__init__.py
```diff
@@ -16,6 +16,19 @@
 s3 = FlaskS3()
 
 
+opensearch_field_name_map = {
+    "file_name": "File name",
+    "description": "Description",
+    "Source-Organisation": "Transferring body",
```
anthonyhashemi: I guess we had the option of either using this or the static field `transferring_body_name` right? Just one of those duplicate data points we have, but this is fine

app/__init__.py
```diff
@@ -16,6 +16,19 @@
 s3 = FlaskS3()
 
 
+opensearch_field_name_map = {
```
anthonyhashemi: do we not already have some map to take the fields in the FE to the query format? If so, we should combine the two. If not, then we should prtobably utilise this there, right?

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -251,6 +251,9 @@ a.govuk-link {
     ~ tr.hidden-row-#{$i} {
     display: table-row;
   }
+  tr:has(> td > div > details.details-toggle-#{$i}[open]) .field-count-#{$i} {
```
anthonyhashemi, line 254: so this is the little leftover change from the previous ticket, making sure we dont show field-count when the accordion is closed 👍

app/__init__.py
anthonyhashemi: just had a look through and it seems like it doesn't get rid of other fields, it still maps fields not in this map to some presentation.

Really I think we should be filtering the search to only look for these fields

app/__init__.py
```diff
@@ -47,10 +35,9 @@ def clean_tags_and_replace_highlight_tag(text, highlight_tag):
 
 def format_opensearch_field_name(field):
     """Format the name of an OpenSearch field using a map or dynamically"""
-    if field in opensearch_field_name_map:
-        return opensearch_field_name_map[field]
-    else:
-        return field.replace("_", " ").capitalize()
+    return opensearch_field_name_map.get(
+        field, field.replace("_", " ").capitalize()
```
anthonyhashemi: I dont't think we should have a fallback. We should never have any other field and if we do, there's a problem and we should be notified. This fallback would mean this error would happen silently.

app/__init__.py
```diff
@@ -10,6 +10,7 @@
 
 from app.logger_config import setup_logging
 from app.main.db.models import db
+from app.main.util.search_utils import opensearch_field_name_map
```
anthonyhashemi: can we capitalize `opensearch_field_name_map` to follow best pep standards for python constants: https://peps.python.org/pep-0008/#constants

app/__init__.py
```diff
@@ -35,9 +35,7 @@ def clean_tags_and_replace_highlight_tag(text, highlight_tag):
 
 def format_opensearch_field_name(field):
     """Format the name of an OpenSearch field using a map or dynamically"""
```
anthonyhashemi: this docstring needs updating, "or dynamically" is no longer true

app/tests/test_jinja_filters.py
```diff
@@ -121,18 +121,16 @@ def test_clean_tags(input_text, expected_output):
...
         ("date_last_modified", "Record date"),
-        ("unknown_field", "Unknown field"),
-        ("another_field", "Another field"),
-        ("custom_field_name", "Custom field name"),
         ("citeable_reference", "Citeable reference"),
         ("series_name", "Series name"),
         ("transferring_body_description", "Transferring body description"),
         ("consignment_reference", "Consignment ref"),
         # edge cases
         # empty string should return empty string
```
anthonyhashemi: this comment is outdated now. Do we even need these comments?

app/tests/test_jinja_filters.py
```diff
@@ -125,10 +125,7 @@ def test_clean_tags(input_text, expected_output):
         ("series_name", "Series name"),
         ("transferring_body_description", "Transferring body description"),
         ("consignment_reference", "Consignment ref"),
-        # edge cases
-        # empty string should return empty string
         ("", None),
-        # single word, no underscore
         ("singleword", None),
         ("unknown_field", None),
```
anthonyhashemi: singleword is probably uneeded, unknown_field field is sufficient, as there is no distinction as to what they are trying to prove: both are showing that a string not in the map results in None. Additionaly I think we should just error on this scenario like I mentioned before, rather than fallback to None, as otherwise if there is an odd scenario where there is an extra field then we would not know about it unless a user raises it with us but it is completely unexpected, so erroring with a 500 makes sense to me!

So OPENSEARCH_FIELD_NAME_MAP[field] would do that instead of OPENSEARCH_FIELD_NAME_MAP.get(field)

anthonyhashemi (approved):

PR #646, opened 2024-11-11
Ayr 1297/audit app logs

colinbowen:
## Changes in this PR
- Update logger_config.py to include the new setup_logging function that configures app_logger and audit_logger and logging levels.
- Update readme on usage of audit logger and app logger

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1297

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/logger_config.py
+++ b/app/logger_config.py
@@ -47,35 +47,27 @@ def setup_logging(app):
         "%(levelname)s in %(module)s: %(message)s"
     )
 
+    app_logger = logging.getLogger("app_logger")
+    app_logger.setLevel(logging.INFO)
+    audit_logger = logging.getLogger("audit_logger")
+    audit_logger.setLevel(logging.INFO)
+
     if os.getenv("ENV") == "aws":
-        app_logger = logging.getLogger("app_logger")
-        app_logger.setLevel(logging.INFO)
         app_handler = CloudWatchHandler(
             log_group="app-logs", stream_name="app-log-stream"
         )
-        app_handler.setFormatter(formatter)
-        app_logger.addHandler(app_handler)
-
-        audit_logger = logging.getLogger("audit_logger")
-        audit_logger.setLevel(logging.INFO)
         audit_handler = CloudWatchHandler(
             log_group="audit-logs", stream_name="audit-log-stream"
         )
-        audit_handler.setFormatter(formatter)
-        audit_logger.addHandler(audit_handler)
-
     else:
-        app_logger = logging.getLogger("app_logger")
-        app_logger.setLevel(logging.INFO)
         app_handler = logging.StreamHandler()
-        app_handler.setFormatter(formatter)
-        app_logger.addHandler(app_handler)
-
-        audit_logger = logging.getLogger("audit_logger")
-        audit_logger.setLevel(logging.INFO)
         audit_handler = logging.StreamHandler()
-        audit_handler.setFormatter(formatter)
-        audit_logger.addHandler(audit_handler)
+
+    app_handler.setFormatter(formatter)
+    audit_handler.setFormatter(formatter)
+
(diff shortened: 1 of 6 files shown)


app/logger_config.py
```diff
@@ -12,15 +14,68 @@ def format(self, record):
...
+
+        audit_logger = logging.getLogger("audit_logger")
+        audit_logger.setLevel(logging.INFO)
+        audit_handler = CloudWatchHandler(
+            log_group="audit-logs", stream_name="audit-log-stream"
+        )
+        audit_handler.setFormatter(formatter)
+        audit_logger.addHandler(audit_handler)
+
+    else:
```
anthonyhashemi: can we move all the shared bits out of the if-else block to reduce duplication / simplify things. I think just the `audit_handler =` and `app_handler =` lines need to be conditional, right?

app/logger_config.py
```diff
@@ -12,15 +14,68 @@ def format(self, record):
         else:
             record.url = None
             record.remote_addr = None
-
         return super().format(record)
 
 
+class CloudWatchHandler(logging.Handler):
```
anthonyhashemi: Have you deployed this to test-one to test? From reading some comments, around SO suggesting this, it seems uncertain whether it is recommended to do this /whether it would work:
https://stackoverflow.com/questions/46101977/redirect-aws-lambda-logs-to-a-particular-log-group-in-cloudwatch/77214307#77214307
https://stackoverflow.com/questions/68544678/single-lambda-multiple-cloudwatch-log-groups

colinbowen: Did see those stackoverflow links earlier, I did want to test this on test one to confirm as it looks like we may have a situation where we have the created log groups and the default lambda log group (the first link)

anthonyhashemi: Okay, I suggest deploying from this branch to test to see if it works there

app/logger_config.py
```diff
@@ -12,15 +14,60 @@ def format(self, record):
...
     )
-    default_handler.setFormatter(formatter)
 
-    app.logger.setLevel(logging.INFO)
+    app_logger = logging.getLogger("app_logger")
+    app_logger.setLevel(logging.INFO)
+    audit_logger = logging.getLogger("audit_logger")
+    audit_logger.setLevel(logging.INFO)
+
+    if os.getenv("ENV") == "aws":
```
anthonyhashemi: I think we can hook into the existing CONFIG_SOURCE env var which we could use to distinguish between AWS deployment and non.

Ideally we could keep all AWS specific code in the Config classes but I don't think we can define `app_handler` and `audit_handler` on it, as the flask config object needs to be serialisable so the properties can't be objects.

aside: our calls to s3 for getting objects are directly in the routes so its not actually true that all boto calls are abstracted to the Config classes but as much as possible would be ideal. (I'd like to abstract the s3 calls to a FileHandler class; we could then follow a similar pattern to the Config class there and in any other place where the call to AWS can't be made at startup)

app/main/routes.py
```diff
@@ -760,7 +760,7 @@ def download_record(record_id: uuid.UUID):
             as_attachment=True,
             download_name=download_filename,
         )
-    current_app.logger.info(
+    current_app.app_logger.info(
```
anthonyhashemi: should be current.audit_logger.info

colinbowen: Revised 👍

colinbowen:
- Remove newly created log group and log streams and move to use a filter which is exported directly to S3.

anthonyhashemi (approved):

app/logger_config.py
anthonyhashemi: unneeded now 👌

colinbowen: Discussed PR on call

PR #666, opened 2024-11-25
Use https for iiif manifest urls not http

anthonyhashemi:
- Use https for iiif manifest urls not http
- Use consistent context referebce if presentation api 3


colinbowen (dismissed):

anthonyhashemi:
Codecov diff went down, not sure why but definitely no issue with this PR's test coverage.

ltrasca (approved):
lgtm

PR #668, opened 2024-11-25
AYR-1356 - Sorting for new OS implementation

ltrasca:
## Changes in this PR
- added logic and tests for sorting in the context of the new OpenSearch implementation

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1356

- Multi-match Opensearch field boosting: https://opensearch.org/docs/latest/search-plugins/sql/full-text/


app/tests/test_search.py
```diff
@@ -1698,16 +1698,17 @@ def test_search_transferring_body_with_search_term_happy_path_inner_table(
         assert response.status_code == 200
         soup = BeautifulSoup(response.data, "html.parser")
         select = soup.find("select", {"id": "sort"})
-        option = select.find("option", selected=True)
-        option_value = option.get("value")
+        # BUG: value not being saved in dropdown
```
anthonyhashemi, line 1701: is there a followup PR ready for this?

app/tests/test_search_utils.py
```diff
@@ -119,39 +146,119 @@ def test_format_opensearch_results(results, expected):
...
+            [
+                *fields_without_file_name,
+                "file_name^3",
             ],
+            "file_name",
+            [{"_score": {"order": "desc"}}],
+        ),
+        (
+            "invalid_area",
+            [*fields_without_file_name, "file_name^3"],
```
anthonyhashemi: why does an invalid area result in fields without file name

app/tests/test_search.py
ltrasca: @anthonyhashemi not yet but will create a ticket soon and make a PR for the fix

anthonyhashemi: I dont think a new ticket is needed, just a PR for that

app/tests/test_search_utils.py
anthonyhashemi: spoke offline, makes sense.

app/main/util/search_utils.py
```diff
@@ -56,15 +56,50 @@ def post_process_opensearch_results(results):
...
+def remove_field(fields, field_name):
+    return [field for field in fields if field_name != field]
+
+
+def get_open_search_fields_to_search_on_and_sorting(
+    search_area, sort="file_name"
+):
+    """Retrieve a list of fields depending on the search area (all fields, metadata, record, etc.) and sorting."""
+    sort_order = "asc" if sort == "least_matches" else "desc"
+    sorting = [{"_score": {"order": sort_order}}]
```
anthonyhashemi: as discussed on call, probably could do with a refactor as quite hard to understand why everything is happening here atm.

something like

```
def get_open_search_fields_to_search_on_and_sorting(search_area, sort="file_name"):
    """
    Retrieve fields and sorting configuration for an OpenSearch query based on the search area
    (all fields, metadata, record, etc.) and sorting preference.
    """
    # Sort option configuration
    sort_option_map = {
        "file_name": {
            "boosts": {"file_name": 3},
            "order": "desc",
        },
        "description": {
            "boosts": {"description": 3, "file_name": 2},
            "order": "desc",
        },
        "metadata": {
            "boosts": {"file_name": 0.2, "content": 0.1},
            "order": "desc",
        },
        "content": {
            "boosts": {"content": 3, "file_name": 2},
            "order": "desc",
        },
        "least_matches": {
            "boosts": {},  # No boosting
            "order": "asc",  # Ascending order for least matches
        },
        "most_matches": {
            "boosts": {},  # No boosting
            "order": "desc",  # Descending order for most matches
        },
    }

    fields = determine_fields_by_search_area(search_area)
    fields = apply_boosts_for_sorting(fields, sort, sort_option_map)
    sorting = get_sorting_config(sort, sort_option_map)
    return fields, sorting


def determine_fields_by_search_area(search_area):
    """
    Determine the base fields to use based on the search area.
    """
    all_fields = list(OPENSEARCH_FIELD_NAME_MAP.keys())
    if search_area == "metadata":
        return get_filtered_list(all_fields, ["file_name", "content"])
    elif search_area == "record":
        return ["content"]
    return all_fields


def apply_boosts_for_sorting(fields, sort, sort_option_map):
    """
    Adjust fields by applying boosts or penalties based on the sorting preference.
    """
    sort_config = sort_option_map.get(sort, {})
    boost_map = sort_config.get("boosts", {})

    # Exclude fields to be boosted from the base fields
    fields = [field for field in fields if field not in boost_map]
    # Apply boosts to the boosted fields
    return apply_field_boosts(fields, boost_map)


def get_sorting_config(sort, sort_option_map):
    """
    Get the sorting configuration based on the sort preference.
    """
    sort_config = sort_option_map.get(sort, {})
    sort_order = sort_config.get("order", "desc")  # Default to descending
    return [{"_score": {"order": sort_order}}]


def apply_field_boosts(fields, boost_map):
    """
    Apply boost values to fields as per the boost map.
    """
    return [f"{field}^{boost_map.get(field, 1)}" for field in fields]
```

what do you think?

ltrasca: @anthonyhashemi looks good, one thing this is that `fields = [field for field in fields if field not in boost_map]` shouldn't be needed anymore in this (would remove boosted fields from the array and inside apply_field_boosts these wouldnt be found by boost_map.get(field, 1))

going to add 👍

anthonyhashemi: amazing

anthonyhashemi (approved):

PR #674, opened 2024-11-28
AYR-1348 - Page view logs

ltrasca:
## Changes in this PR
- Added middleware for logging page views on authenticated pages plus tests to cover the changes

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1348

## Screenshots of UI changes

N/A


app/tests/test_middlewares.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @app.route(route_path, methods=[method])
+    @log_page_view
+    def dynamic_route():
+        return route_function()
+
+    with app.app_context():
+        current_app.audit_logger = mock_logger
+
+    if user_id_in_session is not None:
```
anthonyhashemi: can we split this out into separate focussed tests? i.e no conditions in the tests

app/main/middlewares/log_page_view.py
```diff
@@ -0,0 +1,20 @@
...
+
+from flask import current_app, request, session
+
+
+def log_page_view(route_function):
+    @wraps(route_function)
+    def wrapper(*args, **kwargs):
+        user_id = session.get("user_id", "anonymous")
+        log_data = {
+            "event": "page_view",
```
anthonyhashemi: some thoughs:

1. every authenticated route has this decorator and I think that makes sense but the download route is not a "page". 
2. the download route already has the audit log about the user downloading a file in a different structure
3. there is also route and method in the log data here. Does the event field add anything?

based on this I'm thinking we could just remove the event field and the existing audit log call for the user as long as the route includes the detail id for detail routes e.g. /download/id.

Worth me you and @colinbowen  having a chat about this

app/tests/test_middlewares.py
ltrasca: realized after taking a 2nd look the if statement wasn't really needed, I've updated the approach for replacing the session object and pushed changes

app/tests/test_middlewares.py
```diff
@@ -0,0 +1,83 @@
...
+def test_log_page_view(
+    app,
+    client: FlaskClient,
+    route_path,
+    method,
+    session_object,
+    route_function,
+    expected_response,
+    expected_log_data,
+):
```
anthonyhashemi, line 48: could you a docstring to the test so it is easier to understand at a glance?

app/tests/test_middlewares.py
```diff
@@ -0,0 +1,83 @@
...
+    mock_logger = MagicMock()
```
anthonyhashemi: we don't need to mock to assert the logged output.

We can use the caplog fixture as described in the pytyest docs here: https://docs.pytest.org/en/latest/how-to/logging.html#caplog-fixture

So we will not need this line or the app.context block

ltrasca: definitely, I've replaced with caplog - seems like we also do this elsewhere so we have a fixture setup

anthonyhashemi (approved):

PR #680, opened 2024-11-29
update logger config to json format

colinbowen:
## Changes in this PR
- Update logging config to use JSON
- Add updated readme to show examples of logs returned inside lambda
- Update tests

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1358

### After


app/logger_config.py
```diff
@@ -1,17 +1,37 @@
+import json
 import logging
 
 from flask import has_request_context, request
 
 
 class RequestFormatter(logging.Formatter):
     def format(self, record):
```
anthonyhashemi, line 9: I know it's not unique to this PR, but what does the argument `record` represent? If it's a specific record in db, this formatter does not make sense for routes not focussed on one specific record

colinbowen: The record should be a LogRecord class that gets created when logger.<> is called so it should work for any log event.

https://docs.python.org/3/library/logging.html#logging.LogRecord

README.md
```diff
@@ -630,7 +630,42 @@ current_app.audit_logger.error('Some error message')
 
 The logs from the webapp, when used as above are output as a stream to stdout in the following format:
 
+LOG ENTRY for a page view using the @log_page_view decorator:
+
 ```sh
+START RequestId: 4e2b263b-fc27-4db2-b87e-867fc4596e69 Version: $LATEST
+
+[DEBUG]	2024-12-11T12:43:26.106Z	4e2b263b-fc27-4db2-b87e-867fc4596e69	host found: [test-one.np.ayr.nationalarchives.gov.uk]
```
anthonyhashemi: can we omit this specific domain name in the readme

README.md
```diff
@@ -630,7 +630,42 @@ current_app.audit_logger.error('Some error message')
...
+
+{
+    "log_type": "audit_logger",
+    "timestamp": "2024-12-11 12:43:26,106",
+    "level": "INFO",
+    "remote_addr": "13.42.156.7",
+    "url": "https://test-one.np.ayr.nationalarchives.gov.uk/callback?code=24a260d2-b142-4db5-8675-c2acddcb7e7f.2d9cb0a3-8ce0-400a-afa9-c13298eb33ac.29a32c8f-97d2-4680-b58e-ac9120a4f94d&iss=https://auth.tdr-staging.nationalarchives.gov.uk/realms/tdr&session_state=2d9cb0a3-8ce0-400a-afa9-c13298eb33ac&state=",
```
anthonyhashemi: and the keycloak one

app/tests/test_middlewares.py
```diff
@@ -16,23 +18,32 @@
...
+            ),
         ),
         (
             "/post_route",
             "POST",
             {"user_id": "test_user"},
             lambda: "Post Response",
             b"Post Response",
-            f'{{"event": "api_request", "user_id": "test_user", "route": "{HOST}/post_route", "method": "POST"}}',
+            json.loads(
```
anthonyhashemi, line 43: why json.loads vs just defining the dict like before?

app/main/middlewares/log_page_view.py
```diff
@@ -13,6 +13,8 @@ def wrapper(*args, **kwargs):
             "user_id": user_id,
             "route": request.url,
             "method": request.method,
+            "caller_function": route_function.__name__,
```
anthonyhashemi: Is there a reason we can't move most of this info from this decorator and put it directly in a AudiRequestFormatter. I think we can access all this info there. Can then differentiate between audit and app logs by having an AuditRequesatFormatter and AppRequestFormatter.

That way whenever someone uses the audit_logger, all audit info will be included by default and users of it won't need to remember. Saves us needing this intermediary configuration in this log_page_view decorator. Unless there is a reason audit_logger wouldn't always want all this info?

app/tests/test_middlewares.py
colinbowen: Either way works, though it would be good to load the dict directly

anthonyhashemi: cool

app/main/middlewares/log_page_view.py
colinbowen: Looking at moving it to the audit log which is a good idea

ltrasca (approved):
looks good

PR #688, opened 2024-12-03
Create consignment indexing lambda for opensearch

anthonyhashemi:
## Changes in this PR

- Added consignment indexing lambda
- Adjusted the indexing dockerfile to point to this new lambda function, not the individual file indexer.

Have left the individual file indexing lambda code in the repo as we may want to do some work at some point to add an additional lambda function to leverage it for when we want to run the indexer on specific files that failed during consignment indexing but for now we can just rerun the consignment indexer.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1309


colinbowen (approved):
Looks good 👍

PR #697, opened 2024-12-10
AYR-1304 - Expand the list of textract formats

ltrasca:
## Changes in this PR
- Added more file formats that textract can extract text from
- Added tests to cover text related formats

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1304

## Screenshots of UI changes

N/A


anthonyhashemi (approved):
Nice!

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -7,11 +7,40 @@
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
+SUPPORTED_TEXTRACT_FORMATS = [
```
anthonyhashemi, line 10: so is this all the formats it supports?

anthonyhashemi: also which ones do we not add tests for?

ltrasca: @anthonyhashemi yes should be all of them, taken from this list

image

At the moment tests dont cover the ones that require system dependencies like Tesseract or antiword

anthonyhashemi: Can you create a ticket for writing e2e tests for those please. Not to do now. We need to think of the best way to run e2e tests for the indexing.

ltrasca: @anthonyhashemi sure, created a ticket

PR #710, opened 2024-12-17
update e2e tests

colinbowen:
### After

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -35,6 +35,7 @@ def utils():
 @pytest.fixture
 def page(page, request) -> Page:
     page.context.set_default_timeout(10000)
+    page.set_default_timeout(10000)
     if "test_css_" not in request.node.name and callable(request.node.obj):
 
         def route_intercept(route):
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -34,7 +34,7 @@ def utils():
 
 @pytest.fixture
 def page(page, request) -> Page:
-    page.context.set_default_timeout(5000)
+    page.context.set_default_timeout(10000)
     if "test_css_" not in request.node.name and callable(request.node.obj):
 
         def route_intercept(route):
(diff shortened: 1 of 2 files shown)


ltrasca (dismissed):

e2e_tests/conftest.py
```diff
@@ -55,10 +56,19 @@ def create_user_page(
...
-        page.goto("/sign-in")
+        page.wait_for_timeout(10000)
+        page.set_default_timeout(10000)
+        page.goto("/sign-in", timeout=10000)
         page.get_by_label("Email address").fill(username)
         page.get_by_label("Password").fill(password)
         page.get_by_role("button", name="Sign in").click()
+
+        # keycloak local
+        # page.get_by_label("Username or email").click()
```
anthonyhashemi, line 67: can we remove the comments

e2e_tests/conftest.py
```diff
@@ -94,17 +104,28 @@ def _create_keycloak_user(groups, user_type):
...
+                    "enabled": True,
+                    "groups": groups,
+                    "credentials": [{"value": user_pass, "type": "password"}],
+                }
+            )
+        except Exception as e:
+            print(f"Error creating user: {e}")
+            raise  # Re-raise the exception to fail the test
+
+        if not user_id:
```
anthonyhashemi, line 125: does this block ever get hit if you raise the exception above?

e2e_tests/navigation_journeys/test_search_to_record_journey.py
```diff
@@ -24,7 +24,9 @@ def test_search_to_record(aau_user_page: Page):
 
     aau_user_page.get_by_role("textbox").first.fill("a")
     aau_user_page.get_by_role("button", name="Search").click()
-    expect(aau_user_page).to_have_url("search_results_summary?query=a")
+    expect(aau_user_page).to_have_url(
```
anthonyhashemi, line 27: so this test is the one about nvaigating to a record i na search journey.

I think the tests in `class TestSearchResultsSummary:` (and possible the other test classes there) in `e2e_tests/test_search.py` would still be broken though, will those be done in a separate PR?

We just need to make sure that we cover the 2 ac in the ticket: https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-1236&text=AYR-1236

The AC do have 2 actions and currently that file splits out these scenarios into 2 test classes with the second starting from the summary results page, but do whatever feels best to you.

PR #712, opened 2024-12-18
AYR-1386 - File download timeout

ltrasca:
## Changes in this PR

- Replaced previous file transfer logic with a redirect towards a presigned url for the file

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1386


app/tests/test_download.py
```diff
@@ -38,17 +38,10 @@ def test_invalid_id_raises_404(self, client: FlaskClient):
...
     ):
-        """
-        Given a File in the database with corresponding file in the s3 bucket
-        When a standard user with access to the file's transferring body makes a
-            request to download record
-        Then the response status code should be 200
-        And the file should contain the expected content
-        And the downloaded filename should be the File's CiteableReference with extension
-        """
+        """ """
```
colinbowen: The function name for this pretty clear, I still think that having a docstring here would be useful.

app/tests/test_download.py
```diff
@@ -60,107 +53,35 @@ def test_download_record_standard_user_with_citable_reference_with_file_extensio
...
     ):
-        """
-        Given a File in the database with corresponding file in the s3 bucket
-        When a standard user with access to the file's transferring body makes a
-            request to download record
-        Then the response status code should be 200
-        And the file should contain the expected content
-        And the downloaded filename should be the filename with extension
-        """
+        """ """
```
colinbowen: Same here for the docstring.

app/main/routes.py
```diff
@@ -733,48 +730,31 @@ def download_record(record_id: uuid.UUID):
...
+        )
     except Exception as e:
-        current_app.app_logger.error(f"Error reading S3 file content: {e}")
+        current_app.app_logger.error(f"Failed to generate presigned URL: {e}")
         abort(500)
 
-    content_type = s3_file_object.get("ContentType", "application/octet-stream")
-
-    if render:
-        return send_file(
```
anthonyhashemi, line 764: does universal viewer render still work?

app/main/routes.py
```diff
@@ -733,48 +730,31 @@ def download_record(record_id: uuid.UUID):
...
+            ExpiresIn=3600,
+        )
     except Exception as e:
-        current_app.app_logger.error(f"Error reading S3 file content: {e}")
+        current_app.app_logger.error(f"Failed to generate presigned URL: {e}")
         abort(500)
 
-    content_type = s3_file_object.get("ContentType", "application/octet-stream")
-
-    if render:
```
anthonyhashemi, line 763: This condition block was already redundant anyway wasn't in, since we moved render to use presigned urls directly.

app/tests/test_download.py
```diff
@@ -38,17 +38,10 @@ def test_invalid_id_raises_404(self, client: FlaskClient):
         assert response.status_code == 404
 
     @mock_aws
-    def test_download_record_standard_user_with_citable_reference_with_file_extension(
```
anthonyhashemi, line 41: we still need the test cases for `citable_reference_with_file_extension` and `citable_reference_without_file_extension` unless we move the logic for the downloaded_filename into a helper function and test that

app/main/routes.py
```diff
@@ -733,48 +731,32 @@ def download_record(record_id: uuid.UUID):
...
     try:
-        s3_file_object = s3.get_object(Bucket=bucket, Key=key)
-    except Exception as e:
-        current_app.app_logger.error(f"Failed to get object from S3: {e}")
-        abort(404)
-
-    download_filename = file.FileName
+        s3.head_object(Bucket=bucket, Key=key)
+    except s3.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] == "404":
```
anthonyhashemi, line 741: currently this will continue without raising an error if the error from s3 is not a 404. Maybe we should raise any other error as a 500 unexpected error?

app/main/routes.py
```diff
@@ -733,48 +731,37 @@ def download_record(record_id: uuid.UUID):
...
-        abort(404)
-
-    download_filename = file.FileName
-
-    if file.CiteableReference:
-        if len(file.FileName.rsplit(".", 1)) > 1:
-            download_filename = (
-                file.CiteableReference + "." + file.FileName.rsplit(".", 1)[1]
+        s3.head_object(Bucket=bucket, Key=key)
+    except s3.exceptions.ClientError as e:
```
anthonyhashemi: can you add tests for these 2 exception blocks

app/main/routes.py
```diff
@@ -51,7 +50,6 @@
     generate_pdf_manifest,
     get_download_filename,
     get_file_details,
-    get_file_mimetype,
```
anthonyhashemi, line 54: As mentioned in the other PR, since you are removing the only use of this function `get_file_mimetype` here, you can go ahead and delete the function definition itself as well

anthonyhashemi (approved):

PR #730, opened 2025-01-08
Ayr 1422/manifest error

colinbowen:
## Changes in this PR
- Add CloudFront distribution to CSP policy
- Allow UV scripts in CSP policy
- Allow CSP values to be read from a .env or STS depending on configuration option set

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1422

### After


anthonyhashemi (approved):
Looks good, nice abstraction 👍

PR #745, opened 2025-01-22
AYR-1303 - Extent UV types

ltrasca:
## Changes in this PR
- Extended the range of file types/extensions that UV is given at manifest creation

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1303


app/tests/test_render_utils.py
```diff
@@ -11,11 +12,25 @@
...
+        ("jpg", "image/jpg"),
+        ("gif", "image/gif"),
+        # test a few other types (code files, archives, multimedia)
+        ("py", "application/py"),
+        ("zip", "application/zip"),
+        ("mp3", "application/mp3"),
+    ],
+)
+def test_get_file_mimetype(file_type, expected_mimetype):
+    assert get_file_mimetype(file_type) == expected_mimetype
```
anthonyhashemi: Can we have a test case for an unsupported filtetype

ltrasca: @anthonyhashemi sure, been thinking about this one slightly, would an unsupported mime type just be application/octet-stream by default?

anthonyhashemi: great question, but actually I think this function is no longer used once your other PR is merged in. That's right?

anthonyhashemi: so in that case let's just remove it in that PR and then won't need to have any changes related to it here

app/main/routes.py
```diff
@@ -816,9 +818,9 @@ def generate_manifest(record_id: uuid.UUID):
     filename = file.FileName
     file_type = filename.split(".")[-1].lower()
 
-    if file_type == "pdf":
+    if file_type in UNIVERSAL_VIEWER_SUPPORTED_DOCUMENT_TYPES:
```
anthonyhashemi, line 821: would be good to test this `generate_manifest` route for this changed logic, i.e setup the test with different doc types, some docs, some images, an unsupported

anthonyhashemi (approved):

PR #751, opened 2025-01-28
Ayr 1236/os e2e tests

colinbowen:
## Changes in this PR
- Updated search E2E tests to align with the new structure of the search results table.
- Added fuzzy search tests 
- Updated playwright version

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1236

## Screenshots of UI changes
N/A

### After


e2e_tests/test_pagination.py
```diff
@@ -138,17 +142,22 @@ def test_search_transferring_body_pagination_get_ellipses_page(
...
+        assert page_links[-1].text_content() == last_page
+
         ellipsis_link = aau_user_page.locator(
             ".govuk-pagination__item--ellipses"
         ).all()
-        assert ellipsis_link[0].inner_text() == "…"
+        if ellipsis_link:
+            assert ellipsis_link[0].inner_text().strip() == "…"
+        else:
+            print("No ellipses links were found in the pagination.")
```
anthonyhashemi, line 157: If we have multiple scenarios, we should split them into different tests rather than add conditional to the test function, to keep the tests easier to understand

e2e_tests/pyproject.toml
```diff
@@ -12,6 +12,7 @@ pytest-playwright = "^0.5.0"
 PyJWT = "^2.8.0"
 itsdangerous = "^2.2.0"
 python-keycloak = "^4.1.0"
+playwright = "^1.49.1"
```
anthonyhashemi, line 15: did we not already specify playwright?

anthonyhashemi: ah we specifiy pytest-playwright which installs playwright I assume. What do we need from this pinned version of playwright?

e2e_tests/test_login.py
```diff
@@ -48,7 +48,6 @@ def test_sign_in_succeeds_when_valid_credentials(
         "sub",
         "typ",
         "azp",
-        "session_state",
```
anthonyhashemi, line 51: Have you run this against test-one and this isnt there? Or is this just missing on local keycloak?

e2e_tests/test_record.py
```diff
@@ -177,7 +177,7 @@ def test_record_download_record(
 
         Given the user navigates to the record page with ID "100251bb-5b93-48a9-953f-ad5bd9abfbdc"
         When the user clicks the "Download record" button
-        Then the file "TSTA 1_ZD5B3S.doc" should be downloaded
+        Then the file "file-a2.txt" should be downloaded
```
anthonyhashemi, line 180: do we know why this changed?

e2e_tests/test_search.py
```diff
@@ -149,3 +176,87 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+
+    @property
+    def browse_transferring_body_route_url(self):
+        return "/search/transferring_body"
+
+    @property
+    def transferring_body_id(self):
+        return "c3e3fd83-4d52-4638-a085-1f4e4e4dfa50"
+
+    def test_test_search(self, aau_user_page: Page):
```
anthonyhashemi: why do we have test_test rather than just test? (here and the test function below)

e2e_tests/test_search.py
```diff
@@ -149,3 +176,87 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+        """
+        url = f"{self.browse_transferring_body_route_url}/{self.transferring_body_id}#browse-records"
+
+        standard_user_page.goto(url)
+        expect(standard_user_page).to_have_url(url)
+
+        standard_user_page.locator("#search-input").click()
+        standard_user_page.locator("#search-input").fill("a")
+        standard_user_page.get_by_role("button", name="Search").click()
+        standard_user_page.locator("#search_filter").fill("samplt")
```
anthonyhashemi: are we trying to test the filter logic here? If not, I suggest we minimise the actions the user is making, and just do one search with a typo. Remember we can just put multiple terms in the search box directly

e2e_tests/test_search.py
```diff
@@ -149,3 +176,87 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+        standard_user_page.get_by_role("button", name="Apply terms").click()
+        standard_user_page.locator("#search_filter").fill("a1")
+        standard_user_page.get_by_role("button", name="Apply terms").click()
+        rows = standard_user_page.locator("tbody .govuk-table__row")
+        assert rows.count() == 18
```
anthonyhashemi: re everything below:

If we cover the tags/search terms displaying in the ui in the normal flask tests already I dont think we need to assert on them here at all - they aren't dependent on external apis

e2e_tests/test_search.py
```diff
@@ -149,3 +176,87 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
         standard_user_page.get_by_role("link", name="Clear all").click()
 
         expect(standard_user_page).to_have_url(url)
+
+
+class TestOpenSearchResults:
```
anthonyhashemi: I think we shouldn't refer to opensearch. The e2e tests should be from the perspective of a user, who should not care about how we implement search under the hood.

e2e_tests/test_search.py
```diff
@@ -149,3 +176,87 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+    def test_test_search(self, aau_user_page: Page):
+        """
+        Given a valid search query is entered on the search page
+        When the query is submitted
+        Then the search_results_summary page should display relevant results,
+        ensuring the results are fetched from OpenSearch and shown with correct metadata formatting
+        And when a transferring body is clicked on
+        Then the search_results/transferring_body/<body_id> page should display the filtered results correctly
+        """
+        url = f"{self.search_route_url}?query=a&search_area=everywhere"
```
anthonyhashemi: is the purose of this test any different to `test_search_returns_results_summary` in the above test class?

e2e_tests/test_search.py
colinbowen: Fixed and renamed to just test_

e2e_tests/test_record.py
colinbowen: Looks to be just the test data, I think this may have been from the previous database we was using

e2e_tests/test_search.py
colinbowen: Changed to just test for the fuzzy logic

e2e_tests/pyproject.toml
colinbowen: There was a build issue installing the browsers needed (Chromium / Webkit / Firefox) when using playwright that comes with python-keycloak. Pinning playwright itself to the most recent version fixed it. 

A similar issue had a suggestion to install the latest playwright resolved it: https://github.com/microsoft/playwright/issues/30503

e2e_tests/test_record.py
anthonyhashemi: but the record id hasn't changed? 

I was thinking it could be to do with how we moved over to use use presigned urls rather than proxying the file for download but I think we were meant to keep the filename the same as before. Also, previously the filename has the citable ref whereas on the right it's a filename, the only thing that I can't figure out is why the file extensions are different. Maybe this is flagging a bug. I dont think we should change the expected filename just to make the test pass without understanding this further

I am going to check this record id on test-one

anthonyhashemi: Okay it sees the citable ref for this record is not currently being picked up on test one. Thats odd as I dont think we changed the db. Im going to query the db directly to take a look

anthonyhashemi: okay I remember now that we removed some data from the test db dump and so we must have deleted it then for some reason (intentionally or unintentionally).

So all good

e2e_tests/test_search.py
```diff
@@ -149,3 +176,44 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+        expect(standard_user_page).to_have_url(url)
+
+        standard_user_page.locator("#search-input").click()
+        standard_user_page.locator("#search-input").fill("smaplt")
+        standard_user_page.get_by_role("button", name="Search").click()
+        rows = standard_user_page.locator("tbody .govuk-table__row")
+        assert rows.count() == 20
+
+        found = False
+        for i in range(rows.count()):
```
anthonyhashemi: it would be good to be able to assert that it presents itself as a highlight, so within the highlight tag elements, otherwise it could just be showing that sample was in a snippet but not in the highlight

e2e_tests/test_search.py
```diff
@@ -149,3 +176,40 @@ def test_click_on_clear_all_as_standard_user_redirects_to_browse_transferring_bo
...
+        standard_user_page.locator("#search-input").click()
+        standard_user_page.locator("#search-input").fill("smaplt")
+        standard_user_page.get_by_role("button", name="Search").click()
+        rows = standard_user_page.locator("tbody .govuk-table__row")
+        assert rows.count() == 20
+
+        tbody_locator = standard_user_page.locator("tbody.govuk-table__body")
+        inner_html = tbody_locator.inner_html()
+
+        assert "<mark>sample</mark>" in inner_html
```
anthonyhashemi, line 215: lovely

anthonyhashemi (dismissed):

e2e_tests/test_login.py
colinbowen: Yes run against test one, think this is the claims that have been removed. I'm looking in to confirm this

anthonyhashemi: did we intentionally remove this claim?

colinbowen: It's not something we've intentionally removed. The session state is usually set in a keycloak cookie when the user logs in or off from keycloak and holds the previous session info. From: https://github.com/keycloak/keycloak/issues/31651 says It represents the latest session_state returned by keycloak during a token response and has to be put in through a custom mapper. 

I can't see why it was there originally but it should usually be inside a refresh_token response. We should add a test to make sure it's there for refresh tokens

anthonyhashemi: fair, yeah I see its not in the token for any env now. Also not sure why it was there initially. And sure could be worth adding a refresh token test; both on the expected keys but also the expected time for expiry if possible, based on the config for that instance but happy for you to leave for this pr, up to you

colinbowen: I'll add the extra tests to this pr

anthonyhashemi (approved):

PR #783, opened 2025-02-21
AYR-1446 - Bold service name

ltrasca:
## Changes in this PR
- Modified service name styles in the header to be bold

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1446

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_header.scss
+++ b/app/static/src/scss/includes/_header.scss
@@ -24,6 +24,7 @@
     white-space: normal;
     margin-top: 0;
     margin-bottom: 0;
+    font-weight: bold;
 
     @include on-mobile {
       font-size: 1.8rem;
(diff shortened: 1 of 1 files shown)


Terry-Price (dismissed):
Hi @ltrasca I think GDS is a value of 700 rather than 'bold'.

anthonyhashemi (dismissed):

anthonyhashemi (approved):

anthonyhashemi (approved):

app/static/src/scss/includes/_header.scss
```diff
@@ -24,6 +24,7 @@
     white-space: normal;
     margin-top: 0;
     margin-bottom: 0;
+    font-weight: 700;
```
anthonyhashemi, line 27: have approved but we should start using variables like what we started in app/static/src/scss/includes/_variables.scss
and then even start using components etc, but first going to have convos about what sort of component library framework we want to use etc

ltrasca: @anthonyhashemi absolutely

PR #784, opened 2025-02-21
AYR-1460 - Record page text inconsistencies 

ltrasca:
## Changes in this PR
- Adjusted styles for "You are viewing" text on record page to be consistent with all the other pages
- Also adjusted styles for breadcrumbs to be consistent with other pages - this is because 1) the gap between the text above it and itself wasnt consistent, 2) the space between the above breadcrumbs and the file name seemed too small - an in-line layout seems more natural

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1460

### Before
image

### After
image


Terry-Price (approved):
Thanks @ltrasca - now lloks great and is consistent.

PR #788, opened 2025-02-24
AYR-1442 - Review search readable labelling

ltrasca:
## Changes in this PR

- Wrapped elements of the search component with a fieldset element
- Moved form wrapper outside the fieldset
- Made the previous text input label a heading element inside the legend of the fieldset
- Added an aria label to the search input to make up for the missing label element - ideally one should exist

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1442

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_overrides.scss
+++ b/app/static/src/scss/includes/_overrides.scss
@@ -15,3 +15,11 @@ a {
     font-size: 16px;
   }
 }
+
+.govuk-fieldset {
+  &__legend {
+    &--search {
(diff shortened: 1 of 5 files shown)


anthonyhashemi (approved):
Looks good, and made me relearn why we use fieldsets and legends: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend

PR #791, opened 2025-02-25
Update opensearch indexer lambda dockerfile dependencies

anthonyhashemi:
## Changes in this PR

- Update opensearch indexer lambda dockerfile base image to custom image (and thus explicit awslambdric calling)
- Upgrade to ptypon 3.13 and Use temporary fork of textract https://pypi.org/project/textract-py3/ which has fixed the dep issue with python 3.12 and above (see https://github.com/deanmalmgren/textract/issues/476#issuecomment-1676351111)

Reasoning:

bug found in prod where antiword was not installed when trying to extract text from .doc files, and similarly on staging for tesseract-ocr (on pdfs I think).

We were previously using an amazon linux base docker image. They only allow certain restricted packages on that distro. Things like antiword and tesseract-ocr were not actually being installed due to yum being silent when it could not find anything to install.

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1526
https://national-archives.atlassian.net/browse/AYR-1527

### DIFF
--- a/data_management/opensearch_indexer/requirements.txt
+++ b/data_management/opensearch_indexer/requirements.txt
@@ -3,6 +3,7 @@ opensearch-py==2.7.1
 requests-aws4auth==1.3.1
 SQLAlchemy==2.0.32
 pg8000==1.31.2
-textract==1.6.5
+textract-py3==2.1.0
+xlrd==1.2.0
 testing-postgresql==1.3.0
 psycopg2-binary==2.9.10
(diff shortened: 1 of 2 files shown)


data_management/opensearch_indexer/requirements.txt
```diff
@@ -3,6 +3,6 @@ opensearch-py==2.7.1
 requests-aws4auth==1.3.1
 SQLAlchemy==2.0.32
 pg8000==1.31.2
-textract==1.6.5
+textract-py3==2.1.0
```
ltrasca, line 6: failing unit test might have to do with the version of `xlrd` this new library uses, their `pyproject.toml` contains `xlrd = ">=1.2.0"` which might mean we are installing version 2.0.0+ of `xlrd` along with it, this version doesn't seem to support any other format than `.xls`

image

anthonyhashemi: good catch, I have specified xlsd 1.2.0 in the requirements.txt now to lock it down and have sent the maintainer of the fork a message about restricting the constraints!

ltrasca (approved):

PR #793, opened 2025-02-26
AYR-1531 - Extended BaseConfig file formats

ltrasca:
## Changes in this PR
- Removed mentions of supported file formats from the app itself and instead placed them inside of `BaseConfig` to establish a single source of truth (this refers to `UNIVERSAL_VIEWER_SUPPORTED_IMAGE_TYPES`, `UNIVERSAL_VIEWER_SUPPORTED_DOCUMENT_TYPES` and `UNIVERSAL_VIEWER_SUPPORTED_OTHER_TYPES`)
- Modified the endpoints that made use of the variables mentioned above to now use properties of the current app config instead
- `SUPPORTED_RENDER_EXTENSIONS` inside of `BaseConfig` is now all of the arrays mentioned above combined 

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1531

## Screenshots of UI changes
N/A


anthonyhashemi:
Have you tested this locally, can't remember how many filetypes we had in that dataset though?
You could be maybe deploy to the int instance from this branch and see how it looks there. We should expand the dev dataset if its not sufficent though

ltrasca:
> Have you tested this locally, can't remember how many filetypes we had in that dataset though? You could be maybe deploy to the int instance from this branch and see how it looks there. We should expand the dev dataset if its not sufficent though

@anthonyhashemi will test, made this pr a draft for now because some changes I made are breaking unit tests, will make it a normal pr once fixed and tested

anthonyhashemi (approved):

PR #795, opened 2025-02-27
refactor search transferring body func

colinbowen:
## Changes in this PR
- Remove the search_filter query parameter from url
- Using the query parameter to display user’s search terms. 
- Use url encoding for queries within URL

## JIRA ticket
- https://national-archives.atlassian.net/browse/AYR-1371

### After

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -57,6 +57,7 @@ from app.main.util.render_utils import (
 from app.main.util.search_utils import (
     build_search_results_summary_query,
     build_search_transferring_body_query,
+    check_additional_term,
     execute_search,
     extract_search_terms,
     get_open_search_fields_to_search_on_and_sorting,
@@ -577,13 +578,15 @@ def search_transferring_body(_id: uuid.UUID):
 
     additional_term = request.args.get("search_filter", "").strip()
 
+    check_additional_term(additional_term, query, request.args.copy(), _id)
+
     if additional_term:
         if " " in additional_term and not (
             additional_term.startswith('"') and additional_term.endswith('"')
         ):
             additional_term = f'"{additional_term}"'
 
-        query = f"{query}&{additional_term}" if query else additional_term
+        query = f"{query}+{additional_term}" if query else additional_term
 
         args = request.args.copy()
         args.pop("search_filter", None)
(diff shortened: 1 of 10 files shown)


app/main/routes.py
```diff
@@ -574,11 +574,29 @@ def search_transferring_body(_id: uuid.UUID):
     highlight_tag = f"uuid_prefix_{uuid.uuid4().hex}"
 
     query, search_area = get_query_and_search_area(request)
-    search_filter = (
-        request.args.get("search_filter", "").strip()
-        if request.args.get("search_filter")
-        else None
-    )
+
+    additional_term = request.args.get("search_filter", "").strip()
```
ltrasca, line 579: Would be good to break this logic for additional term into its own function for better readability & testability

app/tests/test_search.py
```diff
@@ -1375,7 +1375,7 @@ def test_search_transferring_body_display_filter_tray_standard_user(
 
         term1 = "TDR-2023-FI1"
         term2 = "first"
-        query = f"{term1},{term2}"
+        query = f"{term1}&{term2}"
```
ltrasca: are there any concerns around terms that might contain the symbol "&"?

app/main/util/search_utils.py
```diff
@@ -386,23 +384,30 @@ def build_search_transferring_body_query(
...
-    quoted_phrases = re.findall(r'"([^"]*)"', query)
+    quoted_phrases = []
+    single_terms = []
 
-    # Remove quoted phrases and split remaining terms by spaces
-    remaining_terms = re.sub(r'"[^"]*"', "", query).replace(",", " ")
-    single_terms = [
-        term.strip() for term in remaining_terms.split() if term.strip()
-    ]
+    query_parts = query.split("&")
```
ltrasca: as a pattern, Google uses "+" (which I think is just encoded symbol for space) as the symbol for concatenating search terms (as seen in the screenshot below for the search of "foo bar"), maybe this could be something more suitable than & we could borrow?

image

app/main/routes.py
colinbowen: Created a function to handle the additional term logic

app/tests/test_search.py
colinbowen: Yes, changed to use the unicode %2B (+) which is the separator we was originally using which is now url encoded

ltrasca (approved):

PR #801, opened 2025-03-04
AYR-1532 - Modifications to no results found page

ltrasca:
## Changes in this PR

- On `search_results_summary` page the breadcrumbs are now removed if there are not results
- The padding at the top of the `no_results_found` template is now been reduced
- "browse view" is now an anchor that leads to`/browse`

Note: the breadcrumbs are added in each individual page, meaning that they still exist inside other pages when there are no results, this made sense to me for `search/transferring_body` where the transferring body thats currently being searched within is important and should be displayed to the user @Terry-Price let me know your view on this

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1532

### Before
image

### After
image


Terry-Price (approved):
Thanks @ltrasca - looking good now.

PR #803, opened 2025-03-05
Fix expansion of record types to be supported for render

anthonyhashemi:
## Changes in this PR

Fix expansion of record types to be supported for render

Couple of the new files added:

Screenshot 2025-03-26 at 14 14 29

Screenshot 2025-03-26 at 14 33 17


ltrasca (dismissed):

anthonyhashemi:
@ltrasca @colinbowen Let's catch up on this, before deploying, we should fix the manifest generation, otherwise we'll be misleading users that we can render these docs when we can't, as currently it shows a blank screen in the universal viewer due to trying to create invalid manifests for different do. types, e.g currently trying to create pdf manifests for docx, txt etc.

ltrasca (approved):

PR #808, opened 2025-03-10
AYR-1491 - Pages form structure

ltrasca:
## Changes in this PR

- Split the main form that surrounded sorting, table and filters into just surrounding sorting and filtering individually.

Why?
- Having sorting, the table itself and the filtering section inside a single form may confuse users that make use of screen readers
- "Sorting" and "Filtering" and two different concepts (even if our backend logic we use unifies them into a single form) this is also true visually speaking for these two components, being placed in different areas of the page
- Long term, this approach should be more maintainable if there is ever a decision that filtering and sorting need to be split

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1491

## Screenshots of UI changes

N/A - there are no visual changes


app/templates/main/browse-all.html
```diff
@@ -9,88 +9,91 @@
...
-<form action="{{ url_for('main.browse', _anchor='browse-records') }}"
-      method="get">
-    <div class="browse-layout">
-        <!-- SORT -->
-        {% if num_records_found > 0 %}
-            <div class="govuk-form-group sort-container__form">
+<div class="browse-layout">
+    <!-- SORT -->
+    {% if num_records_found > 0 %}
+        <div class="govuk-form-group sort-container__form">
```
anthonyhashemi, line 15: SORT comment is more appropriate here, rather than outside if the if block, right?

app/templates/main/browse-consignment.html
```diff
@@ -18,87 +18,90 @@
...
-<form action="{{ url_for('main.browse_consignment', _id=request.view_args['_id'], _anchor='browse-records') }}"
-      method="get">
-    <div class="browse-layout">
-        <!-- SORT -->
-        {% if num_records_found > 0 %}
-            <div class="govuk-form-group sort-container__form">
+<div class="browse-layout">
+    <!-- SORT -->
+    {% if num_records_found > 0 %}
+        <div class="govuk-form-group sort-container__form">
```
anthonyhashemi, line 24: here too

app/templates/main/browse-transferring-body.html
```diff
@@ -14,79 +14,82 @@
...
-<form action="{{ url_for('main.browse_transferring_body', _id=request.view_args['_id'], _anchor='browse-records') }}"
-      method="get">
-    <div class="browse-layout">
-        <!-- SORT -->
-        {% if num_records_found > 0 %}
-            <div class="govuk-form-group sort-container__form">
+<div class="browse-layout">
+    <!-- SORT -->
+    {% if num_records_found > 0 %}
+        <div class="govuk-form-group sort-container__form">
```
anthonyhashemi, line 20: and here

app/templates/main/browse-series.html
```diff
@@ -15,68 +15,69 @@
...
-<form action="{{ url_for('main.browse_series', _id=request.view_args['_id'], _anchor='browse-records') }}"
-      method="get">
-    <div class="browse-layout">
-        <!-- SORT -->
-        {% if num_records_found > 0 %}
-            <div class="govuk-form-group sort-container__form">
+<div class="browse-layout">
+    <!-- SORT -->
+    {% if num_records_found > 0 %}
+        <div class="govuk-form-group sort-container__form">
```
anthonyhashemi, line 21: and here

anthonyhashemi:
Really appreciate the clear explanation about the motivation for this ticket, made it much easier to review. I'm wondering, is it possible to extract a results page template, as it looked like there was a lot of duplicate code, but they are quite similar, just different columns etc. Not for this PR necessarily, but could be worth creating a ticket for it

app/templates/main/browse-consignment.html
```diff
@@ -190,6 +193,6 @@ <h3 class="govuk-heading-s govuk-heading-s--consignment-filter">Record dates</h3
                                 </div>
                             </div>
                         </div>
-                        <!-- END FILTERS -->
-                    </div>
-                </form>
+                    </form>
```
anthonyhashemi, line 196: is it just me or does the indentation seem off?

ltrasca: @anthonyhashemi seems like the linter is messing up on this page for some reason 🤨 will investigate

anthonyhashemi (approved):
Will create a techdebt ticket to invewtdsigate why djlint is messing up the closing tags.

PR #809, opened 2025-03-10
AYR-1455 - Changes to dates filters styling

ltrasca:
## Changes in this PR

- Modified the `Transfer date` label across pages to match the styling of all the other labels we use in weight and size
- Removed the use of a <br> tag between the `Transfer date` label 
- By removing the tag mentioned above the gap between the `Transfer date` label and the input group below it has been decreased 
- Reduced the spacing between the two groups of date inputs to 1 rem (plus the padding of the label below it)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1455

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_browse-consignment.scss
+++ b/app/static/src/scss/includes/_browse-consignment.scss
@@ -162,6 +162,10 @@
     margin-bottom: 0;
   }
 
+  &--dates {
+    margin-bottom: 1rem;
+  }
+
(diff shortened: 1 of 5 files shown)


Terry-Price (approved):
Thanks Lucian, looks good,
was there another card to take out the white gaop bewteen all filter sections, so it becomes just one panel?
Perhaps that was on another card??

PR #821, opened 2025-03-17
AYR-1487 - Deepen UV window

ltrasca:
## Changes in this PR
- On desktop and tablet devices, increased the height of the UV window to 80vh from 60vh
- Removed download record button from the top of UV window per the catchup with @Terry-Price and @georgeopteryx 
- Increased default zoom of UV to 2
- Fixed case where styles of UV pagination would get overwritten by out own page styles and causing the page number input to be too large & obstruct other elements
- Distance between breadcrumbs and UV window should already be 1rem following changes from other older tickets

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1487

### Before
image

### After
image

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -118,9 +118,9 @@
         "filename": "app/templates/main/record-view.html",
         "hashed_secret": "b49139ad0afdb487229b83c6d3bda217b7e14977",
         "is_verified": false,
-        "line_number": 15
+        "line_number": 7
       }
     ]
   },
-  "generated_at": "2024-09-02T20:02:22Z"
+  "generated_at": "2025-03-17T06:28:13Z"
 }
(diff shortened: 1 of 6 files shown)


Terry-Price (commented):
Hi @ltrasca This is looking much better with the increased zoom particularly.
Please can I check if you added a border bottom to the window (I think we said 2px)
We can also make a card to display the nav tools so they are vertically centred in that narrow panel.

ltrasca:
> Hi @ltrasca This is looking much better with the increased zoom particularly. Please can I check if you added a border bottom to the window (I think we said 2px) We can also make a card to display the nav tools so they are vertically centred in that narrow panel.


Hey @Terry-Price I've updated the button group to be vertically centered as such 
image

Terry-Price (approved):
the pagination now looks great, thanks @ltrasca

PR #828, opened 2025-03-20
AYR-1552 - Overflowing text inside record arrangement

ltrasca:
## Changes in this PR
- Modified styles for `Record arrangement` component so text inside `li` elements does not overflow its container

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1552

### Before
image

### After
image


anthonyhashemi (approved):
Nice! another which should be nicely covered by the visual regression tests when we add 😎

PR #829, opened 2025-03-20
AYR-1522 - Row margins on mobile

ltrasca:
## Changes in this PR

- On mobile, on the browse all page, modified the 2nd row in the first column (that becomes visible on mobile) to have 5px less margin above and 5px more margin below it

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1522

### Before
image

### After
image


anthonyhashemi (approved):
I was going to suggest a test for this but this feels like it would be best served by a visual regression test? If you agree we can reintroduce the infrastructure for that in the next couple sprints in a dedicated ticket as obviously it's not a quick addition here, but if you think it could be served by a simpler test, obviously feel free to do that here. 

Approved

PR #839, opened 2025-03-27
AYR-1459 - UV chevrons

ltrasca:
## Changes in this PR
- Added some JS logic once UV is created to add roles to divs that have a specific class as "button" - this is so screen readers properly describe the function of the elements to the user
- Added a visible border to the chevrons while they are focused

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1459
### Before
image

### After
image

### DIFF
--- a/app/static/init.uv.js
+++ b/app/static/init.uv.js
@@ -91,4 +91,10 @@ document.addEventListener("DOMContentLoaded", function () {
       uvElement.style.padding = "0.25rem";
     }
   }
+
+  document.querySelectorAll(".btn").forEach((button) => {
+    if (button.tagName.toLowerCase() === "div") {
+      button.setAttribute("role", "button");
+    }
+  });
 });
--- a/app/static/src/scss/includes/_uv.scss
+++ b/app/static/src/scss/includes/_uv.scss
@@ -11,4 +11,9 @@
     width: fit-content;
     top: -10px;
   }
+
+  div.btn:focus {
+    outline: none;
+    border: 2px solid blue;
+  }
 }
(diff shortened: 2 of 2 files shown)


app/static/init.uv.js
```diff
@@ -91,4 +91,10 @@ document.addEventListener("DOMContentLoaded", function () {
       uvElement.style.padding = "0.25rem";
     }
   }
+
+  document.querySelectorAll(".btn").forEach((button) => {
```
anthonyhashemi, line 95: so this is an example of us fixing acc issues that ideally would be done by uv themselves. Going forward like we discussed I think we will hold off doing more of this and have a conversation with UV/ try make prs in their repo

app/static/src/scss/includes/_uv.scss
```diff
@@ -11,4 +11,9 @@
     width: fit-content;
     top: -10px;
   }
+
+  div.btn:focus {
```
anthonyhashemi, line 15: this is separate to the role attribute change of course and just fixes it how Terry wanted it right?

anthonyhashemi (approved):

app/static/src/scss/includes/_uv.scss
ltrasca: @anthonyhashemi yes its separate to the role change, but also relates to the DAC report where they wanted the focus state to be visible - overall its the same as the above comment, we should communicate with the UV team to make these pseudo-buttons (that are divs) actual button elements or do it ourselves, which would remove the need for basically both these changes

PR #840, opened 2025-03-27
AYR-1457 - Fixed style inconsistencies in filters

ltrasca:
## Changes in this PR
- Changed wording of "Filter" heading to "Filters" across all pages
- Normally `govuk-labels__filters` now have a font weight of 700
- Deleted instances where labels were wrapped by `h3` elements 
- Made all labels use consistent classes for styles
- Changed wording of "Date from" and "Date to" to "From date" and "To date" respectively (to be more consistent with FCS, for example)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1457

### Before
Browse all
image

Browse transferring body
image

Browse series
image

Browse consignment
image

### After
Browse all
image

Browse transferring body
image

Browse series
image

Browse consignment
image

### DIFF
--- a/app/static/src/scss/includes/_overrides.scss
+++ b/app/static/src/scss/includes/_overrides.scss
@@ -13,6 +13,11 @@ a {
 .govuk-label {
   &__filter {
     font-size: 16px;
+    font-weight: 700;
+  }
+
+  &__filter-date {
+    font-size: 16px;
   }
 }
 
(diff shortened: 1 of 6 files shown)


anthonyhashemi (approved):
Looks good. We should have a session to discuss how we would best like to test these sorts of changes. Might iust be - reintroduce snapshots, but we could easily assert on the html content like we do already in the flask tests

PR #848, opened 2025-04-02
Fix/e2e tests

colinbowen:
## Changes in this PR
- Update e2e tests

## JIRA ticket
N/A

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/e2e_tests/conftest.py
+++ b/e2e_tests/conftest.py
@@ -63,7 +63,7 @@ def utils():
 
 @pytest.fixture
 def page(page, request) -> Page:
-    page.context.set_default_timeout(10000)
+    page.context.set_default_timeout(15000)
     if "test_css_" not in request.node.name and callable(request.node.obj):
 
         def route_intercept(route):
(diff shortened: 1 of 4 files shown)


e2e_tests/conftest.py
```diff
@@ -63,7 +63,7 @@ def utils():
 
 @pytest.fixture
 def page(page, request) -> Page:
-    page.context.set_default_timeout(10000)
+    page.context.set_default_timeout(15000)
```
anthonyhashemi, line 66: was there an issue that increasing this fixed? I guess it's an indication of severely bad performance (probably lambda warmup time) if we are having to increasde to 15ms?

e2e_tests/test_browse_transferring_body.py
```diff
@@ -61,9 +61,9 @@ def test_browse_transferring_body_no_results_found(
         assert standard_user_page.inner_html(
             "text='Try changing or removing one or more applied filters.'"
         )
-        assert standard_user_page.inner_html(
-            "text='Alternatively, use the breadcrumbs to navigate back to the browse view.'"
-        )
+        assert standard_user_page.locator(
```
anthonyhashemi, line 64: nice, is_visible is always better 👌

anthonyhashemi (approved):

e2e_tests/conftest.py
colinbowen: Most (99%)  were completing within the previous time limit of 10s but some tests were flaky and caused failures to show up. Definitely due to lambda warmup time.

PR #859, opened 2025-04-09
AYR-1540 - Changes to fix indexer timeout

ltrasca:
## Changes in this PR
- Changed indexer logic to use threading
- Changed unit tests to fix above

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1540
## Screenshots of UI changes

N/A


data_management/opensearch_indexer/tests/test_consignment_lambda_handler.py
```diff
@@ -311,6 +312,8 @@ def test_lambda_handler_invokes_bulk_index_with_correct_file_data(
                     "file_id": str(file_3_id),
                 },
             ]
+
+            unittest.TestCase().assertCountEqual(args[0], expected_list)
```
anthonyhashemi: I've never seen `assertCountEqual` before and just read up on it, very cool, it works for unhashable objects unlike the option I was going to suggest of casting to set and asserting on sets!

refs:

- https://stackoverflow.com/questions/41605889/does-pytest-have-an-assertitemsequal-assertcountequal-equivalent
- https://docs.python.org/3.5/library/unittest.html#unittest.TestCase.assertCountEqual

anthonyhashemi:
Looks good but I know you are working to improve this so it is even more efficient so will hold of approving for now

anthonyhashemi:
Reposting comment from JIRA Ticket here for visibility:

Okay yeah I read into asyncio aioboto etc and their incompatibility with moto’s synchronous nature. Anyway, I think it is worth us thinking about the best way top optimise this even further as well, potentially using AWS’s step function distributed maps alongside either threading, multiprocessing or async/await will be worthwhile. So yeah for now, let’s go ahead with merging the initial change in and seeing if it fixes for this batch

anthonyhashemi (approved):
Sorry, forgot to approve when leaving the comment yesterday

PR #862, opened 2025-04-14
AYR-1563 - Overflowing table on consignment browse page

ltrasca:
## Changes in this PR
- Made content in the file name column break text anywhere and not just words

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1563

### Before
image

### After
image

### DIFF
--- a/app/templates/main/browse-consignment.html
+++ b/app/templates/main/browse-consignment.html
@@ -65,10 +65,10 @@
                             <tr class="govuk-table__row govuk-table__row-ref">
                                 <td class="govuk-table__cell govuk-table__cell--on-mobile--flex-layout-col">
                                     {{ record["date_last_modified"] }}
-                                    <a class="govuk-table--invisible-on-desktop govuk-!-static-margin-top-4"
+                                    <a class="govuk-table--invisible-on-desktop govuk-!-static-margin-top-4 word-break"
                                        href="{{ url_for('main.record', record_id=record['file_id']) }}">{{ record["file_name"] }}</a>
                                 </td>
-                                <td class="govuk-table__cell govuk-table--invisible-on-mobile">
+                                <td class="govuk-table__cell govuk-table--invisible-on-mobile word-break">
                                     <a href="{{ url_for('main.record', record_id=record['file_id']) }}">{{ record["file_name"] }}</a>
                                 </td>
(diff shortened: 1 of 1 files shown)


app/templates/main/browse-consignment.html
```diff
@@ -65,10 +65,10 @@
                             <tr class="govuk-table__row govuk-table__row-ref">
                                 <td class="govuk-table__cell govuk-table__cell--on-mobile--flex-layout-col">
                                     {{ record["date_last_modified"] }}
-                                    <a class="govuk-table--invisible-on-desktop govuk-!-static-margin-top-4"
+                                    <a class="govuk-table--invisible-on-desktop govuk-!-static-margin-top-4 word-break"
```
anthonyhashemi, line 68: interesting, does this suggest there is no standard way using gds to have a table that doesn't overflow into a component on the right of a grid?

ltrasca: @anthonyhashemi by default no unfortunately, tables overflow their containers when the content in the column is too large in width, so the solutions would be 1) break the column content anywhere (there may be some classes for this somewhere, we just have a custom class) or 2) let the container of the table scroll on the x axis, which would not be ideal

image

anthonyhashemi: that is unfortunate. I will try to source the github repo and raise an issue there. If we end up needing to reuse this extra class then we can pull it out into defined style itself but since it's just here for now I think this is fine 👌

anthonyhashemi (approved):

app/templates/main/browse-consignment.html
anthonyhashemi: found an issue for exactly this. Will share this PR as another example to hopefully get it fixed quicker!

https://github.com/alphagov/govuk-frontend/issues/5289

anthonyhashemi: Before merging this in @ltrasca can you verify that the potential issue mentioned in https://github.com/alphagov/govuk-frontend/issues/5289#issuecomment-2326650626 from adding word-break is or isn't happening here?

anthonyhashemi: We're not quite sure about the comment there but it seems to work for us here. I will ask him to clarify but think this PR is good to go for now

PR #863, opened 2025-04-14
AYR-1576 - Search radio buttons aria labels [DAC]

ltrasca:
## Changes in this PR
- Changed the aria label laguage of the radio buttons in the top search component to contain the same language as the main label above them (`Search for digital records`) @Terry-Price please review if the language is correct

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1576

## Screenshots of UI changes

N/A

### DIFF
--- a/app/templates/main/top-search.html
+++ b/app/templates/main/top-search.html
@@ -7,7 +7,7 @@
     {
         "value": "metadata",
         "label": "Record metadata only",
-        "aria_label": "Search digital records metadata only"
+        "aria_label": "Search digital records within record metadata only"
     },
     {
         "value": "record",
--- a/app/templates/main/top-search.html
+++ b/app/templates/main/top-search.html
@@ -2,17 +2,17 @@
     {
         "value": "everywhere",
         "label": "Search everywhere",
-        "aria_label": "Search for digital records across all records and metadata"
+        "aria_label": "Search everywhere in digital records"
     },
     {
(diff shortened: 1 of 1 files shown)


anthonyhashemi (approved):

anthonyhashemi (changes requested):
I believe the DAC recc in the ticket is to make the label text and aria label, match, so just use "label" for the aria labels too?

ltrasca:
> I believe the DAC recc in the ticket is to make the label text and aria label, match, so just use "label" for the aria labels too?

@anthonyhashemi They do mention it should contain it but making it all the same would mean losing context of what each radio button does, would basically all read out the same to screen readers

anthonyhashemi:
> > I believe the DAC recc in the ticket is to make the label text and aria label, match, so just use "label" for the aria labels too?
> 
> @anthonyhashemi They do mention it should contain it but making it all the same would mean losing context of what each radio button does, would basically all read out the same to screen readers

Ah yeah very fair, and they do say it should contain but not necessarily match, but I don't think that's the case still?

anthonyhashemi (approved):

anthonyhashemi (changes requested):
See latest comment

anthonyhashemi (approved):
Looks good, now it fits the requirement: https://www.w3.org/WAI/WCAG22/quickref/#label-in-name

PR #867, opened 2025-04-15
add docker compose test for data management

colinbowen:
## Changes in this PR
- Add docker compose test file for running indexer tests
- Setup PostgreSQL database inside Docker for integration tests
- Update test data with file metadata and document content
- Fix metadata field mapping in document objects

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1426

## Screenshots of UI changes
N/A

### After


data_management/opensearch_indexer/tests/conftest.py
```diff
@@ -88,11 +88,41 @@ def temp_db():
...
 def database(request):
-    # Launch new PostgreSQL server
-    postgresql = PostgresqlFactory(cache_initialized_db=True)()
-    yield postgresql
-
-    # PostgreSQL server is terminated here
-    @request.addfinalizer
-    def drop_database():
-        postgresql.stop()
+
```
anthonyhashemi, line 91: why did this have to change?

data_management/opensearch_indexer/tests/test_text_extraction.py
```diff
@@ -26,10 +26,15 @@ class TestExtractText:
...
+            (
+                "multiline.doc",
+                "doc",
+                "\nExpected content\nSecond line\nThird line\n",
+            ),
             (
                 "multiline.pdf",
                 "pdf",
-                "This is line 1\nThis is line 2\nThis is line 3\nThis is line 4, the ﬁnal line.\n\n\x0c",
+                "This is line 1\nThis is line 2\nThis is line 3\nThis is line 4, the final line.\n\n\x0c",
```
anthonyhashemi, line 37: how come this changed? Previously the ocr was picking it up as `ﬁ` not `fi`

docker-compose.test.yml
```diff
@@ -0,0 +1,40 @@
+services:
+  integration-runner:
```
anthonyhashemi: currently this docker-compose file is just running the tests from `data_management/opensearch_indexer` so maybe we move it/ rename it accordingly, and likewise rename this service name from `integration-runner`  to `opensearch_indexer_tests` or similar

docker-compose.test.yml
```diff
@@ -0,0 +1,39 @@
...
+      - .:/app
+    working_dir: /app
+    entrypoint: ""
+    command:
+    - sh
+    - -c
+    - |
+      set -e
+
+      echo 'Installing PostgreSQL...'
```
anthonyhashemi: Not sure why I didn't suggest before, but instead of installing postres here, you can just run it as another service. something along the lines of:

```
version: "3.8"
services:
  opensearch_indexer-test-runner:
    build:
      context: ./data_management/opensearch_indexer
      dockerfile: Dockerfile
    volumes:
      - .:/app
    working_dir: /app
    entrypoint: ""
    command: >
      sh -c "
      pip install -r data_management/opensearch_indexer/requirements-dev.txt &&
      pytest --cov=data_management/opensearch_indexer/opensearch_indexer --cov-report term-missing -vvs data_management/opensearch_indexer/tests
      "

  postgres:
    image: postgres:15
    environment:
      POSTGRES_USER: testuser
      POSTGRES_PASSWORD: testPass123
      POSTGRES_DB: testdb
    ports:
      - "5432:5432"
    volumes:
      - pgdata:/var/lib/postgresql/data

volumes:
  pgdata: ....
```

anthonyhashemi: Also a cleaner alternative to the overriding of entrypoint and cmd could be to use multistage docker builds:

e.g.

https://docs.docker.com/guides/java/run-tests/

data_management/opensearch_indexer/tests/conftest.py
colinbowen: Previously it was using a mock postgres instance inside python and this is the updated connection for using the new docker postgres one. There was an error loading the mock postgres db inside docker:

Screenshot 2025-04-15 at 14 46 44

anthonyhashemi: My understanding is that https://github.com/tk0miya/testing.postgresql creates a real postgresql instance, not a mock, dedicated to running the tests for when the tests start and then destroys it after, so that it is isolated from any existing databases in the postgres instance we point it to.

I see here you have your test db that you create currently in the cmd override in the compose file and nothing else is interacting with it.

Keeping testing.postgresql by pointing it to the postgres instance running in the other container if possible would reduce the coupling of the test code with the compose override code but probably not much in it. If we are to change how we interact with postgres in tests though it would be good to do so also in the webapp tests where we do the same thing.

Probably worth jumping on a call to talk through the options together

colinbowen: As discussed on call, reverted back to original postgres setup using [testing.postgresql](https://github.com/tk0miya/testing.postgresql)

data_management/opensearch_indexer/tests/test_text_extraction.py
colinbowen: Discussed on call

.pre-commit-config.yaml
```diff
@@ -14,52 +14,53 @@ repos:
     -   id: check-merge-conflict
 
 - repo: https://github.com/Riverside-Healthcare/djLint
```
anthonyhashemi, line 16: interesting, why havent these been updated by dependabot, we may need to configure that...

.pre-commit-config.yaml
```diff
@@ -14,52 +14,53 @@ repos:
...
 
 -   repo: https://github.com/PyCQA/bandit
-    rev: 1.7.7
+    rev: 1.8.3
     hooks:
     -   id: bandit
         args: ["-c", "pyproject.toml", "--exclude", "node_modules/*", "-r", "."]
 
--   repo: local
+-   repo: https://github.com/python-poetry/poetry
```
anthonyhashemi: what's this change? No more safety? This just generates a requirements.txt file but does nothing with itfro what I can tell?

data_management/opensearch_indexer/Dockerfile
```diff
@@ -5,7 +5,9 @@ RUN apt-get update && apt-get install -y \
     libxml2-dev libxslt1-dev \
     antiword unrtf poppler-utils tesseract-ocr \
     flac ffmpeg lame libmad0 libsox-fmt-mp3 sox \
-    libjpeg-dev swig
+    libjpeg-dev swig postgresql \
+    postgresql-client \
```
anthonyhashemi: Preferably we would not install these postgres things at this point since the webapp itself doesn't need it. When this is run in labda for exaple these would be unnecessary. 

Haven't read the whole updated PR yet, but I think a separate container running postgres in the docker-copose file would be best. If we were to do that, we can't use testing.postgresqls init method which runs within this container itself anymore, but that's fine, it no longer serves our needs. The stuff it does in init, i.e creating the table can be done in the compose file

data_management/opensearch_indexer/tests/test_bulk_index_files_in_opensearch.py
```diff
@@ -508,14 +508,20 @@ def test_fetch_files_in_consignment_invalid_column_with_logging(
     # Create the engine and session
     engine = create_engine(database_url)
 
-    # Minimal setup, omitting necessary schema parts
+    with engine.connect() as connection:
+        connection.execute(text('DROP TABLE IF EXISTS "FileMetadata" CASCADE'))
```
anthonyhashemi: what's the motivation with these additions?

docker-compose.test.yml
```diff
@@ -0,0 +1,52 @@
...
+        chmod -R u+w /app || echo "Chmod failed; check volume permissions"
+
+        echo 'Running tests...'
+        su testuser -c "pytest \
+          --cov=data_management/opensearch_indexer/opensearch_indexer \
+          --cov-report=xml:/test_results/coverage.xml \
+          --cov-report=term-missing \
+          -vvs data_management/opensearch_indexer/tests"
+
+  postgres:
```
anthonyhashemi, line 28: oh you are using a separate postgres container.... then is the stuff I commented on before additions you forgot to remove fro an earlier iteration?

docker-compose.test.yml
```diff
@@ -0,0 +1,52 @@
...
+    image: postgres:15
+    environment:
+      POSTGRES_USER: testuser
+      POSTGRES_PASSWORD: testPass123 # pragma: allowlist secret
+      POSTGRES_DB: testdb
+    ports:
+      - "55432:5432"
```
anthonyhashemi: I think this should be 
```suggestion
      - "5432:5432"
```

right?

data_management/opensearch_indexer/tests/test_bulk_index_files_in_opensearch.py
colinbowen: There was an issue with these tests creating the database schema before. It could be using a different created db, will check

data_management/opensearch_indexer/Dockerfile
colinbowen: Yes will remove installing postgres here as it isn't needed.

.pre-commit-config.yaml
colinbowen: Reverted this, left over from a prev commit to check if wiz picked up the full reqs.txt generated from poetry

.pre-commit-config.yaml
colinbowen: https://national-archives.atlassian.net/browse/AYR-1623

Created this.

anthonyhashemi: cheers

data_management/opensearch_indexer/tests/test_bulk_index_files_in_opensearch.py
anthonyhashemi: ok, I'm hoping it shouldn't be needed now with your change to using a separate postgres container

anthonyhashemi (dismissed):
Nice one, this was really involved, but should be a big boost in trust and consistency in the system

docker-compose.test.yml
```diff
@@ -0,0 +1,52 @@
...
+    working_dir: /app
+    entrypoint: ""
+    command:
+    - sh
+    - -c
+    - |
+        set -e
+
+        apt-get update
+        apt-get install -y postgresql postgresql-client postgresql-contrib
```
anthonyhashemi: this line can go now right?

docker-compose.test.yml
```diff
@@ -0,0 +1,52 @@
...
+        echo 'Installing Python test dependencies...'
+        pip install -r data_management/opensearch_indexer/requirements-dev.txt
+
+        echo 'Creating testuser...'
+        if ! id -u testuser > /dev/null 2>&1; then
+          echo "Creating system user testuser..."
+          useradd testuser
+        fi
+
+        echo 'Granting write permissions to testuser...'
```
anthonyhashemi: this stuff too

colinbowen:
PR contains updated e2e tests

data_management/opensearch_indexer/Dockerfile
```diff
@@ -5,7 +5,8 @@ RUN apt-get update && apt-get install -y \
     libxml2-dev libxslt1-dev \
     antiword unrtf poppler-utils tesseract-ocr \
     flac ffmpeg lame libmad0 libsox-fmt-mp3 sox \
-    libjpeg-dev swig
+    libjpeg-dev swig \
+    postgresql-contrib
```
anthonyhashemi: postgresql-contrib is not needed

data_management/opensearch_indexer/tests/conftest.py
```diff
@@ -88,11 +87,39 @@ def temp_db():
...
+            return self._url
+
+        def __call__(self):
+            return self._engine
+
+    db = DatabaseURL(database_url, engine)
+
+    if (
+        not request.function.__name__
+        == "test_fetch_files_in_consignment_invalid_column_with_logging"
```
anthonyhashemi: can we remove this condition?

anthonyhashemi (approved):
Great work man

PR #897, opened 2025-05-09
AYR-1569 - Reflow issue [DAC]

ltrasca:
## Changes in this PR
- Changed styles so the accordion expansion section of the search page is responsive

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1569

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_search-transferring-body.scss
+++ b/app/static/src/scss/includes/_search-transferring-body.scss
@@ -411,8 +411,8 @@ a.govuk-link {
   @media screen and (width <= $breakpoint-medium) {
     width: 100%;
     flex-direction: column;
-    gap: 16px;
-    margin-bottom: 18px;
+    gap: 0.5rem;
+    margin-bottom: 0.5rem;
   }
 
   .browse__sort-container {
--- a/app/static/src/scss/includes/_search-transferring-body.scss
+++ b/app/static/src/scss/includes/_search-transferring-body.scss
@@ -408,20 +408,39 @@ a.govuk-link {
   margin-top: 2rem;
   padding: 0;
 
+  @media screen and (width <= $breakpoint-medium) {
+    width: 100%;
+    flex-direction: column;
+    gap: 16px;
+    margin-bottom: 18px;
+  }
+
   .browse__sort-container {
     margin-top: 0;
   }
 
   .govuk-select {
     max-width: 240px;
+
+    @media screen and (width <= $breakpoint-medium) {
+      max-width: 100%;
+    }
   }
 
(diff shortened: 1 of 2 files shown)


anthonyhashemi (approved):

anthonyhashemi (changes requested):
Sorry I meant to approve the other pr

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -408,20 +408,39 @@ a.govuk-link {
   margin-top: 2rem;
   padding: 0;
 
+  @media screen and (width <= $breakpoint-medium) {
+    width: 100%;
+    flex-direction: column;
+    gap: 16px;
```
anthonyhashemi: There are instances of hardcoded values like gap: 16px or gap: 0.5rem. Could we use relative units like em, rem or % or some central units we define?

app/static/src/scss/includes/_search-transferring-body.scss
```diff
@@ -408,20 +408,39 @@ a.govuk-link {
...
   &__checkboxes {
     display: flex;
     justify-content: flex-end;
     align-items: center;
-    gap: 16px;
+    gap: 0.5rem;
     height: 40px;
+    flex-shrink: 0;
+
+    @media screen and (width <= $breakpoint-medium) {
```
anthonyhashemi, line 437: Out of curiosity, would it potentially be possible to utilise flex-wrap: wrap and width:100% for dynamic responsiveness without needing media screen sections?

ltrasca: Its definitely possible but will produce layout changes in-between our specific breakpoints, such as this:

image

app/static/src/scss/includes/_search-transferring-body.scss
ltrasca: Changed them to rem, thanks for pointing this out. Discussed having units that are defined in a standardized manner but this hasn't materialized yet – in my mind the most sensible solution is just using those already defined by GDS, will create a ticket to deal with this soon.

anthonyhashemi: amazing, thanks Lucian

app/static/src/scss/includes/_search-transferring-body.scss
anthonyhashemi: I see, would be cool to spend a session going through this idea together. Fine as is for now though

anthonyhashemi (approved):
Looks good 👌

PR #899, opened 2025-05-09
AYR-1578 - Letter spacing issue [DAC]

ltrasca:
## Changes in this PR
Adjusted styles on the search transferring body "Search within results" section to:
- Wrap the button and anchor elements when the letter spacing is too large
- Unpin the width of the apply button so the text affects width and doesn't break line and overflow the button element itself
- Increase the font size of the button so its consistent with all the other buttons we use within the app
## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1578
### Before
image

With 2px letter spacing

image

### After

image

With 2px letter spacing

image


anthonyhashemi (approved):
Looks much better now when text spacing mode is on

PR #902, opened 2025-05-13
add importer file

colinbowen:
## Changes in this PR
- Create mds_test_file_importer to populate Minio & AYR test databases
- Allow for PNG, JPG, PDF, TIFF files
- Create cleanup_test_data to clear populated databases
- Index added files to documents in opensearch and remove documents index when running clear data script

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1224

### After


local_services/mds_data_generator/cleanup_test_data.py
```diff
@@ -0,0 +1,149 @@
...
+        print(f"MinIO cleanup completed - deleted {deleted_count} objects")
+        return True
+    except Exception as e:
+        print(f"Error cleaning up MinIO: {e}")
+        return False
+
+
+def cleanup_test_data():
+    """Cleanup of all test data from database, example files, and MinIO."""
+    results = {
```
anthonyhashemi: do we need this dict? we return it in this function but don't access the value in the calling function

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,256 @@
...
+def main():
+    """
+    Generate and process test files for AYR testing using Factory Boy.
+
+    Example usage:
+        python mds_test_file_importer.py --num-pdf 2 --num-png 1
+    """
+    parser = argparse.ArgumentParser(
+        description="Generate and process test files for AYR testing"
+    )
```
anthonyhashemi, line 279: maybe it's worth setting the default values to 1 of each?

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,256 @@
...
+
+    return created_files
+
+
+def process_files(file_paths):
+    """Process files by uploading to S3 and adding metadata to database."""
+    app = create_app(EnvConfig)
+    with app.app_context():
+        try:
+            body = BodyFactory.create_batch(
```
anthonyhashemi: can just use create rather than create_batch with 1

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,256 @@
...
+                1,
+                Name="Test Transferring Body",
+                Description="Test Transferring Body Description",
+            )[0]
+            print(f"Created or using Body: {body.Name} (ID: {body.BodyId})")
+
+            series = SeriesFactory.create_batch(
```
anthonyhashemi: same here

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,256 @@
...
+
+    with open(file_path, "rb") as file_data:
+        s3.put_object(
+            Bucket=bucket, Key=s3_key, Body=file_data, ACL="public-read"
+        )
+
+
+def create_test_files(file_type_counts):
+    """Create test files by copying example files of specified types."""
+    example_folder = os.path.join(
```
anthonyhashemi: I'm wondering, would it be worth using the image files we already have in the repo than a version of this "redacted" doc for all the types?

On second thoughts, I guess it does not matter unless we want to be able to test text extraction/content search. For that, the images we have there are not that interesting anyway, and we haven't done the opensearch indexing in this PR so no point just yet.

For the ticket to be compete, I do think we need to cover opensearch indexing though (could be done in a separate PR)

local_services/mds_data_generator/cleanup_test_data.py
```diff
@@ -0,0 +1,149 @@
...
+from db_utils import get_db_connection
+from dotenv import load_dotenv
+
+load_dotenv()
+
+
+def clean_database():
+    """Clean up all test data from the database."""
+    conn = get_db_connection()
+
```
anthonyhashemi, line 47: Instead of this, I wonder if the following utilising the SQLAlchemy ORM is sufficient

```
from app.main.db.models import db

db.drop_all()
db.create_all()
```

local_services/mds_data_generator/db_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        port=os.getenv("DB_PORT"),
+        dbname=os.getenv("DB_NAME"),
+        user=os.getenv("DB_USER"),
+        password=os.getenv("DB_PASSWORD"),
+        sslmode="verify-full",
+        sslrootcert=os.getenv("DB_SSL_ROOT_CERTIFICATE"),
+    )
+
+
+def insert_metadata(conn, document_id, metadata):
```
anthonyhashemi: uneeded now you are using factoryboy

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,256 @@
...
+                ContactName="Test User",
+                ContactEmail="test@example.com",
+                TransferStartDatetime=datetime.utcnow(),
+                CreatedDatetime=datetime.utcnow(),
+            )
+            print(
+                f"Created new consignment: {consignment_ref} (ID: {consignment.ConsignmentId})"
+            )
+
+            for i, (file_path, file_ext) in enumerate(file_paths):
```
anthonyhashemi: Could we instead create random file_ids here which we share between filefactory and upload_to_s3 and referencing the original example file to upload by file type?

e.g. 

```
example_files = "local_services/mds_data_generator/example_files"
example_fille_paths = {"pdf": example_files/"redacted.pdf", .....}
file_path = example_fille_paths[file_ext]
```

It would save you needing the creation and deletion functions for the filesystem and keep the dev filesystems clean

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,349 @@
...
+    file_type_counts = {
+        "pdf": args.num_pdf,
+        "png": args.num_png,
+        "jpg": args.num_jpg,
+        "tiff": args.num_tiff,
+    }
+
+    file_type_counts = {k: v for k, v in file_type_counts.items() if v > 0}
+
+    if not file_type_counts:
```
anthonyhashemi: Can't you just set default in `parser.add_argument(` rather than overriding here?

local_services/mds_data_generator/cleanup_test_data.py
```diff
@@ -0,0 +1,129 @@
...
+        deleted_count = 0
+
+        if "Contents" in response:
+            for obj in response["Contents"]:
+                # Delete objects that are in TEST-* folders
+                if obj["Key"].startswith("TEST-"):
+                    s3.delete_object(Bucket=bucket, Key=obj["Key"])
+                    deleted_count += 1
+
+        print(f"MinIO cleanup completed - deleted {deleted_count} objects")
```
anthonyhashemi, line 115: you should be able to do just `bucket.objects.delete()`

colinbowen: Yes this worked if using an s3 resource vs a client

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,325 @@
...
+
+    file_type_counts = {
+        "pdf": args.num_pdf,
+        "png": args.num_png,
+        "jpg": args.num_jpg,
+        "tiff": args.num_tiff,
+        "txt": args.num_txt,
+    }
+
+    file_type_counts = {k: v for k, v in file_type_counts.items() if v > 0}
```
anthonyhashemi: [simplification/redundancy]: I don't think this filter is needed, since yu loop over over range of count, so if its 0 you dont create any filepath

local_services/mds_data_generator/mds_test_file_importer.py
```diff
@@ -0,0 +1,325 @@
...
+
+                file_id = file.FileId
+                consignment_ref = file.consignment.ConsignmentReference
+
+                try:
+                    s3_key = f"{consignment_ref}/{file_id}"
+                    response = s3.get_object(Bucket=bucket, Key=s3_key)
+                    file_stream = response["Body"].read()
+
+                    file_data = _fetch_file_data(file_id, database_url)
```
anthonyhashemi: you could replace 

```
    file_data = _fetch_file_data(file_id, database_url)
    file_data_with_text_content = add_text_content(file_data, file_stream)
    _index_in_opensearch(
        file_id,
        file_data_with_text_content,
        open_search_host_url,
        open_search_http_auth,
        open_search_ca_certs,
    )
```
with
 `index_file_content_and_metadata_in_opensearch`

anthonyhashemi (approved):
Great work, this is going to be so helpful for local development!

PR #911, opened 2025-05-21
failed files fallback function added

rahmedzaizi:
### After


.pre-commit-config.yaml
```diff
@@ -48,7 +48,7 @@ repos:
     rev: 1.8.3
     hooks:
     -   id: bandit
-        args: ["-c", "pyproject.toml", "--exclude", "node_modules/*", "-r", "."]
+        args: ["-c", "pyproject.toml", "--exclude", "node_modules/*", "--skip", "B404,B603,B607",  "-r", "."]
```
anthonyhashemi: can we whitelist in line in code rather than skip

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -76,5 +80,56 @@ def extract_text(file_stream: bytes, file_extension: str) -> str:
...
+            if file_extension not in CONVERT_MAP:
+                raise e
+            try:
+                logger.info("Attempting to convert via LibreOffice...")
+                converted_path = convert_failed_files(file_path)
+                logger.info(f"Converted to: {converted_path}")
+                text = textract.process(converted_path)
+                return text.decode("utf-8")
+            except Exception as convert_err:
+                logger.error(f"LibreOffice conversion failed: {convert_err}")
```
anthonyhashemi: this log error is not necessarily true, it could be that conversion faield after converting but text extraction still failed. Can move the try except around the `convert_failed_files` line specifically

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -76,5 +80,56 @@ def extract_text(file_stream: bytes, file_extension: str) -> str:
...
+                raise convert_err
+
+
+def convert_failed_files(input_path: str) -> str:
```
anthonyhashemi: thin it would be simpler to test / simplify concerns of this function by passing in the output_file_type as an argument, rather than needing to pass in a constant?

anthonyhashemi:
```suggestion
def convert_file(input_path: str, output_file_type: str) -> str:
```

The function doesn't really care about the fact that text extraction failed or not

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -76,5 +80,56 @@ def extract_text(file_stream: bytes, file_extension: str) -> str:
...
+            context = textract.process(file_path)
+            return context.decode("utf-8")
+
+        except Exception as e:
+            logger.warning(f"Textract failed on {file_path}: {e}")
+
+            if file_extension not in CONVERT_MAP:
+                raise e
+            try:
+                logger.info("Attempting to convert via LibreOffice...")
```
anthonyhashemi:
```suggestion
                logger.info("Attempting to convert to {output_file_type} before trying textract again...")
```

to describe why we are doing the conversioon

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -8,6 +10,8 @@
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 
+CONVERT_MAP = {"xls": "xlsx", "doc": "docx"}
```
anthonyhashemi: maybe worth making this constant name a little more descriptive
```suggestion
TEXTRACT_FILE_FORMAT_FALLBACK_CONVERSION_MAP = {"xls": "xlsx", "doc": "docx"}
```

Also if we end up just using it in one place, does it need to be a constant defined outside of the function?

`textract_file_format_fallback_conversion_map = {"xls": "xlsx", "doc": "docx"}` inside the function could be sufficient for now and saves us having to move out of that context to know what the types are

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -76,5 +80,56 @@ def extract_text(file_stream: bytes, file_extension: str) -> str:
     ) as temp:
         temp.write(file_stream)
         temp.flush()
-        context = textract.process(temp.name)
-    return context.decode("utf-8")
+        file_path = temp.name
+
+        try:
```
anthonyhashemi, line 85: adding a couple of tests for these scenarios inside of
`data_management/opensearch_indexer/tests/test_text_extraction.py` would be good

anthonyhashemi (approved):

PR #922, opened 2025-05-30
Ayr 1634/e2e ci

colinbowen:
### After


.github/workflows/unit_tests.yml
```diff
@@ -68,3 +68,187 @@ jobs:
...
+      #     echo "Waiting for OpenSearch"
+      #     timeout 120s bash -c 'until curl -s http://localhost:9200/_cluster/health | grep -q "yellow\|green"; do sleep 5; done'
+
+      #     echo "Waiting for MinIO"
+      #     timeout 60s bash -c 'until curl -s http://localhost:9000/minio/health/live; do sleep 2; done'
+
+      - name: Install dependencies for textract
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
```
anthonyhashemi, line 130: these are done inside the docker container so arent needed here?

.github/workflows/unit_tests.yml
```diff
@@ -68,3 +68,187 @@ jobs:
...
+            python3-dev libxml2-dev libxslt1-dev antiword unrtf poppler-utils \
+            ghostscript tesseract-ocr flac ffmpeg lame libmad0 libsox-fmt-mp3 \
+            sox libjpeg-dev swig zlib1g-dev
+
+          poetry run python -m pip install --upgrade "pip<24.1"
+          poetry run pip install textract
+
+      - name: Generate SSL certs inside Docker
```
anthonyhashemi, line 227: these extra steps that are run outside the container could maybe be put into a script so that local devs can set up the test env locally without needing to copy paste?

local_services/certs-builder/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine:3.20
```
anthonyhashemi, line 1: why is this a dockerfile?

local_services/certs-builder/webapp_postgres_certs/generate_webapp_postgres_certs.sh
```diff
@@ -0,0 +1,13 @@
+# Create a private key
```
anthonyhashemi, line 7: these already exist at `local_services/webapp_postgres_certs/generate_webapp_postgres_certs.sh` do we need it here too? or have we planning to delete from original location?

PR #926, opened 2025-06-04
Added new main.py for ecs-indexer and updated tests

rahmedzaizi:
Changed Entrypoint logic to migrate from lambda to ECS
Added test file for new logic


data_management/opensearch_indexer/opensearch_indexer/index_consignment/lambda_function.py
```diff
@@ -10,27 +10,7 @@
 
 
 def lambda_handler(event: Dict[str, Any], context: Any) -> None:
```
colinbowen: A small docstring for what this lambda does would be useful. Could keep part of the previous one which was previously there?

"""
AWS Lambda handler function to trigger the indexing of consignment files into OpenSearch.	

This function is invoked by an AWS event containing details of a consignment. It retrieves	
the necessary parameters from the event and environment variables, then calls the indexing	
function to process and index the files into OpenSearch.
"""

rahmedzaizi: Don't need to change this file actually 
I will just revert the changes

colinbowen (dismissed):

data_management/opensearch_indexer/opensearch_indexer/index_consignment/main.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+import os
+
+from .bulk_index_consignment import bulk_index_consignment_from_aws
+
+logging.basicConfig(level=logging.INFO, format="%(levelname)s: %(message)s")
+logger = logging.getLogger()
+
+
+def main():
```
anthonyhashemi: could we rename the file and function to something more descriptive (and thus all the references to main in the tests)

`main`-> `index_consignment` or something like that?

data_management/opensearch_indexer/Dockerfile
```diff
@@ -1,24 +1,25 @@
...
-COPY opensearch_indexer/ ${LAMBDA_TASK_ROOT}/opensearch_indexer
+COPY opensearch_indexer/ ./opensearch_indexer
 
-# Set runtime interface client as default command for the container runtime
-ENTRYPOINT [ "/usr/local/bin/python", "-m", "awslambdaric" ]
-
-# Pass the name of the function handler as an argument to the runtime
-CMD [ "opensearch_indexer.index_consignment.lambda_function.lambda_handler" ]
+# Run main.py
+CMD ["python", "-m", "opensearch_indexer.index_consignment.main"]
```
anthonyhashemi, line 25: you were saying how we would have both the lambda version and the ecs version deployed for a bit before deleting the lambda. Even though we have changeds the dockerfile, that will be fine because when you deploy,  you jus won't deploy a new lambda and so it will point to the old docker image? Probably covered in the other PR, but are you creating a new ECR repo for this new image, probably makes sense to, right?

rahmedzaizi: Yes 
Creating a new Repo for ecs image so lambda will keep on using the image that is currently deployed

anthonyhashemi (approved):

PR #938, opened 2025-06-13
use end date

colinbowen:
## Changes in this PR
- Use end_date if available for Date of Record on /browse/consignment & /record

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1594

### After


app/templates/main/record.html
```diff
@@ -35,7 +35,7 @@
             "Description": record['description'],
             "Citeable reference": record['citeable_reference'],
             "Status": record['closure_type'],
-            "Date of record": record['date_last_modified'],
+            "Date of record": record['end_date'] if record['end_date'] != None else record['date_last_modified'],
```
anthonyhashemi: could we make `Date of record` a field in record itself  e.g. `record[date_of_record]` so we don't have to repeat this logic in these 3 templates? maybe even with coalesce like in the sort query although maybe theres something more efficient, if we dont need these fields separately?

app/main/db/queries.py
```diff
@@ -577,9 +637,15 @@ def _get_file_metadata_query(file_id: uuid.UUID):
                 current_app.config["DEFAULT_DATE_FORMAT"],
             ).label("opening_date"),
             func.to_char(
-                sub_query.c.date_last_modified,
+                func.coalesce(
+                    sub_query.c.end_date, sub_query.c.date_last_modified
+                ),
+                current_app.config["DEFAULT_DATE_FORMAT"],
+            ).label("date_of_record"),
```
anthonyhashemi, line 644: nice!

app/tests/test_queries.py
```diff
@@ -280,9 +287,18 @@ def test_get_file_metadata_return_results_with_no_metadata(
                 ].Value,
                 "closure_period": record_files[3]["closure_period"].Value,
                 "opening_date": record_files[3]["opening_date"].Value,
-                "date_last_modified": record_files[3][
-                    "date_last_modified"
-                ].Value,
+                "date_of_record": (
+                    None
+                    if record_files[3]["end_date"].Value is None
```
anthonyhashemi: this test looks a bit complicated. Ideally tests shouldn't need conditional logic. I feel like splitting these conditions out into distinct test cases would be clearer, so each test is testing one scenario; easier to debug and nicer as a piece of documentation of the function.

The test isn't paramterized, so really the condition is redundant anyway, right, as its always true or false?

app/main/db/queries.py
```diff
@@ -352,9 +388,17 @@ def build_browse_consignment_query(
             query = query.filter(date_filter)
 
     if sorting_orders:
-        query = _build_sorting_orders(query, sub_query, sorting_orders)
+        if "date_last_modified" in sorting_orders:
```
anthonyhashemi, line 391: not sure about this. The ticket didn't initially consider the sorting. I feel like if the sort dropdown says Date last modified and so the param we send here is `date_last_modified`, then  we shouldn't be getting end_date. Ideally we would change it to be consistent with what is shown in the table, Date of record, and then this all makes sense. Worth speaking to Terry about the dropdown.

app/main/db/queries.py
```diff
@@ -384,7 +428,14 @@ def _build_browse_filters(query, sub_query, filters):
 
 def _build_sorting_orders(query, sub_query, sorting_orders):
     for field, order in sorting_orders.items():
-        column = getattr(sub_query.c, field, None)
+        if field == "end_date":
```
anthonyhashemi: not sure about this. The ticket didn't initially consider the sorting. I feel like if the sort dropdown says `End date` modified and so the param we send here is `end_date`, then  we shouldn't be getting date_last_modified when it doesnt exist. Ideally we would change it to be consistent with what is shown in the table, Date of record, and then this all makes sense. Worth speaking to Terry about the dropdown.

Its possible I am just completely misunderstanding this

app/tests/test_build_browse_consignment_query.py
```diff
@@ -66,3 +70,32 @@ def test_build_browse_consignment_query_no_results(
         )
         results = query.all()
         assert results == []
+
+    def test_build_browse_consignment_query_filters_by_end_date_when_available(
+        self, client: FlaskClient, mock_standard_user, browse_consignment_files
+    ):
+        """
+        Given a consignment with files that have both end_date and date_last_modified
+        When build_browse_consignment_query is called with date_last_modified filtering
```
anthonyhashemi, line 79: isn't it more appropriate for the filed to be Date of record which then uses either date_last_modified or end_date behind the scenes?

app/main/db/queries.py
anthonyhashemi: I think this is what I am talking about in the tests, this could be some form "date of record" to match the UI.

UPDATE: I see the mapping of 

 "Date of record (newest)": "end_date-desc",
    "Date of record (oldest)": "end_date-asc",
    
 maybe its worth catching up because I think I am confusing myself

colinbowen: I think changing this 

Date of record (newest)": "end_date-desc",
"Date of record (oldest)": "end_date-asc",

to 

"Date of record (newest)": "date_of_record-desc",
    "Date of record (oldest)": "date_of_record-asc",
    
   is what we are looking for, with the checks for end_date -> date_last_modified done via the queries

anthonyhashemi: yeah, that makes sense to me, nice one 👌

app/templates/main/browse-consignment.html
```diff
@@ -65,7 +65,7 @@
                         {% for record in results %}
                             <tr class="govuk-table__row govuk-table__row-ref">
                                 <td class="govuk-table__cell govuk-table__cell--word-break-all govuk-table__cell--on-mobile--flex-layout-col">
-                                    {{ record["date_last_modified"] }}
+                                    {{ record['end_date'] if record['end_date'] != None else record["date_last_modified"] }}
```
anthonyhashemi: you can use record["date_of_record"] here as well right?

app/tests/test_queries.py
colinbowen: Added some test cases for each scenario we should experience for this

app/templates/main/browse-consignment.html
colinbowen: Yep, added this change

app/main/db/queries.py
colinbowen: Changed to use date_of_record as the query param with all end_date, date_last_modified fallback via queries

anthonyhashemi (dismissed):

anthonyhashemi (dismissed):

anthonyhashemi (approved):

PR #969, opened 2025-06-17
Fix/e2e tests

colinbowen:
## Changes in this PR
- Update e2e tests

### After


anthonyhashemi (commented):
do we know when this changed/ what caused it

anthonyhashemi (approved):

colinbowen:
> do we know when this changed/ what caused it

Checking this, looks like it was a fe change that was merged in and we just didn't update the tests.

PR #970, opened 2025-06-17
AYR-1535 - Move page to view top of table when pagination is used

ltrasca:
## Changes in this PR
- Added a fragment identifier of `#tbl_result` to the hrefs of the pagination anchor links so if an element with the same id attribute exists the page will show it on load
- Upgraded setuptools to `^78.1.1` because of a vulnerability warning

## JIRA ticket
[national-archives.atlassian.net/browse/AYR-1535](https://national-archives.atlassian.net/browse/AYR-1535)
## Screenshots of UI changes

N/A


anthonyhashemi (approved):
Nice 👌

PR #979, opened 2025-06-18
add flask assets to style src csp

colinbowen:
## Changes in this PR
- Add cloudfront assets url to CSP

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1660

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/configs/base_config.py
+++ b/configs/base_config.py
@@ -186,7 +186,10 @@ class BaseConfig(object):
 
     @property
     def CSP_STYLE_SRC(self):
-        return [SELF]
+        return [
+            SELF,
+            self.FLASKS3_CDN_DOMAIN,
+        ]
 
     @property
     def CSP_STYLE_SRC_ELEM(self):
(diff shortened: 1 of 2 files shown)


configs/base_config.py
```diff
@@ -186,7 +186,10 @@ def CSP_SCRIPT_SRC_ELEM(self):
 
     @property
     def CSP_STYLE_SRC(self):
-        return [SELF]
+        return [
+            SELF,
+            self.FLASKS3_CDN_DOMAIN,
```
anthonyhashemi, line 191: what issue was this causing?

colinbowen: Screenshot 2025-06-18 at 12 03 28

The CSP policy in safari is blocking the styles from being loaded which is different compared to chrome which seems to be allowing it through csp_style_src_elem. I think safari's implementation of csp policies is more strict

See:
https://auto-bugs.webkit.org/show_bug.cgi?id=276931

anthonyhashemi (approved):

PR #1008, opened 2025-06-25
AYR-1538 - Adjustments to table top margin

ltrasca:
## Changes in this PR
- On desktop increased the gap between the sort panel and the table/filter to 1.5rem, on mobile this is 0.5rem
- On mobile, increased the gap between the filters and table to 1rem (on mobile the table sits under filters)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1538

### Before
image

### After
image

### DIFF
--- a/app/static/src/scss/includes/_browse-consignment.scss
+++ b/app/static/src/scss/includes/_browse-consignment.scss
@@ -254,5 +254,9 @@
 }
 
 .sort-container__form {
-  margin-bottom: 24px;
+  margin-bottom: 1.5rem;
+
+  @include on-mobile {
+    margin-bottom: 0.5rem;
+  }
 }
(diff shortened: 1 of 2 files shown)


app/static/src/scss/includes/_browse-consignment.scss
```diff
@@ -254,5 +254,9 @@
 }
 
 .sort-container__form {
```
anthonyhashemi, line 256: is it worth us using flex on a parent container rather than margins on children?

anthonyhashemi (approved):

app/static/src/scss/includes/_browse-consignment.scss
ltrasca: @anthonyhashemi it would, but would need a bit of refactoring, though should be 100% done at some point

PR #1026, opened 2025-07-01
Ayr 1667 - slack alerts for indexer skipped failed files

rahmedzaizi:
### After


data_management/opensearch_indexer/tests/test_text_extraction.py
```diff
@@ -91,6 +91,13 @@ def test_extract_text(self, file_name, file_type, expected_output):
         assert extract_text(file_stream, file_type) == expected_output
 
 
+# Mock ENVIRONMENT for slack alerts
+@pytest.fixture(autouse=True)
```
anthonyhashemi, line 95: Why is this being patched here and  in the with statement in the tests?

Standard way to do this, which is probably clearer, is:
in each test use the monkeypatch fixture `def test_name(monkeypatch):` and do:
`monkeypatch.setenv('ENIVRONMENT', 'test-env')`

anthonyhashemi (commented):
I was thinking whether there'd be a good way to do this via cloudwatch to keep within our alarms structure but really this is about particular issues in the application/a process, as opposed to general infrastructure issues so I think it is fine to be different.

This approach means the lambdas are coupled to slack, an alternative could be to just write logs and create alarms based off of those, but I think this is fine for now as long as we document it in our alarms confluence page

data_management/opensearch_indexer/tests/test_text_extraction.py
rahmedzaizi: Other tests were failing with `'NoneType' object has no attribute 'upper'`  because it's within `add_text_content` function
So set it at the top and forgot to remove it from the test itself

rahmedzaizi:
> I was thinking whether there'd be a good way to do this via cloudwatch to keep within our alarms structure but really this is about particular issues in the application/a process, as opposed to general infrastructure issues so I think it is fine to be different.
> 
> This approach means the lambdas are coupled to slack, an alternative could be to just write logs and create alarms based off of those, but I think this is fine for now as long as we document it in our alarms confluence page

Thats was my initial plan, to create slack alerts from Logs, but then thought it's an issue for this particular process and also easier to implement 

I will document it

data_management/opensearch_indexer/tests/test_text_extraction.py
anthonyhashemi: ah ok

anthonyhashemi (dismissed):

anthonyhashemi (approved):

PR #1037, opened 2025-07-03
AYR-1577 & 1565 - Accordion accessibility fixes

ltrasca:
## Changes in this PR

- Replaced the `<details>` element that was previously used to manage the accordion with a button that contains several accessibility attributes, including:
	- An `aria-label` so we can dictate to screen readers what the purpose of the button is without needing to modify the content
	- `aria-controls`, with a list of rows that the button programatically controls (for screen readers in conjunction with aria-expanded), for more info see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-controls
	- `aria-expanded`, which denotes the status of the entire accordion

- Rows now also have `aria-expanded`, which works together with the `aria-controls` of the associated button to create the relationship that communicates to screen readers when rows are expanded or not. Rows also have the attribute `aria-hidden` used to entirely remove or add them from the accessibility tree.

- Added some simple JS logic to manage all of the above, including handling the case where all accordions need to be expanded via the checkbox
- By default, when there is no JS enabled, the buttons are disabled and the rows are expanded (see screenshot below)

The result should be:
- The control (now a button) has an accessible name given by the `aria-label`
- There is a clear relationship between the control and the elements that are being controlled (the rows), and the nature of the control is expansion (via `aria-expanded`)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1565
https://national-archives.atlassian.net/browse/AYR-1577

### Before

Normal case (with one manually opened accordion)
image

Normal case, expanded accordions checkbox checked
image

No JS enabled
image

### After

Normal case (with one manually opened accordion)
image

Normal case, expanded accordions checkbox checked
image

No JS enabled
image


anthonyhashemi:
Looks good. Can you update the visual regression tests to match these changes?

colinbowen (approved):

PR #1044, opened 2025-07-09
use FFIDMetadata for file_extension

rahmedzaizi:
## Changes in this PR
- Updated file metadata queries to include file_extension from the FFIDMetadata table instead of deriving from file_name
## JIRA ticket
[AYR-1658](https://national-archives.atlassian.net/browse/AYR-1658)

[AYR-1658]: https://national-archives.atlassian.net/browse/AYR-1658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/main/db/models.py
```diff
@@ -64,3 +70,24 @@ class FileMetadata(db.Model):
...
+    FileId = db.Column(
+        UUID(as_uuid=True), db.ForeignKey("File.FileId"), primary_key=True
+    )
+    Extension = db.Column(Text)
+    PUID = db.Column(Text)
+    FormatName = db.Column(Text)
+    ExtensionMismatch = db.Column(Text)
+    FFID_Software = db.Column("FFID-Software", Text)
+    FFID_SoftwareVersion = db.Column("FFID-SoftwareVersion", Text)
+    FFID_BinarySignatureFileVersion = db.Column(
```
anthonyhashemi, line 87: I was going to suggest that these names be like `FFIDBinarySignatureFileVersion` but there seems to be an inconsistency between the table name and these fields in the db anyway so fine to leave like this.

app/tests/test_record.py
```diff
@@ -242,10 +241,7 @@ def test_record_record_alert_banner_is_visible_when_unsupported_file_extension(
         Then the alert banner responsible with alerting the user should be visible
         """
 
-        file = FileFactory(
-            FileName="open_file_once_closed.foobar",
-            FileType="file",
-        )
+        file = record_files[0]["file_object"]
```
anthonyhashemi: I'd suggest continuing to use filefactory directly. We are trying to move away from file factory as it makes the test setup so obscure when reading and understanding a test

app/tests/test_record.py
```diff
@@ -274,10 +270,7 @@ def test_record_record_alert_banner_is_not_visible_when_supported_file_extension
         Then the alert banner responsible with alerting the user should NOT be visible
         """
 
-        file = FileFactory(
-            FileName="open_file_once_closed.pdf",
-            FileType="file",
-        )
+        file = record_files[1]["file_object"]
```
anthonyhashemi: same here

app/tests/test_routes.py
```diff
@@ -126,12 +125,7 @@ def test_route_generate_pdf_manifest(
         )
 
         mock_all_access_user(client)
-        file_id = "efb1b3b4-0b3b-4b3b-8b3b-0b3b4b3b4b3b"
-        file = FileFactory(
-            FileId=file_id,
-            FileType="File",
-            FileName="open_file_once_closed.pdf",
-        )
+        file = record_files[1]["file_object"]
```
anthonyhashemi: same here

data_management/opensearch_indexer/tests/test_consignment_main.py
```diff
@@ -138,6 +152,30 @@ def test_main_invokes_bulk_index_with_correct_file_data(monkeypatch, database):
                 PropertyName="Key6",
                 Value="Value6",
             ),
+            FFIDMetadata(
+                FileId=file_ids[0],
+                Extension="txt",
+                PUID="fmt/115",  # example, depends on your app logic
```
anthonyhashemi: what does the comment mean?

data_management/opensearch_indexer/tests/test_consignment_main.py
```diff
@@ -10,7 +10,21 @@
...
+    Consignment,
+    FFIDMetadata,
+    File,
+    FileMetadata,
+    Series,
+)
+
+
+# Mock ENVIRONMENT for slack alerts
+@pytest.fixture(autouse=True)
```
anthonyhashemi, line 24: was it breaking without this?

data_management/opensearch_indexer/tests/conftest.py
```diff
@@ -75,6 +75,27 @@ class FileMetadata(Base):
     file = relationship("File", foreign_keys="FileMetadata.FileId")
 
 
+class FFIDMetadata(Base):
```
anthonyhashemi, line 78: we know we have this tech debt of repeating here and in the app models file, and have a ticket to consolidate, just need to prioritise

app/tests/conftest.py
```diff
@@ -760,6 +760,10 @@ def record_files():
         FileReference="ABCDE",
         FilePath="data/content/test_folder/open_file.docx",
         CiteableReference="first_body/ABCDE",
+        ffid_metadata__Extension="docx",
```
rahmedzaizi, line 763: Not needed by tests as we are using Factories instead of this fixture but added for consistency

app/tests/factories.py
```diff
@@ -54,12 +79,15 @@ class Meta:
     FileId = factory.LazyFunction(uuid.uuid4)
     consignment = factory.SubFactory(ConsignmentFactory)
     FileReference = FuzzyText(length=10)
-    FileType = FuzzyText(length=10)
+    FileType = FuzzyChoice(["File"])
```
rahmedzaizi, line 82: We only have FileType=file for now
 This avoids generating random strings and removes the need to override FileType in every test.

data_management/opensearch_indexer/tests/test_consignment_main.py
rahmedzaizi: Yes 
I added to other tests when added slack_alerts and forgot to add it here
 but this was passing somehow 
I guess other changes may have triggered it

anthonyhashemi (approved):
Nice one

PR #1049, opened 2025-07-11
removed single file indexing code and tests

rahmedzaizi:
## Changes in this PR

Removed single file indexing code and tests


anthonyhashemi (approved):

anthonyhashemi:
note this also removes the function for indexing all files in db but we can up date the opensearch disaster recovery workflow to reflect this and the new ecs structure.

PR #1064, opened 2025-07-22
added fallback to using file name to extract file extension 

rahmedzaizi:
…Metadata not available

## Changes in this PR
added fallback to using file name to extract file extension when FFIDMetadata not available
## JIRA ticket
[AYR-1658](https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?jql=assignee%20%3D%2060ba6c41f8a90b006f65b5fb&selectedIssue=AYR-1658)

[AYR-1658]: https://national-archives.atlassian.net/browse/AYR-1658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -717,7 +717,11 @@ def record(record_id: uuid.UUID):
     validate_body_user_groups_or_404(file.consignment.series.body.Name)
 
     file_metadata = get_file_metadata(file.FileId)
-    file_extension = file.ffid_metadata.Extension.lower()
+    if file.ffid_metadata and file.ffid_metadata.Extension:
+        file_extension = file.ffid_metadata.Extension.lower()
+    else:
+        file_extension = file.FileName.split(".")[-1].lower()
+
     can_render_file = (
         file_extension in current_app.config["SUPPORTED_RENDER_EXTENSIONS"]
     )
@@ -849,7 +853,10 @@ def generate_manifest(record_id: uuid.UUID) -> Response:
     file_name = file.FileName
     file_url = create_presigned_url(file)
     manifest_url = f"{url_for('main.generate_manifest', record_id=record_id, _external=True)}"
-    file_type = file.ffid_metadata.Extension.lower()
+    if file.ffid_metadata and file.ffid_metadata.Extension:
+        file_type = file.ffid_metadata.Extension.lower()
+    else:
+        file_type = file.FileName.split(".")[-1].lower()
 
     if (
         file_type
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -52,6 +52,7 @@ from app.main.util.render_utils import (
     generate_image_manifest,
     generate_pdf_manifest,
     get_download_filename,
+    get_file_extension,
 )
 from app.main.util.search_utils import (
     build_search_results_summary_query,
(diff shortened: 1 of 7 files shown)


app/main/routes.py
```diff
@@ -717,7 +717,11 @@ def record(record_id: uuid.UUID):
     validate_body_user_groups_or_404(file.consignment.series.body.Name)
 
     file_metadata = get_file_metadata(file.FileId)
-    file_extension = file.ffid_metadata.Extension.lower()
+    if file.ffid_metadata and file.ffid_metadata.Extension:
```
anthonyhashemi: Isn't 
```suggestion
    if file.ffid_metadata.Extension:
```

sufficien?

app/main/routes.py
```diff
@@ -849,7 +853,10 @@ def generate_manifest(record_id: uuid.UUID) -> Response:
     file_name = file.FileName
     file_url = create_presigned_url(file)
     manifest_url = f"{url_for('main.generate_manifest', record_id=record_id, _external=True)}"
-    file_type = file.ffid_metadata.Extension.lower()
+    if file.ffid_metadata and file.ffid_metadata.Extension:
```
anthonyhashemi: same here

app/tests/test_record.py
```diff
@@ -262,6 +262,38 @@ def test_record_record_alert_banner_is_visible_when_unsupported_file_extension(
         banner = soup.find("h2", string="Unable to display this record")
         assert banner is not None
 
+    @mock_aws
+    def test_record_record_alert_banner_is_visible_when_unsupported_file_extension_no_ffid_metadata(
```
anthonyhashemi, line 266: is this a real scenario we expect with the db?

app/tests/test_routes.py
```diff
@@ -31,6 +31,90 @@ def verify_cookies_header_row(data):
     ] == expected_row[0]
 
 
+def build_expected_pdf_manifest(file, extension):
```
anthonyhashemi: My preference would be to have left the epxected variables defined explcitly in the tests so that the tests are easily understood with minimal context switching. What was the reason to extract? The functions are not used anywhere else.

Was it just to make the test shorter?

app/tests/test_routes.py
```diff
@@ -133,49 +217,44 @@ def test_route_generate_pdf_manifest(
...
+        app.config["RECORD_BUCKET_NAME"] = bucket_name
+        create_mock_s3_bucket_with_object(bucket_name, file)
+
+        response = client.get(f"{self.record_route_url}/{file.FileId}/manifest")
+        assert response.status_code == 200
+
+        expected_pdf_manifest = build_expected_pdf_manifest(file, "pdf")
 
         actual_manifest = json.loads(response.text)
+        print(f"Expected PDF manifest:\n{expected_pdf_manifest}")
```
anthonyhashemi: Do we need these print statements? pytest -vvv should give us a full diff output I think

app/tests/test_routes.py
rahmedzaizi: No 
Added the another test 'test_route_generate_pdf_manifest_no_ffid_metadata` which checks if it's able to generate manifest even if ffid_metadata not available 
So it's used in 2 tests
Added it to avoid repetition

app/tests/test_routes.py
rahmedzaizi: Forgot to remove 
I just add prints to make it easier to see instead of using verbose 
will remove now

app/tests/test_record.py
rahmedzaizi: same as test above but checking when there's no FFIDMetadata

app/tests/test_routes.py
anthonyhashemi: oh sorry I had searched for it but didn't see the extra use in that test.

Thanks, the motivation makes sense now. 

I do wonder though whether we are testing at the right level. I think potentially we are doing too much edge case testing at a high level which results in these big similar asserts, just to cover an edge case around file extensions. Do you see what I mean?

colinbowen (approved):

PR #1142, opened 2025-08-20
add access copy converter

rahmedzaizi:
### After


data_management/access-copy-converter/access-copy-converter/main.py
```diff
@@ -0,0 +1,162 @@
...
+        raise
+
+    consignment_reference = sns_message.get("parameters", {}).get("reference")
+
+    if not consignment_reference:
+        raise Exception(
+            "Missing reference in SNS Message required for indexing"
+        )
+
+    process_consignment(consignment_reference)
```
colinbowen: Not needed, but would be good to have general exception handling for the process function here

colinbowen:
Can approve once the codeql change is done.

colinbowen (approved):

PR #1143, opened 2025-08-20
render using access copy bucket

rahmedzaizi:
## Changes in this PR
Updated routes to render if file_extension is in CONVERTIBLE_EXTENSIONS
### After

- [x] Requires env variable(s) to be updated


colinbowen (dismissed):

colinbowen (approved):
lgtm

PR #1185, opened 2025-09-05
added bulk access copy converter

rahmedzaizi:
### After


data_management/access_copy_converter/Dockerfile.bulk_access_copy_converter
```diff
@@ -0,0 +1,23 @@
+FROM python:3.9-slim
```
anthonyhashemi, line 1: can the filepath just be `data_management/access_copy_converter/Dockerfile`?

anthonyhashemi, line 1: can't we use a higher version of python? Ideally 3.12 or 3.13, 3.12?

anthonyhashemi: oh sorry just sawe the other dockerfile in this folder

data_management/access_copy_converter/bulk_access_copy_converter/main.py
```diff
@@ -0,0 +1,197 @@
...
+    try:
+        s3.head_object(Bucket=bucket, Key=key)
+        return True
+    except ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            return False
+        raise
+
+
+def get_extension(file_id, conn, metadata):
```
anthonyhashemi, line 44: isn't a lot of this duplicate work from `data_management/access_copy_converter/access_copy_converter/main.py`? Can we try to abstract it out and reuse a shared function?

data_management/access_copy_converter/Dockerfile.bulk_access_copy_converter
```diff
@@ -0,0 +1,23 @@
...
+
```
anthonyhashemi, line 2: do we even need a specific dockerfile for this? Can't we just override the CMD/entrypoint to point to the bulk function, I think all the deps are the same?

data_management/access_copy_converter/bulk_access_copy_converter/main.py
rahmedzaizi: yeah was rushing this a little bit 
Wanted to refactor and both functions and create a shared folder for both functions 
we could create a separate refactor ticket as the focus is on excel conversion now

data_management/access_copy_converter/Dockerfile.bulk_access_copy_converter
rahmedzaizi: We could 
would have to update workflow that builds the images 
preferably want to keep the workflow generic though

anthonyhashemi: lets chat on slack

PR #1209, opened 2025-09-12
AYR-1734 added functions for bulk access copy convertion

rahmedzaizi:
### After


data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,107 +71,169 @@ def get_extension(file_id, conn, metadata):
...
+def main():
+    app_secret_id = os.getenv("APP_SECRET_ID")
+    app_secret = get_secret_string(app_secret_id)
+    source_bucket = app_secret["RECORD_BUCKET_NAME"]
+    dest_bucket = app_secret["ACCESS_COPY_BUCKET"]
+    convertible_extensions = set(
+        json.loads(app_secret["CONVERTIBLE_EXTENSIONS"])
+    )
+    engine = get_engine()
+    metadata = MetaData()
```
anthonyhashemi: why instantiate this here and not inside `process_consignment` or even `get_extension`?

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -50,7 +57,11 @@ def get_extension(file_id, conn, metadata):
     except SQLAlchemyError as e:
         logger.error(f"Error querying FFIDMetadata table: {e}")
         conn.rollback()
+        return "ERROR"
```
anthonyhashemi: discussed: just raise error, and catch it in calling function

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,107 +71,169 @@ def get_extension(file_id, conn, metadata):
...
+            logger.info(f"FileName: {filename}")
             if "." in filename:
                 return filename.rsplit(".", 1)[1].lower()
+            logger.warning("No extension in FileName")
     except SQLAlchemyError as e:
         logger.error(f"Error querying File table: {e}")
         conn.rollback()
-
-    return None
+        return "ERROR"
```
anthonyhashemi: same thing here

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,107 +71,169 @@ def get_extension(file_id, conn, metadata):
...
-                    dest_key = f"{consignment_ref}/{file_id}"
-                    s3.upload_file(output_path, dest_bucket, dest_key)
-                    logger.info("Uploaded converted PDF to Access Copy Bucket")
-            else:
+            if extension not in convertible_extensions:
                 logger.info(f"File {file_id} does not require conversion")
+                continue
+
+            dest_key = f"{consignment_ref}/{file_id}"
+            if already_converted(dest_bucket, dest_key):
```
anthonyhashemi, line 146: if its in access bucket already, then dont bother 👍

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,107 +71,169 @@ def get_extension(file_id, conn, metadata):
...
+    source_bucket = app_secret["RECORD_BUCKET_NAME"]
+    dest_bucket = app_secret["ACCESS_COPY_BUCKET"]
+    convertible_extensions = set(
+        json.loads(app_secret["CONVERTIBLE_EXTENSIONS"])
+    )
+    engine = get_engine()
+    metadata = MetaData()
+    conversion_type = os.getenv("CONVERSION_TYPE")
+    if conversion_type == "BULK":
+        paginator = s3.get_paginator("list_objects_v2")
```
anthonyhashemi: can we pull each of these blocks into their own functions, just to keep this high level easier to parse at a glance?

`create_access_copies_for_all_consingments`

and

`create_access_copy_from_sns`

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,167 @@ def get_extension(file_id, conn, metadata):
...
+def process_consignment(
+    consignment_ref,
+    source_bucket,
+    dest_bucket,
+    convertible_extensions,
+):
     prefix = f"{consignment_ref}/"
-
     engine = get_engine()
     metadata = MetaData()
```
rahmedzaizi, line 117: MetaData object is needed for Table Class
[metadata_object](https://docs.sqlalchemy.org/en/20/tutorial/metadata.html#:~:text=the%20ORM%E2%80%99s%20configuration.-,Setting%20up%20MetaData%20with%20Table%20objects,be%20how%20we%20will%20refer%20to%20the%20table%20in%20application%20code%3A,-%3E%3E%3E%20from%20sqlalchemy)

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
         result = conn.execute(stmt).first()
         if result and result[0]:
             filename = result[0]
+            logger.info(f"FileName: {filename}")
```
anthonyhashemi: SECURITY CONCERN: dont log filenames, they can be sensitive; specifically can hold redacted /closed names

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
...
+            with tempfile.TemporaryDirectory() as tmpdir:
+                input_path = os.path.join(tmpdir, f"input.{extension}")
+                output_path = os.path.join(tmpdir, "input.pdf")
+
+                s3.download_file(source_bucket, key, input_path)
+                logger.info(f"Downloaded {key} to {input_path}")
+
+                try:
+                    if extension in ("xls", "xlsx"):
+                        temp_ods = os.path.join(tmpdir, "input.ods")
```
anthonyhashemi: Do you think this block could be its own function, so we can have a focussed test for converting spreadsheets?

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
...
                 "--headless",
                 "--convert-to",
-                "pdf",
+                convert_to,
                 "--outdir",
                 os.path.dirname(output_path),
                 input_path,
             ],
             check=True,
+            stdout=subprocess.PIPE,
```
anthonyhashemi: why are these needed?

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
...
+    consignments = set()
+    for page in response:
+        for obj in page["Contents"]:
+            consignments.add(obj["Key"].split("/")[0])
 
+    logger.info(f"Found {len(consignments)} consignments")
 
-def main():
+    for consignment_ref in consignments:
+        return consignment_ref
```
anthonyhashemi: Why are we returning the first confignment_ref?

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
...
-                    except Exception as e:
-                        logger.error(f"Conversion failed for {file_id}: {e}")
-                        continue
-
-                    dest_key = f"{consignment_ref}/{file_id}"
-                    s3.upload_file(output_path, dest_bucket, dest_key)
-                    logger.info("Uploaded converted PDF to Access Copy Bucket")
-            else:
-                logger.info(f"File {file_id} does not require conversion")
+def create_access_copies_for_all_consignments(source_bucket):
```
anthonyhashemi: this function is not creating any access copies, it just returns the consigment_ref found in the s3 bucket?

data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -60,106 +70,169 @@ def get_extension(file_id, conn, metadata):
...
+            consignments.add(obj["Key"].split("/")[0])
 
+    logger.info(f"Found {len(consignments)} consignments")
 
-def main():
+    for consignment_ref in consignments:
+        return consignment_ref
+
+
+def create_access_copy_from_sns():
```
anthonyhashemi: this function does not create any access copies, it just returns the consignment_reg mentioned in the sns message

data_management/access_copy_converter/access_copy_converter/main.py
anthonyhashemi: alternatively could use a sqlalchemy;s orm model definitions like we do in main app and the data_management conftest.py file but this is fine for now

data_management/access_copy_converter/access_copy_converter/main.py
rahmedzaizi: Was debugging something 
Will remove now

data_management/access_copy_converter/access_copy_converter/main.py
rahmedzaizi: removed

anthonyhashemi:
Thanks for addressing the comments.

Could we add some tests. I think one success path e2e test plus some lower level tests would be good

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.fixture(autouse=True)
+def patch_boto_clients(monkeypatch):
+    s3 = mock.Mock()
+    sm = mock.Mock()
+    monkeypatch.setattr(converter, "s3", s3)
+    monkeypatch.setattr(converter, "sm", sm)
+    return {"s3": s3, "sm": sm}
+
+
+@pytest.fixture
```
anthonyhashemi, line 28: can we reuse the engine defined in `data_management/conftest.py`?

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -0,0 +1,255 @@
+import subprocess
+from unittest import mock
+
+import access_copy_converter.main as converter
```
anthonyhashemi: can we import the functions from main that we are testing instead, bit more standard that way; reduces the noise of `converter.` everywhere.

The `patch_boto_clients` would have to change but Id recommend just using `from moto import mock_aws` and decorate each test function that needs it:
https://docs.getmoto.org/en/latest/docs/getting_started.html

```
import boto3
from moto import mock_aws
from mymodule import MyModel

@mock_aws
def test_my_model_save():
    conn = boto3.resource("s3", region_name="us-east-1")
    # We need to create the bucket since this is all in Moto's 'virtual' AWS account
    conn.create_bucket(Bucket="mybucket")

    s3 = boto3.client("s3", region_name="us-east-1")
    s3.put_object(Bucket="mybucket", Key="steve", Body="is awesome")
    model_instance.save()

    body = conn.Object("mybucket", "steve").get()[
        "Body"].read().decode("utf-8")

    assert body == "is awesome"
```

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -0,0 +1,255 @@
...
+    assert "Error querying FFIDMetadata table" in str(exc.value)
+
+
+def test_get_extension_file_query_failure(monkeypatch, sqlite_conn):
+    conn, metadata, ffid, file_table = sqlite_conn
+
+    def execute_raise(stmt, *args, **kwargs):
+        s = str(stmt).lower()
+
+        if "ffidmetadata" in s:
```
anthonyhashemi: why do we need this conditional?

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -0,0 +1,255 @@
...
+
+        raise SQLAlchemyError("file error")
+
+    monkeypatch.setattr(conn, "execute", execute_raise)
+    with pytest.raises(Exception) as exc:
+        converter.get_extension("file123", conn, metadata)
+    assert "Error querying File table" in str(exc.value)
+
+
+def test_convert_with_libreoffice_success(monkeypatch, tmp_path):
```
anthonyhashemi: Can we split these tests into sections so its easy to see where were testing different functions

can group tests for a particular function in a class like 

`class TestConvertWithLibreOffice`

data_management/access_copy_converter/tests/test_access_copy_converter.py
rahmedzaizi: It returns a mock None for FFID table so it falls back to File table

anthonyhashemi: but its only used in one place, so I dont see the need for the condition, just do the fallback unconditonally?

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -0,0 +1,255 @@
...
+    assert "No extension in FileName" in caplog.text
+
+
+def test_get_extension_ffid_query_failure(monkeypatch, sqlite_conn):
+    conn, metadata, ffid, file_table = sqlite_conn
+    orig_execute = conn.execute
+
+    def execute_raise(stmt, *args, **kwargs):
+        s = str(stmt).lower()
+        if "ffidmetadata" in s:
```
anthonyhashemi: same here

anthonyhashemi (approved):
Great stuff guys, thanks for the refactors

PR #1262, opened 2025-09-30
Fix for search results when there is no highlights

karwan-zaizi:
## Changes in this PR
Showing search results if there is only `transferring_body` without highlights 

## JIRA ticket
[AYR-1749]([AYR-1749](https://national-archives.atlassian.net/browse/AYR-1749))

### Before
6bfb89d3-23e5-4061-8cf8-b3563485495c

### After
it should look like this without the highlight 
f7928edc-06d1-4239-97c8-f37cf628db15

[AYR-1749]: https://national-archives.atlassian.net/browse/AYR-1749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AYR-1749]: https://national-archives.atlassian.net/browse/AYR-1749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/templates/main/search-transferring-body.html
+++ b/app/templates/main/search-transferring-body.html
@@ -132,13 +132,9 @@
                                                             </td>
                                                         {% else %}
                                                             {% set file_name = record['_source']['file_name'] %}
+                                                            <td class="govuk-table__cell">File name</td>
                                                             <td class="govuk-table__cell">
-                                                                File name
-                                                            </td>
-                                                            <td class="govuk-table__cell">
-                                                                <a href="{{ url_for('main.record', record_id=record['_source']['file_id']) }}">
-                                                                    {{ file_name }}
-                                                                </a>
+                                                                <a href="{{ url_for('main.record', record_id=record['_source']['file_id']) }}">{{ file_name }}</a>
                                                             </td>
                                                         {% endif %}
                                                     </tr>
--- a/app/templates/main/search-transferring-body.html
+++ b/app/templates/main/search-transferring-body.html
@@ -100,74 +100,80 @@
                                                 {% for record in results %}
                                                     {% set main_loop_index = loop.index %}
                                                     <tr class="govuk-table__row govuk-table__row--primary">
-                                                        {% set first_field = record['highlight'].keys() | list | first %}
-                                                        {% set first_field_formatted = first_field | format_opensearch_field_name %}
-                                                        {% set first_result_value = record['highlight'].values() | list | first | join(' ... ')| clean_tags_and_replace_highlight_tag(highlight_tag) | safe %}
-                                                        {% set total_fields = record['highlight'].keys() | length %}
-                                                        {% set file_name = record['_source']['file_name'] %}
-                                                        <td class="govuk-table__cell">
-                                                            <div class="govuk-table__cell__data">
-                                                                <button disabled
-                                                                        data-main-row="{{ main_loop_index }}"
-                                                                        data-file-name="{{ file_name }}"
-                                                                        class="govuk-table__toggle-button"
-                                                                        aria-label="Expand rows for file {{ file_name }}"
-                                                                        aria-controls="{% for i in range(total_fields) %}hidden-row-{{ main_loop_index }}-{{ i }}{% if not loop.last %} {% endif %}{% endfor %}"
-                                                                        aria-expanded="false"
-                                                                        type="button"></button>
-                                                                {% if record['highlight'] %}
+                                                        {% if record['highlight'] %}
+                                                            {% set first_field = record['highlight'].keys() | list | first %}
+                                                            {% set first_field_formatted = first_field | format_opensearch_field_name %}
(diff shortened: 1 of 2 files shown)


app/templates/main/search-transferring-body.html
```diff
@@ -100,74 +100,76 @@
...
-                                                                <button disabled
-                                                                        data-main-row="{{ main_loop_index }}"
-                                                                        data-file-name="{{ file_name }}"
-                                                                        class="govuk-table__toggle-button"
-                                                                        aria-label="Expand rows for file {{ file_name }}"
-                                                                        aria-controls="{% for i in range(total_fields) %}hidden-row-{{ main_loop_index }}-{{ i }}{% if not loop.last %} {% endif %}{% endfor %}"
-                                                                        aria-expanded="false"
-                                                                        type="button"></button>
-                                                                {% if record['highlight'] %}
+                                                        {% if record['highlight'] %}
```
anthonyhashemi, line 103: could we add a test for this scenario please

karwan-zaizi: @anthonyhashemi  Unit test has been added for results with no highlights

app/tests/test_search.py
```diff
@@ -2221,3 +2221,132 @@ def test_search_transferring_body_with_quotes(
...
+
+        response = client.get(
+            f"/search/transferring_body/{transferring_body_id}?query=plain"
+        )
+        assert response.status_code == 200
+        soup = BeautifulSoup(response.data, "html.parser")
+        table_body = soup.find("tbody")
+        rows = table_body.find_all("tr")
+        found = False
+        for row in rows:
```
anthonyhashemi: can we avoid looping through all rows and instead use a selector / find method which gets the specific row / element, like in the other tests in this file?

app/tests/test_search.py
```diff
@@ -2221,3 +2221,132 @@ def test_search_transferring_body_with_quotes(
...
+        response = client.get(
+            f"/search/transferring_body/{transferring_body_id}?query=plain"
+        )
+        assert response.status_code == 200
+        soup = BeautifulSoup(response.data, "html.parser")
+        table_body = soup.find("tbody")
+        rows = table_body.find_all("tr")
+        # Find the row with file name
+        found = False
+        for row in rows:
```
anthonyhashemi: similar to the other comment

app/tests/test_search.py
karwan-zaizi: @anthonyhashemi Updated both tests to make it similar to others

app/tests/test_search.py
karwan-zaizi: @anthonyhashemi same here

app/tests/test_search.py
```diff
@@ -2274,18 +2274,11 @@ def test_search_transferring_body_no_highlights_renders_file_name_and_search_res
...
-                link = cells[1].find("a")
-                assert link
-                assert file_name in link.get_text()
-                assert url_for("main.record", record_id=file_id) in link["href"]
-                found = True
-        assert found
+        table_rows_cell_values = get_table_rows_cell_values(table_body)
+        assert ["File name", file_name] in table_rows_cell_values
+
+        assert ["series_name", "series_y"] not in table_rows_cell_values
```
anthonyhashemi: I don't quite understand the intention of this assertion; the data that we mock doesn't include a "series_y" so this would inevitably true. same with cref2.

You could change it to series_x and cref if you wanted to show that those aren't in it. Alternatively you could just assert the exact response of what table_rows_cell_values looks like 
e.g. `assert expected == table_rows_cell_values`

app/tests/test_search.py
```diff
@@ -2339,14 +2332,8 @@ def test_search_transferring_body_empty_highlights_renders_file_name_and_search_
...
-        for row in rows:
-            cells = row.find_all("td")
-            if len(cells) == 2 and "File name" in cells[0].get_text():
-                link = cells[1].find("a")
-                assert link
-                assert file_name in link.get_text()
-                assert url_for("main.record", record_id=file_id) in link["href"]
-                found = True
-        assert found
+        table_rows_cell_values = get_table_rows_cell_values(table_body)
```
anthonyhashemi, line 2335: Ah I see now series_y makes sense for this test but not the one above, I guess was just typo. But yeah, I think due to this, a test asserting the expected table_rows_cell_values is probably better than asserting things not in it, as it avoids this possibility of typos resulting in unhelpful assertions

app/tests/test_search.py
karwan-zaizi: My mistake, I was going bottom to top so I just copied what from the second test, I didn't realise the values were different, it is updated now

anthonyhashemi (approved):

PR #1289, opened 2025-10-08
Fallback for access copy failure rendering

rahmedzaizi:
### After


app/templates/main/record-view.html
```diff
@@ -17,6 +17,10 @@
             <script id="init-uv" nonce="{{ csp_nonce() }}" manifest_url="{{ manifest_url }}" type="text/javascript" src="{{ url_for('static', filename='init.uv.js') }}">
             </script>
         </div>
+    {% elif access_copy_failed %}
```
anthonyhashemi, line 20: can we have a test for this new scenario, especially as the logic between `can_render_file` and `access_copy_failed` are intertwined by that try-except block.

anthonyhashemi (approved):

PR #1291, opened 2025-10-08
fixed closed_connection error on access_copy_converter

rahmedzaizi:
### After

### DIFF
--- a/data_management/access_copy_converter/access_copy_converter/main.py
+++ b/data_management/access_copy_converter/access_copy_converter/main.py
@@ -125,50 +125,49 @@ def process_consignment(
         f"{len(response['Contents'])} files in consignment {consignment_ref}"
     )
 
-    with conn:
-        for obj in response["Contents"]:
-            key = obj["Key"]
-            file_id = key.split("/")[-1]
-            logger.info(f"Checking file_id: {file_id}")
+    for obj in response["Contents"]:
+        key = obj["Key"]
+        file_id = key.split("/")[-1]
+        logger.info(f"Checking file_id: {file_id}")
+
+        try:
+            extension = get_extension(file_id, conn, metadata)
+        except Exception as e:
+            logger.error(f"Failed to get file_extension for {file_id}: {e}")
+            continue
+        logger.info(f"File extension: {extension}")
+
+        if extension not in convertible_extensions:
+            logger.info(f"File {file_id} does not require conversion")
+            continue
+
+        dest_key = f"{consignment_ref}/{file_id}"
+        if already_converted(dest_bucket, dest_key):
+            logger.info(f"Skipping {file_id}, already converted")
+            continue
+
+        logger.info(f"File {file_id} requires conversion to PDF")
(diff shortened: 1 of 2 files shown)


data_management/access_copy_converter/access_copy_converter/main.py
```diff
@@ -125,50 +125,49 @@ def process_consignment(
         f"{len(response['Contents'])} files in consignment {consignment_ref}"
     )
 
-    with conn:
-        for obj in response["Contents"]:
-            key = obj["Key"]
-            file_id = key.split("/")[-1]
-            logger.info(f"Checking file_id: {file_id}")
+    for obj in response["Contents"]:
```
anthonyhashemi, line 128: this change is needed because we now pass in the connection to the function as opposed to creating one per consignment

anthonyhashemi (commented):
My usual comment, can a test catch this?

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -275,3 +275,49 @@ def test_process_consignment_skips_nonconvertible(
...
+                CreateBucketConfiguration={"LocationConstraint": "eu-west-2"},
+            )
+        s3_client.put_object(
+            Bucket="source-bucket", Key="cons1/fileA", Body=b"A"
+        )
+        s3_client.put_object(
+            Bucket="source-bucket", Key="cons2/fileB", Body=b"B"
+        )
+
+        monkeypatch.setattr(main_module, "s3", s3_client)
```
anthonyhashemi, line 296: you shouldnt need to monkeypatch s3 when using mock_aws from moto.

See https://docs.getmoto.org/en/latest/docs/getting_started.html#moto-usage

data_management/access_copy_converter/tests/test_access_copy_converter.py
```diff
@@ -275,3 +275,49 @@ def test_process_consignment_skips_nonconvertible(
...
+        conn.execute(insert(ffid).values(FileId="fileA", Extension="docx"))
+        conn.execute(insert(ffid).values(FileId="fileB", Extension="docx"))
+        conn.commit()
+
+        def fake_convert(input_path, output_path, convert_to="pdf"):
+            from pathlib import Path
+
+            Path(output_path).write_bytes(b"%PDF-1.4")
+
+        monkeypatch.setattr(
```
anthonyhashemi, line 308: I know for the intention of this test mocking the conversion isnt a problem, but out of interest, do we have any tests that actually run the conversion?

rahmedzaizi: Yeah under class `TestConvertWithLibreoffice:`

data_management/access_copy_converter/tests/test_access_copy_converter.py
anthonyhashemi: apparently doesnt work without, probably because the s3 client inside of process_consignment does still try to build a real client not use moto

anthonyhashemi (approved):

PR #1306, opened 2025-10-13
Ayr 1677 uv page previews

karwan-zaizi:
## Changes in this PR
Set page previews to display within universal viewer render for multiple page PDFs

## JIRA ticket
[AYR-1677](https://national-archives.atlassian.net/browse/AYR-1677)

### Before
Screenshot 2025-10-13 at 12 43 01

### After
Screenshot 2025-10-13 at 12 43 30

[AYR-1677]: https://national-archives.atlassian.net/browse/AYR-1677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/main/util/render_utils.py
```diff
@@ -128,11 +286,26 @@ def generate_image_manifest(
...
+    elif image_format == "gif":
+        mime_type = "image/gif"
+    elif image_format == "webp":
+        mime_type = "image/webp"
+    else:
+        mime_type = f"image/{image_format}"
+
     manifest = {
-        "@context": "https://iiif.io/api/presentation/3/context.json",
+        "@context": "http://iiif.io/api/presentation/3/context.json",
```
anthonyhashemi: Minor but this redirects to https so might as well use the https version
```suggestion
        "@context": "https://iiif.io/api/presentation/3/context.json",
```

app/main/util/render_utils.py
```diff
@@ -45,7 +46,11 @@ def get_download_filename(file):
 
 
 def create_presigned_url(file: File) -> str:
-    s3 = boto3.client("s3")
+    s3 = boto3.client(
+        "s3",
+        endpoint_url=current_app.config.get("AWS_ENDPOINT_URL"),
+        verify=False,  # Disable SSL verification for self-signed certificates
```
anthonyhashemi: this should remain on in production. Why does this even need to change, the manifests were working fine before locally?

app/main/util/render_utils.py
```diff
@@ -128,11 +286,26 @@ def generate_image_manifest(
     image = Image.open(io.BytesIO(s3_file_object["Body"].read()))
     image_width, image_height = image.size
 
+    # Detect image format
+    image_format = image.format.lower() if image.format else "png"
```
anthonyhashemi, line 223: ah good spot, I wonder why it didnt matter to universal viewer that we specified this as png for all of them before (for clarity this doesnt impact the thumbnails)

app/main/util/render_utils.py
```diff
@@ -128,11 +286,26 @@ def generate_image_manifest(
...
+    else:
+        mime_type = f"image/{image_format}"
+
     manifest = {
-        "@context": "https://iiif.io/api/presentation/3/context.json",
+        "@context": "http://iiif.io/api/presentation/3/context.json",
         "@id": manifest_url,
         "@type": "sc:Manifest",
-        "label": {"en": [file_name]},
+        "label": file_name,
```
anthonyhashemi: are you sure it doesnt need the ` {"en": [file_name]},`?

app/main/util/render_utils.py
```diff
@@ -152,8 +325,8 @@ def generate_image_manifest(
                                 "motivation": "sc:painting",
                                 "resource": {
                                     "@id": file_url,
-                                    "type": "dctypes:Image",
-                                    "format": "image/png",
+                                    "@type": "dctypes:Image",
+                                    "format": mime_type,
```
anthonyhashemi, line 262: Not for this line but for the manifests we generate in. general, Im wondering if there is an api or package we can use to verify their validity, as we obviously have got it wrong before

app/tests/test_routes.py
```diff
@@ -225,12 +220,12 @@ def test_route_generate_image_manifest(
                                     "@id": f"https://presigned-url.com/download.{file.ffid_metadata.Extension}",
                                     "@type": "oa:Annotation",
                                     "motivation": "sc:painting",
-                                    "on": f"https://presigned-url.com/download.{file.ffid_metadata.Extension}",
+                                    "on": f"http://localhost/record/{file.FileId}/manifest/canvas/1",
                                     "resource": {
-                                        "@id": f"https://presigned-url.com/download.{file.ffid_metadata.Extension}",
+                                        "@id": f"http://localhost/record/{file.FileId}/manifest/annotation/1",
+                                        "@type": "dctypes:Image",
```
anthonyhashemi: not in this test file, but can we have new test(s) for `generate_pdf_manifest`  but at the lower level where we have all the new edge cases / if statements you have added

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
     return presigned_url
 
 
+def extract_pdf_pages_as_images(pdf_bytes: bytes) -> List[dict]:
```
anthonyhashemi, line 61: tests for this would be good

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
...
+        Response: Flask JSON response containing the IIIF manifest.
+    """
+    FALLBACK_WIDTH = 800
+    FALLBACK_HEIGHT = 1000
+    page_data = []
+    current_app.logger.info(
+        f"Generating PDF manifest for {file_name}, file_obj: {file_obj is not None}"
+    )
+
+    if file_obj:
```
anthonyhashemi: I suggest removing this condition, there's no scenario where we wouldn't have this object to pass in, so I think the condition just makes the function unnecessarily complex

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
...
+            )
+            return jsonify({"error": "Failed to process PDF"}), 500
+    else:
+        current_app.logger.warning(
+            "No file_obj provided, cannot extract PDF pages"
+        )
+
+    canvas_items = []
+
+    if page_data:
```
anthonyhashemi: same here, we don't expect page_data to ever be empty so I'd remove this condition and the fallback. Or is there a scenario where it could be?

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
...
+    current_app.logger.info(
+        f"Generating PDF manifest for {file_name}, file_obj: {file_obj is not None}"
+    )
+
+    if file_obj:
+        try:
+            s3 = boto3.client(
+                "s3",
+                endpoint_url=current_app.config.get("AWS_ENDPOINT_URL"),
+                verify=False,
```
anthonyhashemi: again, we need to make sure we ar verifying in production, and again would ask why we need to turn this off in local if this api call is working on main in local

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
...
+            )
+            bucket = None
+            if file_name.split(".")[-1].lower() == "pdf":
```
anthonyhashemi: we should be getting file type from FFIDMetadata, not the file name to be consistent

app/main/util/render_utils.py
```diff
@@ -56,69 +61,222 @@ def create_presigned_url(file: File) -> str:
...
+                bucket = current_app.config["RECORD_BUCKET_NAME"]
+            else:
```
anthonyhashemi: I suggest moving this logic out of this function to leave this function just responsible for creating the pdf manifest and not having to worry about different s3 buckets etc.
We already have some logic that cares about record bucket vs acces copy bucket in the main `generate_manifest` route which calls this function, so I think all the info is available there.

Also it doesnt look like you are passing file_obj in at the moment anyway, so not sure how it would be working?

Image

app/static/init.uv.js
```diff
@@ -35,66 +35,72 @@ function initUniversalViewer() {
...
+function enhanceButtons() {
   document.querySelectorAll(".btn").forEach((button) => {
     if (button.tagName.toLowerCase() === "div") {
       button.setAttribute("role", "button");
+      button.setAttribute("tabindex", "0");
     }
   });
+}
+
+// MutationObserver to remove attribution if it appears
```
anthonyhashemi: I havent looked at the changes in the js at all yet

app/main/util/render_utils.py
karwan-zaizi: This change has been made

app/main/util/render_utils.py
karwan-zaizi: I was testing something locally thought that was the issue and forgot to change it back

app/main/util/render_utils.py
karwan-zaizi: Sorry, I read in an article that we don't need it, but in the official page suggests it, changed back to it's original

app/main/util/render_utils.py
karwan-zaizi: You are right, now I double checked we definitely don't need it

app/static/init.uv.js
karwan-zaizi: Tested the changes with the old version, it looks like it's working, I will change it back

karwan-zaizi: No need to review this

app/main/util/render_utils.py
karwan-zaizi: Added couple of tests for this

app/main/util/render_utils.py
karwan-zaizi: No, it doesn't impact the thumbnails

app/main/util/render_utils.py
karwan-zaizi: Made it look like this:
Screenshot 2025-10-14 at 16 31 30

app/tests/test_routes.py
karwan-zaizi: I have add couple of tests for this

app/main/routes.py
```diff
@@ -858,41 +858,49 @@ def http_exception(error):
...
+        s3_file_object = s3.get_object(Bucket=bucket, Key=key)
+        return s3_file_object
+
     if (
         file_type
         in current_app.config["UNIVERSAL_VIEWER_SUPPORTED_APPLICATION_TYPES"]
     ):
-        file_url = create_presigned_url(file)
-        return generate_pdf_manifest(file_name, file_url, manifest_url)
+        file_obj = get_s3_file_obj(file)
```
anthonyhashemi: this line is repeated in each conditional block. Instead it can be moved above the if block to reduce duplication and simplify the logic

app/tests/test_routes.py
```diff
@@ -113,72 +113,56 @@ def test_route_terms_of_use(self, client: FlaskClient):
 
     @mock_aws
     @patch("app.main.routes.create_presigned_url")
+    @patch("app.main.routes.generate_pdf_manifest")
     def test_route_generate_pdf_manifest(
         self,
+        mock_generate_pdf_manifest,
```
anthonyhashemi: What was the motivation in changing this test to mock generate_pdf_manifest?

karwan-zaizi: It complained about too many values to unpack

app/main/routes.py
karwan-zaizi: Made the change

app/main/util/render_utils.py
```diff
@@ -56,6 +58,74 @@ def create_presigned_url(file: File) -> str:
...
+                    }
+                )
+
+                # Clean up resources
+                page_image.close()
+                thumbnail.close()
+                pix = None
+
+            return page_data
+    except pymupdf.fitz.FileDataError as e:
```
anthonyhashemi: can you put this try except block around the specific line that raises it

app/main/util/render_utils.py
```diff
@@ -72,53 +142,103 @@ def create_presigned_url_for_access_copy(file: File) -> str:
...
+        current_app.logger.info(f"PDF bytes length: {len(pdf_bytes)}")
+
+        page_data = extract_pdf_pages_as_images(pdf_bytes)
+        current_app.logger.info(f"Extracted {len(page_data)} pages from PDF")
+        return page_data, None
+    except Exception as e:
+        current_app.logger.error(
+            f"Error processing PDF for thumbnails: {e}", exc_info=True
+        )
+        return None, jsonify({"error": "Failed to process PDF"}), 500
```
anthonyhashemi: This is entirely fine to do, it works, but we could just re raise the error rather than having the return signature being a tuple of (page_data, error_response).

Apart from generally being the more pythonic/standard approach, one major benefit is that even if we don't explicitly catch the raised error and do something with it in the flask route this gets called in, it will propagate and trigger flask

Whereas with this approach, the error could technically pass silently if we dont bother to handle it in the calling function.

See https://peps.python.org/pep-0020/#the-zen-of-python

This is unlike, say, Rust which uses something akin to a tuple https://doc.rust-lang.org/std/result/ but it enforces that the error is handled by its compiler.

Go similarly uses the result, err pattern, but for all functions. It doesnt enforce that you handle the errors but because thats the standard approach it makes sense.

Of course, python can do this, but it is not the standard convention and thus I suggest keeping to the standard for simplicity.

app/tests/test_render_utils.py
```diff
@@ -83,3 +87,145 @@ def test_create_presigned_url(mock_boto_client):
...
+def app_context():
+    app = Flask(__name__)
+    with app.app_context():
+        yield
+
+
+@patch("app.main.util.render_utils.pymupdf")
+@patch("app.main.util.render_utils.Image")
+@patch("app.main.util.render_utils.current_app")
+def test_extract_pdf_pages_as_images_success(
```
anthonyhashemi: I think some better tests for `extract_pdf_pages_as_images` would actually test the function without mocking all its internals. As a dev reading this test, I can't understand the purpose of this test without going in and reading the full function to understand what each of these mocked methods are doing. Essentially, we are testing implementation here, rather than the function itself. If the implementation of this function changes, this test breaks, but really we shouldn't care about that.

Mocking is definitely useful in certain cases but I'd suggest to use it sparingly. When you find yourself having to mock as much as in this test, it is generally worthwhile reconsidering the testing approach.


As an alternative, I would suggest something like:


```
def test_extract_pdf_pages_as_images_success():
    pdf_bytes = # some actual input pdf bytes
    expected_pages = [{
                        "page_number": 1,
                        "width": w1,
                        "height": h1,
                        "thumbnail_url": t1,
                        "page_image_url": pu1,
                        "thumbnail_base64_size": page_base64_size1,
                        "page_base64_size": page_base64_size1,
                    }, ...same but for page 2
    ]
    assert extract_pdf_pages_as_images(pdf_bytes) == expected_pages

File: app/main/util/render_utils.py (hunk 16 above)
Comment (karwan-zaizi) [reply app/main/util/render_utils.py]: Another way to do this we can assign a variable to the file and check for error, it's not different from what we have which we use this line `with pymupdf.open("pdf", io.BytesIO(pdf_bytes)) as pdf_document:`

Comment (anthonyhashemi) [reply app/main/util/render_utils.py]: oh are you saying that this is the smallest context window, as it is the line in the with block. In that case, this is fine

Comment (karwan-zaizi) [reply app/main/util/render_utils.py]: That's right, it checks the block code and checks for error

File: app/tests/test_routes.py (hunk 15 above)
Comment (anthonyhashemi) [reply app/tests/test_routes.py]: not sure I understand, what complained?

File: app/tests/test_render_utils.py (hunk 18 above)
Comment (karwan-zaizi) [reply app/tests/test_render_utils.py]: This is been addressed

File: app/main/util/render_utils.py (hunk 17 above)
Comment (karwan-zaizi) [reply app/main/util/render_utils.py]: Done, change has been made

File: app/tests/test_render_utils.py [hunk 19]
```diff
@@ -83,3 +87,116 @@ def test_create_presigned_url(mock_boto_client):
...
+        with open("local_services/mds_data_generator/example_files/file.pdf", "rb") as f:
+            pdf_bytes = f.read()
+
+        result = extract_pdf_pages_as_images(pdf_bytes)
+
+        expected_pages = [
+            {
+                "page_number": 1,
+                "width": result[0]["width"],
+                "height": result[0]["height"],
```
anthonyhashemi: this tests the expected structure of the dictionary and the expected keys, but all the values are just asserting against themselves. We should be asserting against static explicit valuers which we expect, given the input pdf.


e.g. `"width":  500 (probably isnt 500 but whatever it is) etc.

app/main/routes.py
```diff
@@ -858,41 +858,51 @@ def http_exception(error):
...
-        s3_file_object = s3.get_object(Bucket=bucket, Key=key)
         file_url = create_presigned_url(file)
         return generate_image_manifest(
-            file_name, file_url, manifest_url, s3_file_object
+            file_name, file_url, manifest_url, s3_file_object=s3_file_obj
         )
     elif file_type in convertible_extensions:
-        file_url = create_presigned_url_for_access_copy(file)
-        return generate_pdf_manifest(file.FileName, file_url, manifest_url)
+        file_obj = get_s3_file_obj(
```
anthonyhashemi, line 900: you don't need to create it again, it is already defined on line 882 just with a different name

app/main/util/render_utils.py
anthonyhashemi: I still see it here. Let me know if you want to talk through it

app/main/util/render_utils.py
```diff
@@ -56,6 +58,66 @@ def create_presigned_url(file: File) -> str:
...
+                    }
+                )
+
+                # Clean up resources
+                page_image.close()
+                thumbnail.close()
+                pix = None
+
+            return page_data
+    except Exception as e:
```
anthonyhashemi: maybe remove this, as we dont want to silently continue

app/tests/test_routes.py
anthonyhashemi: On Friday we discussed this offline and found the code was broken and this mocking was just masking the issue

app/tests/test_render_utils.py
anthonyhashemi: Spoken offline, still needs some fixing.

app/main/util/render_utils.py
anthonyhashemi: not for this pr. followup ticket created: https://national-archives.atlassian.net/browse/AYR-1809

local_services/mds_data_generator/example_files/file.pdf
anthonyhashemi, line 1: what pdf file is this. one that just says "redacted"? Just asking to make sure it is nothing with a license that isnt open or something sensitive

app/tests/test_routes.py
anthonyhashemi: I suggest we stop mocking `mock_generate_pdf_manifest` and `mock_create_presigned_url` in this test as we should have a test at this level successfully testing the endpoint works wihout mocking underlying logic. 

If we have edge cases, sure, they can be tested at a lower level.

app/tests/test_render_utils.py
anthonyhashemi: has now been fixed

local_services/mds_data_generator/example_files/file.pdf
karwan-zaizi: it is just a file I created with test Page1 in first page and Page2 in page 2, it is not licensed

app/main/routes.py
karwan-zaizi: But we are using different bucket "ACCESS_COPY_BUCKET" in order to get the file

app/tests/test_routes.py
karwan-zaizi: Removed the mock_generate_pdf_manifest, updated the test according new changes we have added

app/main/routes.py
anthonyhashemi: oh I see, ignore me 🤦

anthonyhashemi (dismissed):

anthonyhashemi:
Thanks for addressing all of the comments. I have approved but see that code cov is complaining about missing coverage:
Screenshot 2025-10-31 at 09 34 47
And the poetry.lock has a merge conflict

karwan-zaizi:
@anthonyhashemi all done now, the conflict fixed and added the image extensions to the test

anthonyhashemi (approved):

PR #1315, opened 2025-10-16
AYR-1755 - Fix bad word wrapping

karwan-zaizi:
## Changes in this PR
Fixing Bad Text wrapping in Series Table and Consignment Table

## JIRA ticket
[Ayr-1755](https://national-archives.atlassian.net/browse/AYR-1755)

### Before
Screenshot 2025-10-16 at 17 12 48
Screenshot 2025-10-16 at 17 12 59

### After
Screenshot 2025-10-16 at 17 11 19
Screenshot 2025-10-16 at 17 11 38


anthonyhashemi (dismissed):
I tried to see if there was a css native way of doing this, rather than needing all this template logic but I couldn't find one either, so approving as this does what we want. Can always improve on later if needed.

anthonyhashemi (approved):

PR #1386, opened 2025-11-07
AYR-1792 - Expand pdf access copies to more file types (mpp, wk4)

karwan-zaizi:
## Changes in this PR
Expand pdf access copies to more file types (mpp, wk4)

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1792

### After


anthonyhashemi (approved):
Amazing!

PR #1394, opened 2025-11-10
AYR-1544 - UV| increase default zoom of renders

karwan-zaizi:
## Changes in this PR
UV| increase default zoom of renders

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1544

### Before
Screenshot 2025-11-10 at 10 07 43

### After
Screenshot 2025-11-10 at 14 15 54


colinbowen (dismissed):
Looks good

app/static/init.uv.js
```diff
@@ -12,6 +12,9 @@ function initUniversalViewer() {
     config.modules.centerPanel.options = {
       usePdfJs: true,
     };
+    config.modules.openSeadragonCenterPanel.options = {
+      defaultZoomLevel: 0.0006,
```
anthonyhashemi: have you checked that this looks good with different viewports ?

Mobile etc

karwan-zaizi: Screenshot 2025-11-11 at 08 40 09

Yes, this is a mobile size

anthonyhashemi: Ok cool, can we add a visual regression test for a record at maybe 3 viewports? Or at least a test to ensure this configuration value isn't changed accidentally

Dockerfile
```diff
@@ -37,6 +37,8 @@ COPY main_app.py .flaskenv /docker_app/
 # Restore the built CSS files
 RUN cp -r /tmp/css_backup /docker_app/app/static/src/css
 
+RUN npm run test
```
colinbowen: Is there a reason to test as part of the Dockerfile being used for building? This could be better put as part of the github actions in the unit-tests or e2e tests github workflow

karwan-zaizi: No, no reason
I will add it to github actions and remove it from the Dockerfile

karwan-zaizi: @colinbowen the tests are running as part of github actions now

colinbowen (approved):

PR #1432, opened 2025-11-19
AYR-1782 - Support access copies in local dev

karwan-zaizi:
## Changes in this PR
Support access copies in local dev

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1782

### After


colinbowen:
PR looks good. I've seen in the README on line 214 which could be updated to reference the new automated file upload using mds-data-generator and the new CI docker setup. There's also quite a few files we have now so we could look at using git-lfs in future if we find the repo size grows.

karwan-zaizi:
@colinbowen yeah potentially we can 

The line 214 in README.md file is outdated. We don't add files manually to the bucket anymore. 

One thing I want to mention: the changes in these files `mds_test_file_importer.py`, `index_file_content_and_metadata_in_opensearch.py` only generate files from `example_files`. Most of the generated files can't be rendered except some like ["pdf", "png", "jpg"] with the current setup we have. 

If we want to render convertible file extensions, we need to run `access_copy_converter` to convert the new generated files to pdf, which will automatically upload the converted files to the `access-copy-bucket`. For this, some changes need to be made in `access_copy_converter`, like this: https://github.com/nationalarchives/da-ayr-beta-webapp/pull/1432/commits/b7f854fe077bd2

anthonyhashemi (commented):
Could you please add some documentation in the readme mentioning the different soruce of the documents we have added here. E.g. the panda picture etc. Just so we have an audit trail and can ensure we are not violating any licenses.

README.md
```diff
@@ -799,3 +799,22 @@ Uses [Flask Compress](https://github.com/colour-science/flask-compress) to compr
...
 
 Please see the [contribution guidelines](CONTRIBUTING.md) for how to raise a bug report or feature request.
+
+## New files
+
+	SCOT 13_G85D3R.png : Image of The National Archives's shop
+  SCOT 13_6YTFTC.jpg : Image of The National Archives's shop
+  SCOT 13_UYT6DV.xlsx : Random xlsx template
+  SCOT 13_VTC9WP.xls : Random xls template
+  SCOT 13_Z95P37.ppt :  ppt file from INT enviroment
```
anthonyhashemi: I think we need to be more specific about the original source, but let me double check with some others in the department

karwan-zaizi: I will update the document references shortly

karwan-zaizi: Updated documents references

PR #1458, opened 2025-11-25
Fix/npm build

colinbowen:
## Changes in this PR
- Removed npm from package.json due to having npm install npm creates a circular dependency on itself
https://stackoverflow.com/questions/50786998/should-i-have-npm-as-dependency-in-package-json

### After

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get install -y \
     curl \
     tesseract-ocr \
     antiword \
-    libreoffice \
+    libreoffice
 
 RUN curl -fsSL https://deb.nodesource.com/setup_25.x | bash - \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN poetry config virtualenvs.create false && \
 
 # Copy Node.js dependency files and install
 COPY package*.json /docker_app/
+RUN npm install
```
anthonyhashemi: why is this needed now? We already have npm ci below

colinbowen: This could be removed. Let me build the lock file with new deps. There was an issue previously with the lock file being out of sync https://github.com/nationalarchives/da-ayr-beta-webapp/actions/runs/19676390167/job/56358610410

PR #1470, opened 2025-12-01
Run all data_management tests in CI

anthonyhashemi:
## Changes in this PR

Run all data_management tests in CI

## JIRA ticket

https://national-archives.atlassian.net/browse/AYR-1851


anthonyhashemi:
oh actually that fix wasnt as simple, will put it back in todo, issues with relative imports etc see: https://github.com/nationalarchives/da-ayr-beta-webapp/actions/runs/19823233703/job/56790233851?pr=1470

PR #1530, opened 2025-12-15
Render improvements - UV

colinbowen:
### After


app/main/util/render_utils.py
```diff
@@ -137,64 +207,98 @@ def create_presigned_url_for_access_copy(file: File) -> str:
...
-    page_data = []
     current_app.logger.info(
-        f"Generating PDF manifest for {file_name}, file_obj: {file_obj is not None}"
+        f"Generating PDF manifest for {file_name}, record_id: {record_id}"
     )
 
-    pdf_bytes = file_obj["Body"].read()
+    # Read PDF to get page count and dimensions
+    pdf_bytes = get_pdf_from_s3(bucket, key)
+    # pdf_bytes = file_obj["Body"].read()
```
anthonyhashemi: remove?

colinbowen: Removed 👍

app/main/util/render_utils.py
```diff
@@ -10,6 +11,8 @@
 
 from app.main.db.models import File
 
+pdf_cache = {}
```
anthonyhashemi: I do not believe there is any security implication since, although the cache could mean the server holds a record from user a, when user b makes a request, this is no different to the fact that the server could make a request to that record from s3 but we don't allow that as the key determined to be fetched is based off the file in the endpoint and if the user's keycloak group doesnt suffice we block that endpoint.

However:

1. I think in practice, often the cache will be lost between request as they will be on different lambdas
2. If a pdf is cached once and never requested again it will still in memory indefinitely (not a huge issue given the fact that lambdas will eventually die anyway)
3. Similarly, since we're using a dict as the cache, we couldeventually run out of memory if it accesses enough large pdfs before the cache time runs is met

I am wondering whether this additional logic/complexity is worth it?

I think unless we can show it definitely is worth it in production, we should remove it to avoid unnecessary complexity for now

In the future if we find that we'd benefit from caching we can change the code to fetch from cloudfront or some other distributed caching

colinbowen: Removed the caching implementation for now and will look to use cloudfront instead which should be better for our architecture.

anthonyhashemi (approved):
Nice one, this is such a great improvement in performance. Eventually we might want to build out a dedicated iiif presentation service

app/main/util/render_utils.py
anthonyhashemi: okay yeah that makes sense to me

PR #1544, opened 2025-12-17
AYR-1851 - Run the access copy unit tests in ci

karwan-zaizi:
### After


.github/workflows/unit_tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         sudo apt-get install -y docker-compose
 
     - name: Run Docker Compose Tests
-      run: docker compose -f docker-compose.test.yml up --exit-code-from opensearch_indexer_tests
+      run:  docker compose -f docker-compose.test.yml up --exit-code-from access_copy_converter_tests
```
anthonyhashemi: This would stop running the `opensearch_indexer_tests`, we need to continue running them.

You could have both by eiter having 2 steps:
- 1 for opensearch indexer tests and next for access copy tests

or 

- specify both in the one command

.github/workflows/unit_tests.yml
```diff
@@ -60,7 +60,22 @@ jobs:
         sudo apt-get install -y docker-compose
 
     - name: Run Docker Compose Tests
-      run: docker compose -f docker-compose.test.yml up --exit-code-from opensearch_indexer_tests
+      run: |
+        docker compose -f docker-compose.test.yml up --abort-on-container-exit
```
anthonyhashemi: Is 



 not sufficient?
```suggestion
        docker compose -f docker-compose.test.yml up --exit-code-from opensearch_indexer_tests access_copy_converter_tests
```

karwan-zaizi: This doesn't work, I have tried before and tested it again, it fails

anthonyhashemi: I see, in that case  option 2, splitting them into 2 steps is probably the cleaner approach were we just run each container

```
- name: Run OpenSearch Indexer Tests
  run: |
docker compose -f docker-compose.test.yml run --rm opensearch_indexer_tests

- name: Run Access Copy Converter Tests
  run: |
docker compose -f docker-compose.test.yml run --rm access_copy_converter_tests
```

karwan-zaizi: This failed as well, I have updated the code which listens to both opensearch and access-copy tests

anthonyhashemi: what does it fail with? Let's pair for a few mins and see if we can get something similar working, if not this is good

karwan-zaizi: it depends on postgres service, we need to run it with postgres https://github.com/nationalarchives/da-ayr-webapp/pull/1544/files#:~:text=data_management/access_copy_converter/tests-,postgres,-%3A

Let's pair later when you have time

anthonyhashemi (approved):

PR #1594, opened 2026-01-09
AYR--1418 - TechDebt| [Console warning] Manifest: property 'start_url' ignored, should be same origin as document.

karwan-zaizi:
## Changes in this PR
Delete manifest.json file and relevant code

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1418

### After


anthonyhashemi:
We are doing this because this is not a Progressive Web app
https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps
[https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifes](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest)

anthonyhashemi (approved):

PR #1619, opened 2026-01-16
switched rds proxy auth to IAM-only

rahmedzaizi:
## Changes in this PR
switched rds proxy auth to IAM-only

## JIRA ticket
[AYR-1847](https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?jql=assignee%20%3D%2060ba6c41f8a90b006f65b5fb&selectedIssue=AYR-1847)

[AYR-1847]: https://national-archives.atlassian.net/browse/AYR-1847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -86,7 +86,7 @@ def get_connection():
     return conn
 
 
-def create_app(config_class, local_env):
+def create_app(config_class, database_uri=None, local_env=False):
     app = Flask(__name__, static_url_path="/assets")
     config = config_class()
     inspect.getmembers(config)
@@ -136,6 +136,12 @@ def create_app(config_class, local_env):
         }
 
     csp = get_csp_config(app)
+
+    # setup database uri for testing
+    if database_uri:
+        app.config["SQLALCHEMY_DATABASE_URI"] = database_uri
+
+    # Initialise app extensions
     setup_logging(app)
     s3.init_app(app)
     compress.init_app(app)
@@ -151,19 +157,21 @@ def create_app(config_class, local_env):
 
     # setup database components
     if local_env:
-        print("LOCAL---")
-        print(app.config["SQLALCHEMY_DATABASE_URI"])
         db.init_app(app)
         # create db objects for testing
         with app.app_context():
             db.create_all()
     else:
-        print("Production-------")
         app.config["SQLALCHEMY_DATABASE_URI"] = "postgresql+psycopg2://"
         app.config["SQLALCHEMY_ENGINE_OPTIONS"] = {
             "creator": get_connection,
         }
         db.init_app(app)
+
+    with app.app_context():
+        # create db objects for testing
+        if database_uri:
+            db.create_all()
 
     # Register blueprints
     from app.main import bp as main_bp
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -58,6 +61,31 @@ def format_date_iso(value):
         return value
 
 
+def get_connection():
+    """Return a new psycopg2 connection with fresh IAM token."""
+    print("\n--- New connection attempt ---")
```
anthonyhashemi, line 66: should use the logger or is it notr available here?

app/__init__.py
```diff
@@ -129,12 +150,18 @@ def get_csp_config(app):
...
-        if database_uri:
+    if database_uri:
+        app.config["SQLALCHEMY_DATABASE_URI"] = database_uri
+        db.init_app(app)
+        # create db objects for testing
+        with app.app_context():
             db.create_all()
-        else:
-            db.Model.metadata.reflect(bind=db.engine, schema="public")
+    else:
```
anthonyhashemi, line 164: so if using iam we don't provide the database_uri and this following logic is what is needed, right?

app/__init__.py
rahmedzaizi: Not Available here 
We can just move it 
Was using it to debug

app/__init__.py
```diff
@@ -129,12 +150,18 @@ def get_csp_config(app):
     WTFormsHelpers(app)
 
     # setup database components
-    with app.app_context():
-        # create db objects for testing else use existing database objects
-        if database_uri:
+    if database_uri:
+        app.config["SQLALCHEMY_DATABASE_URI"] = database_uri
+        db.init_app(app)
+        # create db objects for testing
```
anthonyhashemi, line 161: so this is just needed for testing

app/__init__.py
```diff
@@ -129,12 +150,18 @@ def get_csp_config(app):
     WTFormsHelpers(app)
 
     # setup database components
-    with app.app_context():
-        # create db objects for testing else use existing database objects
-        if database_uri:
+    if database_uri:
+        app.config["SQLALCHEMY_DATABASE_URI"] = database_uri
+        db.init_app(app)
```
anthonyhashemi, line 160: can we move this line which is both if and else block out or is it a faff due to the testing stuff below?

app/__init__.py
rahmedzaizi: yes

app/__init__.py
rahmedzaizi: yes
app.config["SQLALCHEMY_DATABASE_URI"] = "postgresql+psycopg2://"
app.config["SQLALCHEMY_ENGINE_OPTIONS"] = {
            "creator": get_connection,
        }
 These configs are used by db.init_app(app) to establish connection 
 We can also just use create_engine

app/__init__.py
rahmedzaizi: yeah 
else is not needed to be honest

anthonyhashemi (dismissed):
nice one

anthonyhashemi (approved):

PR #1632, opened 2026-01-20
Ayr 1747/test data differences

colinbowen:
## Changes in this PR
- adds refresh to keycloak login when running tests
- recreated index for opensearch using 2.15.0
- updated realm export for testing & local setup

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1747

### After


app/main/db/queries.py
```diff
@@ -247,10 +246,19 @@ def build_browse_consignment_query(
                 query = query.order_by(desc(sort_field))
             else:
                 query = query.order_by(sort_field)
+            # Use convert_to for consistent byte-order sorting across all environments
```
anthonyhashemi: Do we need these sort changesI thought this PR was just going to be changes to improve test speed around keycloak connections and pinning opensearch in e2e tests to 2.15 and recreating the opensearch index in accordance?

colinbowen: Removed and kept it to the other changes 👍

local_services/import/realm-export.json
```diff
@@ -71,23 +74,7 @@
       "fullScopeAllowed": true,
       "defaultClientScopes": [
         "group_mapper_client_scope",
-        "basic",
-        "roles",
-        "web-origins"
-      ],
-      "protocolMappers": [
```
anthonyhashemi, line 78: I guess we don't need these?

anthonyhashemi (approved):
Nice 🙏

local_services/import/realm-export.json
colinbowen: not for local testing, we're only using one realm

PR #1662, opened 2026-01-28
support additional file types for textraction

rahmedzaizi:
Extended text extraction support by converting unsupported but convertible file types into formats supported by Textract using LibreOffice before extraction

## JIRA ticket
[AYR-1887](https://national-archives.atlassian.net/browse/AYR-1887)

[AYR-1887]: https://national-archives.atlassian.net/browse/AYR-1887?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
+++ b/data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
@@ -133,7 +133,7 @@ def add_text_content(file: Dict, file_stream: bytes) -> Dict:
                 file_puid
             ]
             target_ext = SUPPORTED_TEXTRACT_PUIDS[target_puid]
-
+            logger.info(f"Converting {file_id} to {target_ext}")
             with tempfile.NamedTemporaryFile(delete=True) as src:
                 src.write(file_stream)
                 src.flush()
--- a/data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
+++ b/data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
@@ -26,7 +26,7 @@ TEXTRACT_FILE_PUIDS_CONVERT_TO_SUPPORTED_MAP = {
     "fmt/116": "fmt/20",  # .bmp to .pdf
     "x-fmt/258": "fmt/20",  # .vsd to .pdf
     "x-fmt/1510": "fmt/20",  # .vsd to .pdf
-    "x-fmt/443": "fmt/20",  # .vsd to .pdf
+    "fmt/443": "fmt/20",  # .vsd to .pdf
     "x-fmt/255": "fmt/20",  # .pub to .pdf
     "x-fmt/332": "fmt/20",  # .fm3 to .pdf
 }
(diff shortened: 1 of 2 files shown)


data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -103,63 +116,83 @@ def add_text_content(file: Dict, file_stream: bytes) -> Dict:
...
+                f"Text extraction skipped for file {file_id} due to unsupported file type: {file_puid}"
             )
-        except Exception as e:
-            logger.error(f"Text extraction failed for file {file_id}: {e}")
             file["content"] = ""
-            file["text_extraction_status"] = TextExtractionStatus.FAILED.value
+            file["text_extraction_status"] = TextExtractionStatus.SKIPPED.value
+            return file
+
+        if file_puid in TEXTRACT_FILE_PUIDS_CONVERT_TO_SUPPORTED_MAP:
```
anthonyhashemi, line 154: Since this function is getting quite large now I think it could be worth extracting out the if and else blocks into a couple of functions like:

```
def convert_file_and_extract_text_in_memory(.....):
            target_puid = TEXTRACT_FILE_PUIDS_CONVERT_TO_SUPPORTED_MAP[
                file_puid
            ]
            target_ext = SUPPORTED_TEXTRACT_PUIDS[target_puid]
            logger.info(f"Converting {file_id} to {target_ext}")
            with tempfile.NamedTemporaryFile(delete=True) as src:
                src.write(file_stream)
                src.flush()

                converted_path = convert_file_with_libreoffice(
                    src.name, target_ext
                )
                
                return extract_text(converted_path, target_puid)
```

and

```
def extract_text_in_memory(...):
            ext = SUPPORTED_TEXTRACT_PUIDS[file_puid]
            with tempfile.NamedTemporaryFile(
                suffix=f".{ext}", delete=True
            ) as temp:
                temp.write(file_stream)
                temp.flush()
                return extract_text(temp.name, file_puid)

```

Potentially we can refactor a bit further to reduce duplication of the context manager and flushing etc but not a huge deal

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
```diff
@@ -16,13 +16,27 @@
 logger.setLevel(logging.INFO)
 libreoffice_lock = threading.Lock()
 
+TEXTRACT_FILE_PUIDS_CONVERT_TO_SUPPORTED_MAP = {
```
anthonyhashemi, line 19: we will add some e2e tests for these once is in https://github.com/nationalarchives/da-ayr-webapp/pull/1432 and we have some more test files but have been manually tested for now

anthonyhashemi (commented):
Generally looks good, can we add some unit tests to cover the new code paths, then will be happy to approve

data_management/opensearch_indexer/opensearch_indexer/text_extraction.py
rahmedzaizi: done

anthonyhashemi (approved):
nice one

PR #1668, opened 2026-01-29
AYR-1844 - Handle validation errors gracefully

karwan-zaizi:
## JIRA ticket
[AYR-1844](https://national-archives.atlassian.net/browse/AYR-1844)

### After

[AYR-1844]: https://national-archives.atlassian.net/browse/AYR-1844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -13,7 +13,7 @@ from flask import (
     url_for,
 )
 from sqlalchemy import func
-from werkzeug.exceptions import HTTPException
+from werkzeug.exceptions import HTTPException, NotFound
 
 from app.main import bp
 from app.main.authorize.access_token_sign_in_required import (
@@ -81,7 +81,6 @@ from app.main.util.search_utils import (
     execute_search,
     extract_search_terms,
     get_open_search_fields_to_search_on_and_sorting,
-    get_pagination_info,
     post_process_opensearch_results,
     setup_opensearch,
 )
@@ -169,6 +168,26 @@ def accessibility():
     return render_template("accessibility.html")
 
 
+def get_page_and_per_page(validated_data):
+    page = validated_data.get("page")
+    per_page = validated_data.get("per_page")
+    if not per_page:
+        per_page = int(current_app.config["DEFAULT_PAGE_SIZE"])
+    return page, per_page
+
+
+def redirect_if_page_invalid(requested_page, default_page, endpoint, **kwargs):
+    """
+    Redirects to the default page if the requested page is invalid (e.g., out of range).
+    Returns a redirect response if a redirect is needed, otherwise None.
+    """
+    if requested_page != default_page:
+        # Copy current query parameters and set page to default_page
+        args = request.args.to_dict()
+        args["page"] = default_page
+        return redirect(url_for(endpoint, **kwargs, **args))
+
+
 @bp.route("/browse", methods=["GET"])
 @access_token_sign_in_required
 @log_page_view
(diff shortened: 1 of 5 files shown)


app/main/routes.py
```diff
@@ -169,13 +168,34 @@ def accessibility():
     return render_template("accessibility.html")
 
 
+def get_page_and_per_page(validated_data):
```
anthonyhashemi: these 2 functions can probably live in a dedicated file, and leave this file for routes only

app/main/routes.py
```diff
@@ -694,9 +725,18 @@ def search_transferring_body(_id: uuid.UUID):
             search_results["hits"]["hits"], sort
         )
 
-        total_records, pagination = get_pagination_info(
-            search_results, page, per_page
+        total_records = (
```
anthonyhashemi, line 735: I'm trying to remember why this route is different to the others... why is there not a paginated query we call before here to get the results like in the other routes

app/main/routes.py
karwan-zaizi: Moved these functions to a different file

app/main/routes.py
karwan-zaizi: I believe it is because of search query, it is different from filtering and sorting query

anthonyhashemi: is it something we could bring inline simply enough as we are here, for consistency?

karwan-zaizi: Not sure what you mean, I believe this is as close as it gets to the other places we handle other page_count and pagination

anthonyhashemi: as in, could we take a brief look to see if we could refactor the query in this endpoint so it is similar to the other endpoints, that way we can treat pagination consistently across all routes?

app/main/routes.py
```diff
@@ -599,13 +629,15 @@ def search_results_summary():
             "aggregate_by_transferring_body"
         ]["buckets"]
 
-        total_records = 0
-        for bucket in results:
-            total_records += bucket["doc_count"]
+        total_records = sum(bucket["doc_count"] for bucket in results)
```
anthonyhashemi, line 630: same for this one.

I didn't see this initially; so it seems the 2 search routes are the distinct ones, which I guess makes sense as the rest are browse. It's been al ong while since we wrote the queries, I just was wondering if there was a way to make the search routes use 

`query.paginate` so we could try except them like you have done in the other routes

app/main/util/search_utils.py
```diff
@@ -240,12 +240,16 @@ def execute_search(open_search, dsl_query, page, per_page):
...
     from_ = per_page * (page - 1)
     try:
-        return open_search.search(
+        response = open_search.search(
             body=dsl_query,
             from_=from_,
             size=per_page,
             timeout=current_app.config["OPEN_SEARCH_TIMEOUT"],
         )
+        total_hits = response.get("hits", {}).get("total", {}).get("value", 0)
```
anthonyhashemi, line 249: the try except only applies to the `response` line. Always good to keep the scope of try excepts narrow for ease of debugging

app/main/util/search_utils.py
```diff
@@ -240,12 +240,16 @@ def execute_search(open_search, dsl_query, page, per_page):
...
+        if page < 1 or (page > 1 and from_ >= total_hits):
+            abort(404)
```
anthonyhashemi, line 251: why are we aborting rather than redirecting to the first page like the other routes?

anthonyhashemi (approved):
Looks good. As we continue to refactor these routes, we will probably end up extracting out more similar blocks of code too

app/main/util/search_utils.py
karwan-zaizi: To make sure the page exists

PR #1922, opened 2026-03-30
use add operator in fuzzy match

colinbowen:
## Changes in this PR
- Use AND inside fuzzy match query to match on all terms in search

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1904

### After


anthonyhashemi:
Can we get an e2e test to show that this addresses the issue outlined in the ticket

anthonyhashemi (approved):

PR #1973, opened 2026-04-13
add mds tests as part of ci run

colinbowen:
## Changes in this PR
- Adds a step in e2e-tests to run a set of files and index them in opensearch
- Verifies files have been imported

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1845

### After

### DIFF
--- a/local_services/mds_data_generator/test_mds_import.py
+++ b/local_services/mds_data_generator/test_mds_import.py
@@ -159,7 +159,7 @@ def run():
         assert result[
             "found"
         ], f"File ID {file_id} not found in OpenSearch index"
-        print(f"  OpenSearch OK: {file_id}")
+        print(f"OpenSearch OK: {file_id}")
 
     print("\nAll assertions passed.")
     return 0
--- a/local_services/mds_data_generator/test_mds_import.py
+++ b/local_services/mds_data_generator/test_mds_import.py
@@ -1,16 +1,23 @@
 """
 Integration test: runs mds_test_file_importer and asserts the expected records
-exist in the database.
+exist in the database and OpenSearch.
 
 Usage:
     python local_services/mds_data_generator/test_mds_import.py
 
 """
 
+import os
 import sys
 
 from dotenv import load_dotenv
-from mds_test_file_importer import create_test_filepaths, process_files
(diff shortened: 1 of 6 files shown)


local_services/mds_data_generator/test_mds_import.py
```diff
@@ -0,0 +1,169 @@
...
+                not missing_keys
+            ), f"File {f.FileName} is missing metadata keys: {missing_keys}"
+
+            print(f"  DB OK: {f.FileName} (ID: {f.FileId})")
+
+        file_ids = [f.FileId for f in files]
+
+    # 5. Index files in OpenSearch
+    print("\nIndexing files in OpenSearch ...")
+    index_in_opensearch(file_paths)
```
anthonyhashemi: I thought we would do this more as a e2e test where we run

`python mds_test_file_importer.py --num-pdf 2 --num-png 1`

this test doesn't currently test that the script itself runs successfuly, it just test some of its individual components, right?

colinbowen: Yes that's correct, I could revise the test to run the script first as-is and then continue with the checks I've added.

anthonyhashemi: Yeah I'd suggest doing that, unless there was a particular reason we wanted to avoid that? My main concern with the test as it currently is, is that it essentially duplicates what we do in the main source code but with some tweaks, and it doesn't feel like we gain any/much speed benefit or simplicity/debugability improvement?

colinbowen: Yes I agree, it would also be simpler to test it directly and reduces the code we need for this

colinbowen: Changed this to use 
```
result = subprocess.run(
        [sys.executable, str(IMPORTER_SCRIPT)],
        capture_output=True,
        text=True,
        cwd=PROJECT_ROOT,
    )
```

anthonyhashemi (approved):

local_services/mds_data_generator/test_mds_import.py
anthonyhashemi: nice one!

PR #2002, opened 2026-04-23
AYR-1943 - Automate performance accessibility tests

karwan-zaizi:
## Changes in this PR
Automate performance accessibility tests

## JIRA ticket
[AYR-1943](https://national-archives.atlassian.net/browse/AYR-1943)

### After

[AYR-1943]: https://national-archives.atlassian.net/browse/AYR-1943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/accessibility_tests.yml
```diff
@@ -0,0 +1,86 @@
...
+
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v3.1.1
+
+    - name: Set up Docker Compose
+      uses: docker/setup-compose-action@8cccb8c14b6500aaffebff1aa49c502c34d2e5e6 # v2.1.0
+      with:
+        version: v2.40.3
+
+    - name: Pre-build webapp image with cache
```
anthonyhashemi: we should try to reuse the docker image we build for the other ci so we don't waste time rebuilding the image multiple times. There are various ways to do that...

see what is done in FCL for example:

https://github.com/nationalarchives/ds-caselaw-public-ui/blob/fe77da9b2e00eae81308d538be630e1db18f074f/.github/workflows/ci.yml#L127
 https://github.com/nationalarchives/ds-caselaw-public-ui/blob/fe77da9b2e00eae81308d538be630e1db18f074f/.github/workflows/ci.yml#L183-L184

but there may be a better way... worth a brief look into best practies

local_services/.docker.env
```diff
@@ -36,7 +36,7 @@ DEFAULT_PAGE_SIZE=10
 DEFAULT_DATE_FORMAT=DD/MM/YYYY
 RECORD_BUCKET_NAME=test-record-download
 ACCESS_COPY_BUCKET=test-access-copy
-PERF_TEST=False
+PERF_TEST=True
```
anthonyhashemi: I really think we should reconsider this `PERF_TEST` @colinbowen as it is a huge security risk if it is ever set to False in production

karwan-zaizi: This has been reverted to False.

.github/workflows/accessibility_tests.yml
karwan-zaizi: This has been implemented.

local_services/.docker.env
anthonyhashemi: but we are still using PERF_TEST bypass when running the performance tests right? I see Colin made a comment to address that in a separate ticket

.github/workflows/accessibility_tests.yml
anthonyhashemi: thank you for that 🙏

anthonyhashemi (approved):
Thanks for demoing this earlier -great start and we can iterate on this to swtup some benchmark thresholds in a followup

PR #2072, opened 2026-05-13
Ayr 1959/pdfjs render

colinbowen:
## Changes in this PR
New routes (routes.py)
- GET /record/<id>/pdf, streams the raw PDF from S3
- Uses the access copy bucket for convertible PUIDs.
- GET /record/<id>/search?q=,  IIIF Content Search API v1 endpoint. Returns an empty annotation list for blank queries; calls search_within_pdf for real queries.

New utility function (render_utils.py)
- search_within_pdf(), fetches a PDF from S3, uses PyMuPDF to find text matches on each page, returns a IIIF sc:AnnotationList
   
PDF manifest changes (render_utils.py)
- generate_pdf_manifest() now includes a service block (IIIF ContentSearch API v1 pointing at the search route) and a rendering block (link to the raw PDF route). This is what enables UV's "search within" input to appear.

New schema (schemas.py)
- SearchWithinRequestSchema
  
UV config (init.uv.js)
- usePdfJs: true (was false), enables PDF.js rendering in the viewer
- fullscreenEnabled: true (was false)
- footerPanelEnabled: true (was false), shows the footer bar with search results navigation
- downloadEnabled: false and shareEnabled: false set on both footerPanel and pdfHeaderPanel
 
CSS (_record.scss)
- .uv .footerPanel .btn.share and .uv .footerPanel .btn.download hidden
  
Env template (.env.env_var.template)
- Added ACCESS_COPY_BUCKET, OPEN_SEARCH_USE_SSL, OPEN_SEARCH_VERIFY_CERTS
  

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1959

### Before
Screenshot 2026-05-15 at 16 16 52

### After
Screenshot 2026-05-15 at 16 02 43


app/static/init.uv.test.js
```diff
@@ -110,46 +110,42 @@ describe("tests for init.uv.js", () => {
   });
 
   it("should configure UV viewer with correct options", () => {
```
karwan-zaizi, line 112: Have we configured init.uv.test.js in CI? this test is failing and hasn't been picked up by CI

`expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

      Object {
        "options": Object {
          "footerPanelEnabled": true,
          "headerPanelEnabled": true,
          "leftPanelEnabled": true,
    +     "preserveViewport": true,
          "rightPanelEnabled": false,
        },
      },`

karwan-zaizi (dismissed):
All looks good

karwan-zaizi (approved):

PR #2102, opened 2026-05-21
Fix 405 MethodNotAllowed causing 500 due to missing template

rahmedzaizi:
Added TemplateNotFound fallback in http_exception error handler to return
JSON response when no template exists for the error code. Prevents false
500 errors and noisy Slack alerts

### DIFF
--- a/app/main/routes.py
+++ b/app/main/routes.py
@@ -6,12 +6,14 @@ from flask import (
     Response,
     abort,
     current_app,
+    jsonify,
     redirect,
     render_template,
     request,
     session,
     url_for,
 )
+from jinja2.exceptions import TemplateNotFound
 from sqlalchemy import func
 from werkzeug.exceptions import HTTPException, NotFound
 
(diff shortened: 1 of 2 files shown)


app/main/routes.py
```diff
@@ -1031,4 +1033,17 @@ def terms_of_use():
 
 @bp.app_errorhandler(HTTPException)
 def http_exception(error):
-    return render_template(f"{error.code}.html"), error.code
+    try:
```
anthonyhashemi, line 1036: perfect 👌

app/main/routes.py
```diff
@@ -1031,4 +1033,17 @@ def terms_of_use():
...
+        return render_template(f"{error.code}.html"), error.code
+    except TemplateNotFound:
```
anthonyhashemi, line 1038: can you add a test for this then will approve

rahmedzaizi: Added test @anthonyhashemi

anthonyhashemi (approved):

PR #2109, opened 2026-05-22
[AYR-1969] update to govukPhaseBanner

donna-tna:
## Changes in this PR
Use the standard govukPhaseBanner macro. So that phase banners are consistent with GDS guidance and easier to maintain.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1969
### Before
before

### After
after


anthonyhashemi (commented):
I guess there must be a very subtle visual difference with the phase banner given all the snapshots needing to be recreated? I can't see it though!

app/templates/base.html
```diff
@@ -32,25 +32,22 @@
...
-            <p class="govuk-phase-banner__content">
-                <strong class="govuk-tag govuk-phase-banner__content__tag">Beta</strong>
-                <span class="govuk-phase-banner__text">
-                    Help us to improve this service by completing our <a class="govuk-link"
-    target="blank"
-    href="https://www.smartsurvey.co.uk/s/ayr-feedback/">feedback
-                survey (opens in a new tab)</a>
-            </span>
-        </p>
+    <div class="govuk-phase-banner govuk-!-padding-top-1 banner__container">
```
colinbowen:
```suggestion
    <div class="govuk-!-padding-top-1 banner__container">
```

donna-tna: @colinbowen the heat is hitting my brain forgot to remove that too -updated thanks

colinbowen (approved):

PR #2115, opened 2026-05-27
[AYR-1970] update to use govukSummaryList

donna-tna:
### Description

The ticket requirements were to replace custom summary list markup in **`record-details.html`** and use `govukSummaryList` rows to render existing content. Preserve current order, labels, and values. Therefore no custom summary list HTML or inline summary logic remains in the template.

In this PR, I have deleted the **`summary_list_items.html`** file entirely. The loop taken from that file has been integrated directly into **`record-details.html`** to prepare the macro rows, successfully consolidating the template logic into a single file.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-1970
### After


app/templates/main/record-details.html
```diff
@@ -1,44 +1,93 @@
...
+                                                            "value": {
+                                                                "html": value_html,
+                                                                "classes": "govuk-summary-list__value--record"
+                                                            }
+                                                        }) %}
+                        {% endif %}
+                    {% endif %}
+                {% endfor %}
+                {{ govukSummaryList({
+                "classes": "govuk-summary-list govuk-summary-list--record",
```
colinbowen:
```suggestion
                "classes": "govuk-summary-list--record",
```

donna-tna: t

> 

thanks updated @colinbowen

colinbowen (approved):

PR #2143, opened 2026-06-09
test supress warning

colinbowen:
## Changes in this PR
Change MuPDF logs to use flask debug logs when a malformed PDF is detected. MuPDF logs a non blocking structure tree warning and continues. A similar message occurs for structure tree validation warnings that MuPDF emits for non conforming PDFs.

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-2042

## Screenshots of UI changes
N/A

### After

### DIFF
--- a/app/main/util/render_utils.py
+++ b/app/main/util/render_utils.py
@@ -9,6 +9,8 @@ from botocore.exceptions import ClientError
 from flask import Response, current_app, jsonify, url_for
 from PIL import Image
 
+from app.main.db.models import File
+
 
 @contextlib.contextmanager
 def _open_pdf(pdf_bytes: bytes):
@@ -29,9 +31,6 @@ def _open_pdf(pdf_bytes: bytes):
                 current_app.logger.debug("MuPDF: %s", messages)
             except RuntimeError:
                 pass  # current_app raises RuntimeError outside app context, debug logging is not critical
-
-
-from app.main.db.models import File
 
 
 def generate_breadcrumb_values(file):
--- a/app/main/util/render_utils.py
+++ b/app/main/util/render_utils.py
@@ -30,6 +30,7 @@ def _open_pdf(pdf_bytes: bytes):
             except RuntimeError:
                 pass
 
+
 from app.main.db.models import File
 
 
--- a/app/main/util/render_utils.py
+++ b/app/main/util/render_utils.py
@@ -1,6 +1,7 @@
 import base64
 import contextlib
 import io
+import logging
 from typing import List
 
 import boto3
@@ -11,6 +12,8 @@ from PIL import Image
 
 from app.main.db.models import File
 
+logger = logging.getLogger(__name__)
+
 
 @contextlib.contextmanager
 def _open_pdf(pdf_bytes: bytes):
@@ -18,16 +21,18 @@ def _open_pdf(pdf_bytes: bytes):
     pymupdf.TOOLS.mupdf_display_errors(False)
     pymupdf.TOOLS.mupdf_display_warnings(False)
     try:
-        with pymupdf.open("pdf", io.BytesIO(pdf_bytes)) as doc:
+        try:
+            doc = pymupdf.open("pdf", io.BytesIO(pdf_bytes))
+        except Exception as e:
+            raise ValueError(f"Failed to open PDF: {e}") from e
+        with doc:
             yield doc
-    except Exception as e:
-        raise ValueError(f"Failed to open PDF: {e}") from e
     finally:
         messages = pymupdf.TOOLS.mupdf_warnings(reset=True)
         pymupdf.TOOLS.mupdf_display_errors(True)
         pymupdf.TOOLS.mupdf_display_warnings(True)
         if messages:
-            current_app.logger.debug("MuPDF: %s", messages)
+            logger.debug("MuPDF: %s", messages)
 
 
 def generate_breadcrumb_values(file):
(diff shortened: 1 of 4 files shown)


karwan-zaizi (dismissed):
Looks good

app/main/util/render_utils.py
```diff
@@ -11,6 +12,27 @@
...
+    except Exception as e:
+        raise ValueError(f"Failed to open PDF: {e}") from e
+    finally:
+        messages = pymupdf.TOOLS.mupdf_warnings(reset=True)
+        pymupdf.TOOLS.mupdf_display_errors(True)
+        pymupdf.TOOLS.mupdf_display_warnings(True)
+        if messages:
+            try:
+                current_app.logger.debug("MuPDF: %s", messages)
+            except RuntimeError:
```
anthonyhashemi: I'd like to understand why `current_app.logger.debug` would raise a RuntimeError. If we catch an error, we should do something with it, at least adding a log statement - obviously here we could get into a cyclic issue but that leads to think this might not be the right solution.

app/main/util/render_utils.py
```diff
@@ -11,6 +12,27 @@
...
 
+@contextlib.contextmanager
+def _open_pdf(pdf_bytes: bytes):
+    """Open a PDF with MuPDF stderr suppressed; log any collected messages at DEBUG."""
+    pymupdf.TOOLS.mupdf_display_errors(False)
+    pymupdf.TOOLS.mupdf_display_warnings(False)
+    try:
+        with pymupdf.open("pdf", io.BytesIO(pdf_bytes)) as doc:
+            yield doc
+    except Exception as e:
```
anthonyhashemi: do we know what exceptions we are catching - it would be better to catch specific exceptions  so we avoid swallowing unexpected bugs

app/main/util/render_utils.py
```diff
@@ -11,6 +12,27 @@
 from app.main.db.models import File
 
 
+@contextlib.contextmanager
+def _open_pdf(pdf_bytes: bytes):
+    """Open a PDF with MuPDF stderr suppressed; log any collected messages at DEBUG."""
+    pymupdf.TOOLS.mupdf_display_errors(False)
+    pymupdf.TOOLS.mupdf_display_warnings(False)
```
anthonyhashemi, line 22: what is the purpose of these TOOLs.mupdf... settings? If we are catching this specific `No common ancestor`  exception in this specific scenario, shouldn't that be sufficient?

app/main/util/render_utils.py
anthonyhashemi: we want to catch any exception when opening the doc, to then re raise it as a ValueError so that we have better error message that is clearly about failing to open a pdf:

`Failed to open PDF: MuPDF error: format error: No common ancestor in structure tree`

My only concern is that ValueError doens't really feel appropriate - we may want to add our own `PDFOpenError` for example

```
class PDFOpenException(Exception):
     pass
```

karwan-zaizi (dismissed):

app/main/util/render_utils.py
colinbowen: I've added a PDFOpenError which originates from https://pymupdf.readthedocs.io/en/latest/document.html (FileDataError).

app/main/util/render_utils.py
colinbowen: The No common ancestor isn't an actual exception from mupdf, it's just a diagnostic message mupdf prints to the logs during page rendering. The TOOLS.mupdf_display_* toggles are just a way to stop that text from showing in the logs.

app/main/util/render_utils.py
colinbowen: I've switched this to use a standard logger that raises the new PDFOpenError. Also removed the try except here.

app/main/util/render_utils.py
```diff
@@ -23,8 +27,8 @@ def _open_pdf(pdf_bytes: bytes):
     try:
         try:
             doc = pymupdf.open("pdf", io.BytesIO(pdf_bytes))
-        except Exception as e:
-            raise ValueError(f"Failed to open PDF: {e}") from e
+        except pymupdf.FileDataError as e:
```
anthonyhashemi: This is a much tighter exception handling, which would make debugging much easier. However, to clarify, we have not seen this exception raised yet and don't expect this to happen in any supported scenario.

I try to think of how we can keep things as simple as possible for maintainability and debugability, and so I'm thinking about whether we should omit this. Currently, all we do with `PDFOpenError` is catch it in the calling function and then abort 500 anyway, which is identical to what happens if pymupdf.FileDataError bubbles up uncaught. The only difference is the specific `PDFOpenError` custom exception.

Basically, this seems to be preempting a potential future need, e.g., monitoring PDF opening errors or changing the handling to a user-facing error, even if we change the implementation of the PDF opening logic later.

app/main/util/render_utils.py
anthonyhashemi: Right! I am still confused that we weren't able to recreate the error alarm we thought was associated with this message, even though we could recreate the error logs.

Suppressing the stderr to then log them at debug level does make sense, though, so we don't disturb our cloudwatch alarms that could potentially look for ERROR logs

app/main/util/render_utils.py
colinbowen: Good point, I've limited this change to just making sure the debug logging is used for now.

anthonyhashemi (approved):

PR #2150, opened 2026-06-10
rename

colinbowen:
## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-2053

### Before
Screenshot 2026-06-10 at 14 23 03

### After
Screenshot 2026-06-10 at 14 08 40


anthonyhashemi (approved):
thanks for fixing, looks good 👌

PR #2153, opened 2026-06-11
Fix webkit fail login in e2e tests

karwan-zaizi:
### After


e2e_tests/conftest.py
```diff
@@ -90,19 +91,66 @@ def create_user_page(
...
-        if page.locator("label:has-text('Email address')").count() > 0:
-            page.get_by_label("Email address").first.fill(username)
-        elif page.locator("label:has-text('Email')").count() > 0:
-            page.get_by_label("Email").first.fill(username)
-        else:
-            page.get_by_label("Username or email").first.fill(username)
-        page.get_by_role("textbox", name="Password").fill(password)
-        page.get_by_role("button", name="Sign in").click()
+
+        # WebKit can occasionally fail first sign-in submit in local Keycloak;
```
anthonyhashemi: Do we understand why this is occasionally failing? Also I wonder if playwright `page` has its own retry mechanism? If we can avoid having to write our own custom retry mechanism, that keeps things simpler

But if not, this is a pragmatic fix to get things unblocked 👌

karwan-zaizi: Good call — I’ve simplified this to use Playwright’s built-in web-first waiting/assertion retries instead of the custom retry loop. I kept the WebKit-specific browser-context fix in place, since that appears to be the key stabiliser for this flow.

anthonyhashemi (dismissed):

anthonyhashemi (approved):

PR #2176, opened 2026-06-17
[AYR-2025] update custom link to align with GDS hover in govuk-table

donna-tna:
## Changes in this PR
Add  a hover state/color to align with GDS in **govuk-table**
## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-2025
### Before
Before

### After
After

### DIFF
--- a/app/static/src/scss/includes/_overrides.scss
+++ b/app/static/src/scss/includes/_overrides.scss
@@ -1,3 +1,5 @@
+@use "variables" as *;
+
 legend {
   padding-inline: 0;
 }
@@ -26,5 +28,14 @@ a {
     &--search {
       margin-bottom: 0;
     }
+  }
+}
+
+.govuk-table a,
+.govuk-table__cell a {
+  &:hover,
+  &:visited:hover {
+    color: $colour-link-hover;
+    text-decoration-thickness: 3px;
   }
 }
--- a/app/static/src/scss/includes/_variables.scss
+++ b/app/static/src/scss/includes/_variables.scss
@@ -7,7 +7,7 @@ $breakpoint-desktop: 1056px;
 // colour__component__type
 $colour-link-normal: #0b0c0c;
 $colour-link-default: #1d70b8;
-$colour-link-hover: #0f385c;
+$colour-link-hover: govuk-functional-colour("link-hover");
 $colour-link-visited: #4c2c92;
 $colour-border-normal: #b1b4b6;
 $colour-container-normal: #f3f2f1;


app/static/src/scss/includes/_overrides.scss
```diff
@@ -28,3 +30,12 @@ a {
     }
   }
 }
+
+.govuk-table a,
+.govuk-table__cell a {
```
anthonyhashemi, line 35: I'm wondering why we even need any of this. Isn't this just the default browser hover link? @donna-tna @Terry-Price 

Probably we are overriding the default hover colour somewhere already since we previously already had a $colour-link-hover defined which is different to the default?

anthonyhashemi: yep looks like it. See: https://github.com/nationalarchives/da-ayr-webapp/blob/8b5ac6753a2e769e368710d0554e0731790fe2c2/app/static/src/scss/includes/_accessibility.scss#L47

Found this by searching for `hover` in the repo: https://github.com/search?q=repo%3Anationalarchives%2Fda-ayr-webapp+colour-link-hover&type=code


So I suggest we just:
- remove the hover bit in the link above
- remove the changes in this PR
- remove the $colour-link-hover variable alltogether


I may be wrong, but I _think_ that should work.

See: https://design-system.service.gov.uk/styles/links/#links-without-a-visited-state for more info on how links and different states are dealt with in GDS

anthonyhashemi: After looking into this https://brand.design-system.service.gov.uk/colour/web/, I obviously haven't dealt with hover links in a long time - there is no default so we do need it defined, so ignore my previous suggestion!

We do have hover link colour behaviour defined in two places though, so its worth reviewing why that's needed. The one in this PR is focussed on table cells, and the one in https://github.com/nationalarchives/da-ayr-webapp/blob/8b5ac6753a2e769e368710d0554e0731790fe2c2/app/static/src/scss/includes/_accessibility.scss#L47 seems general.

donna-tna: @anthonyhashemi the links to add hover in the **govuk-table** which must be specified seperately so has been updated as per the card the other hover is there for the general outside of the table **.govuk-link** we use around the site has a few modefications from line 24 - 50. 

These are in place as AYR is not keeping the exact states the **govuk-link** has for example we do not have the focus yellow state on the links around the site only in the footer which pulls directly from GDS **govuk-link**. So this other hover must stay in the code as it has been so far no removal of it.

anthonyhashemi: Oh I see the difference now, thanks for explaining.

Suggestion: since we're here, it is a good opportunity for us to use the colours defined by the npm module govuk-frontend (which is already installed) as described in https://design-system.service.gov.uk/styles/colour/ so that we don't need to maintain our own colours - if GDS changes things in the future, we will get it for free when we upgrade the package

e.g. `$colour-link-hover: govuk-functional-colour("link-hover")`

and whilst we're at it, it should be trivial to do the same for most of the rest of the colours in `app/static/src/scss/includes/_variables.scss`.

In terms of the duplicate `color: $colour-link-hover;`, since it's only in 2 places it probably doesn't benefit from being pulled out into a https://sass-guidelin.es/#mixins

donna-tna: @anthonyhashemi 
Screenshot 2026-06-19 at 10 42 55
if this PR can go in as it is now then we van leave everything how it is in variables for now and we can create another card and liase with @Terry-Price  about updating variable as it breaks overides the site text see above when changing variables.    This can be investigated separately as we are customising the text to not have visited or focus states

anthonyhashemi: Okay, yeah, let's leave anything that causes any visual change for now.

The screenshot shows you'be set govuk-functional-colour("brand") inside govuk-table__cell. It should be govuk-functional-colour("link-hover")

I assume doing this  for `link-hover` should be fine since we verified the hex colour value we have set in variables.scss is the same as that defined in GDS as per https://design-system.service.gov.uk/styles/colour/

anthonyhashemi (dismissed):
Looks good - but we should align variables with GDS going forward

anthonyhashemi (approved):
Nice, we can now use this as an example going forward so that we use as much as possible from the GDS node module for styling/colours/fonts wherever possible!

PR #2245, opened 2026-07-08
AYR-2069 - User can access the new browse page and see all their available data

karwan-zaizi:
## Changes in this PR
User can access the new browse page and see all their available data in the new table design
## JIRA ticket
[AYR-2069](https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-2069)

### Before
before-browse

### After
Screenshot 2026-07-17 at 11 14 42

[AYR-2069]: https://national-archives.atlassian.net/browse/AYR-2069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Terry-Price (changes requested):
Hi @karwan-zaizi 
Thanks for this can you make the following updates"

- Take out the endash for Open records.
- Change the record opening label to 'Opening date' 
- Take out the dividers except the top and bottom ones.
- Make the first column wider and push the others to the right and close them up if required. We need the first column to be as wide as possible to avoid wrapping as much as we can.

I will pair with you on Wednesday.

donna-tna (dismissed):

Terry-Price (dismissed):
Thanks @karwan-zaizi – this desktop looks good,

app/static/src/scss/includes/_variables.scss
```diff
@@ -9,6 +9,7 @@ $colour-link-normal: #0b0c0c;
 $colour-link-default: #1d70b8;
 $colour-link-hover: govuk-functional-colour("link-hover");
 $colour-link-visited: #4c2c92;
+$secondary-text-colour: #484949;
```
anthonyhashemi: We agreed to use govuk-functional-color or gov-uk-color from now on like we do on line 10

See: https://design-system.service.gov.uk/styles/colour/

Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     tesseract-ocr=5.5.0-1+b1 \
     antiword=0.37-17 \
     unrtf \
-    libreoffice=4:25.2.3-2+deb13u5 \
+    libreoffice \
```
anthonyhashemi, line 13: why did we unpin? Is renovate not updating these?

app/tests/test_schemas.py
```diff
@@ -257,6 +258,44 @@ def test_missing_consignment_id(self):
         assert "_id" in exc_info.value.messages
 
 
+class TestBrowseRecordsRequestSchema:
```
anthonyhashemi, line 261: nice

app/tests/test_build_browse_records_query.py
```diff
@@ -0,0 +1,156 @@
+from flask.testing import FlaskClient
+
+from app.main.db.queries import build_browse_records_query
+
+
+class TestBrowseRecords:
```
anthonyhashemi, line 6: nice!

app/tests/test_browse_records.py
```diff
@@ -0,0 +1,158 @@
...
+        "Status",
+        "Opening date",
+    ]
+
+    assert [
+        header.text.replace("\n", " ").strip(" ") for header in headers
+    ] == expected_row
+
+
+class TestBrowseRecords:
```
anthonyhashemi, line 26: a visual regression test for this new page would still be good, it might even mean some of these are redundant

app/templates/main/browse.html
```diff
@@ -6,15 +6,19 @@
...
             <!-- TABLE -->
             <div class="govuk-grid-column-full govuk-grid-column-full__page_container">
                 <div class="browse-details">
                     <h2 class="govuk-heading-l browse__records-found__text"
                         id="browse-records"
                         aria-live="polite">
                         {% if num_records_found > 0 %}
-                            Browse records {{ num_records_found | format_number_with_commas }}
+                            {% if browse_type == "records" %}
+                                {{ num_records_found | format_number_with_commas }} records
```
anthonyhashemi: does this complete https://national-archives.atlassian.net/browse/AYR-2072 as well then or not quite?

app/templates/main/browse.html
```diff
@@ -6,15 +6,19 @@
     <div class="govuk-grid-row browse__page desktop mobile">
         {{ super() }}
         {% include "top-search.html" %}
-        {% if results %}
+        {% if results or browse_type == "records" %}
```
anthonyhashemi, line 9: So this whole conditional will be removed once we get rid of the other browse pages

app/templates/main/browse.html
```diff
@@ -6,15 +6,19 @@
...
             <!-- TABLE -->
             <div class="govuk-grid-column-full govuk-grid-column-full__page_container">
                 <div class="browse-details">
                     <h2 class="govuk-heading-l browse__records-found__text"
                         id="browse-records"
                         aria-live="polite">
                         {% if num_records_found > 0 %}
-                            Browse records {{ num_records_found | format_number_with_commas }}
+                            {% if browse_type == "records" %}
```
anthonyhashemi: and this one

Dockerfile
karwan-zaizi: The local setup and ci was failing, this update it on main as well, forgot to remove it in this pr

app/static/src/scss/includes/_variables.scss
karwan-zaizi: Totally forgot to update it, I will update it shortly

app/templates/main/browse.html
karwan-zaizi: Yes, it is just to render the page correctly for now, when everything else is done, this can be removed

app/templates/main/browse.html
karwan-zaizi: same here

app/templates/main/browse.html
karwan-zaizi: same here

Dockerfile
anthonyhashemi: do we understand why? The idea we pin is so that we have reproducible builds and know what version we are using for security reasons

app/templates/main/browse.html
anthonyhashemi: okay I was just trying to understand whether this count display it meets the requirements of that ticket since record count was out of scope for this ticket

Dockerfile
colinbowen: This was something I noticed which caused our ci to fail: https://github.com/nationalarchives/da-ayr-webapp/pull/2264. One thing we could do is pin using a wildcard e.g. libreoffice=4:25.2.3-2* \

https://askubuntu.com/questions/1218814/problem-using-wildcard-with-apt

app/templates/main/browse-records.html
```diff
@@ -0,0 +1,82 @@
+<div class="browse-layout records-page-wrapper">
+    {% if num_records_found > 0 %}
+        {% set start_index = ((current_page - 1) * per_page) + 1 %}
+        {% set end_index = [current_page * per_page, num_records_found] | min %}
+        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds browse-grid--mobile-table">
+            <table class="govuk-table browse-grid__table"
```
anthonyhashemi: Image

This table is built around a single “record mega-row” spread across multiple `<tr>`s, with repeated colspan=4 coupling the layout to the current 4-column table.

I’d suggest defining 4 as a constant at the top of the template and passing it through to each colspan call  (say number_of_columns).

Also I suggest extracting that full per-record block (as shown in the purple boc in the screenshot) into its own component/partial, along with its own scoped styling, rather than keeping the structure and styles spread inline here. Even though it is only used on this page, it is still the main repeated UI unit in the loop, so it feels like a proper component in its own right.

That would be more in line with how we already treat distinct UI blocks like the alert banner and belly band, and with the more component-led template structure used in FCL like in https://github.com/nationalarchives/ds-caselaw-public-ui/tree/main/ds_judgements_public_ui/templates/components

It should make the markup easier to follow, keep the styling better contained, and make future column/layout changes safer.

app/static/src/scss/includes/_browse-records.scss
```diff
@@ -0,0 +1,109 @@
...
+
+.records-page-wrapper {
+  .govuk-table {
+    width: 100%;
+    margin-bottom: 30px;
+    border-collapse: collapse;
+  }
+
+  .govuk-table__row {
+    border-bottom: none !important;
```
anthonyhashemi: Could we reduce `!important` usage for core layout rules here by moving more styles onto browse-records-specific selectors (for example row and cell classes for title, meta, path, and consignment rows), rather than overriding broad .govuk-table__* selectors?

That would align better with BEM component ownership and make the CSS cascade easier to reason about - as mentioned in https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity#the_!important_exception, `Using !important to override specificity is considered a bad practice and should be avoided for this purpose. Understanding and effectively using specificity and the cascade can remove any need for the !important flag.`

At the moment, broad overrides plus !important make later changes harder, because new rules often need more force to override previous ones.

If we scope these styles to the component, normal cascading (https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Introduction) and scoped selectors (https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity) can do more of the work, which should make future layout changes safer to maintain.

If any important declarations are still required after that, could we keep them limited to those cases and add a short reason comment for maintainability?

app/tests/test_browse_records.py
karwan-zaizi: visual regression test added

app/static/src/scss/includes/_browse-records.scss
anthonyhashemi: Interestingly, GDS does in fact use !important in supported `Override` styles: https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/README.md#overrides

However, the GDS documentation https://design-system.service.gov.uk/get-started/extending-and-modifying-components/ tells us to `Avoid overwriting GOV.UK Design System code` and has other reccomendations on how to extend GDS styles/components., including using their GDS Override styles where possible. Unfortunately they don't help here.

We can still align with the GDS extension pattern by using only using clearly app owned class names for custom behavior, and avoiding any broad Design System overrides unless absolutely necessary. You are already using app owned classes but we can just swtich the remaining important-based ovuk-table oiverrides in a similar way.

e.g.

Instead of 
```
.govuk-table__cell {
  padding-top: 0 !important;
}
```

```
  <td class="govuk-table__cell">...</td>
```

we can do something like

```
.browse-records__cell {
  padding-top: 0;
}
```

```
  <td class="govuk-table__cell browse-records__cell">...</td>
```

app/templates/main/browse-records.html
karwan-zaizi: Suggested changes are made

app/static/src/scss/includes/_browse-records.scss
karwan-zaizi: Suggested changes are made, would you mind have a look again

anthonyhashemi: Thanks for removing the !important overrides. I think my second comment above came in after you did that, and so partially still stands. In particular, move the remaining spacing rules off GOV.UK element selectors and onto app-owned classes so ownership is fully local. Full reasoning in https://github.com/nationalarchives/da-ayr-webapp/pull/2245#discussion_r3620540926

app/templates/main/browse-records.html
anthonyhashemi: amazing, looks good!

app/static/src/scss/includes/_browse-records.scss
donna-tna:
```
 .browse-record-row__row--meta .browse-records__meta-cell--date,
    .browse-record-row__row--meta .browse-records__meta-cell--series,
    .browse-record-row__row--meta .browse-records__meta-cell--status,
    .browse-record-row__row--meta .browse-records__meta-cell--opening-date,
    .browse-record-row__row--path .browse-records__path-cell {
      padding-top: 0;
      padding-bottom: 5px;
    }

    .browse-record-row__row--consignment .browse-records__consignment-cell {
      padding-top: 0;
      padding-bottom: 15px;
    }
```

```
    <td class="govuk-table__cell browse-records__meta-cell--date">
    <td class="govuk-table__cell govuk-table--invisible-on-mobile browse-records__meta-cell--series">
    <td class="govuk-table__cell browse-records__meta-cell--status"> 
    <td class="govuk-table__cell govuk-table--right-align browse-records__meta-cell--opening-date">


```
@anthonyhashemi is this already update by @karwan-zaizi here in the _browse-records-rows.scss and browse-records-rows.html ?

anthonyhashemi: Spoke with @donna-tna offline and explained that there were still some govuk element selectors still in `_browse-records.scss` which would be best avoided.

Since then an extra commit was made but some still remain due to concerns around the `_override.scss` file having link styling to the same elements and the override file's overrides are stulll in use in other tables so wouldn't be able to remove them.


I was also trying to understand how these all interact together and so checked out the branch and tested out the purely component-class approach I was suggesting locally and visual regression still pass.


I also sanity-checked selector impact by changing the browse-records cell/header padding from 20px to 203px, and that correctly triggered 2 visual diffs on browse records (desktop + mobile), which confirms these rules are active and test-covered.

Given that, I think we should incorporate the above changes into the branch.

I agree that the shared _overrides.scss needed by other pages can stay for now (although we should address this in due course); this comment is specifically about completing ownership within browse-records.


The 3 files I changed locally are as follows:


`app/static/src/scss/includes/_browse-records.scss`
```
@use "variables" as *;

.records-page-wrapper {
  .browse-records__table {
    width: 100%;
    margin-bottom: 30px;
    border-collapse: collapse;

    a {
      &:hover,
      &:visited:hover {
        color: $colour-link-hover;
        text-decoration-thickness: 3px;
      }
    }
  }

  .browse-records__table .browse-record-row__row {
    border-bottom: none;
  }

  .browse-records__table .browse-record-row__cell,
  .browse-records__table .browse-records__header-cell {
    padding: 10px 20px 10px 0;
    border-bottom: none;
  }

  .browse-records__table .browse-records__header-cell {
    border-bottom: 1px solid $colour-border-normal;
  }

  .browse-records__table .browse-record-row__cell--right,
  .browse-records__table .browse-records__header-cell--right {
    text-align: right;
  }

  .browse-records__table .govuk-table__body .browse-record-row__row:last-child .browse-record-row__cell {
    border-bottom: 1px solid $colour-border-normal;
  }
}

```


`app/templates/main/browse-records.html`
```
<div class="browse-layout records-page-wrapper">
    {% if num_records_found > 0 %}
        {% set number_of_columns = 4 %}
        {% set start_index = ((current_page - 1) * per_page) + 1 %}
        {% set end_index = [current_page * per_page, num_records_found] | min %}
        <div class="govuk-grid-column-two-thirds browse-grid--two-thirds browse-grid--mobile-table">
            <table class="govuk-table browse-grid__table browse-records__table"
                   id="tbl_result"
                   aria-label="Browse records table">
                <thead class="govuk-table__head">
                    <tr class="govuk-table__row">
                        <th scope="col" class="govuk-table__header browse-records__header-cell">Showing {{ start_index }}–{{ end_index }} of {{ num_records_found | format_number_with_commas }}</th>
                        <th scope="col"
                            class="govuk-table__header browse-records__header-cell govuk-table--invisible-on-mobile">Series</th>
                        <th scope="col" class="govuk-table__header browse-records__header-cell">Status</th>
                        <th scope="col" class="govuk-table__header browse-records__header-cell browse-records__header-cell--right govuk-table--right-align">Opening date</th>
                    </tr>
                </thead>
                <tbody class="govuk-table__body">
                    {% for record in results %}
                        {% include "components/browse-record-row.html" %}
                    {% endfor %}
                </tbody>
            </table>
            {% with view_name='main.browse_records', id=None %}
                {% include "pagination.html" %}
            {% endwith %}
        </div>
    {% endif %}
</div>

```


`app/templates/main/components/browse-record-row.html`
```
<tr class="govuk-table__row browse-record-row__row browse-record-row__row--title">
    <td colspan="{{ number_of_columns }}"
        class="govuk-table__cell browse-record-row__cell browse-records__title-cell">
        <a href="{{ url_for('main.record', record_id=record['file_id']) }}"
           class="browse-records__file-link">{{ record["file_name"] }}</a>
    </td>
</tr>
<tr class="govuk-table__row browse-record-row__row browse-record-row__row--meta">
    <td class="govuk-table__cell browse-record-row__cell browse-records__meta-cell--date">
        <p class="govuk-body-s govuk-!-margin-0">
            <time datetime="{{ record['date_of_record'] | format_date_iso }}">{{ record['date_of_record'] }}</time> – <span class="browse-records__last-modified-text">last modified</span>
        </p>
    </td>
    <td class="govuk-table__cell browse-record-row__cell govuk-table--invisible-on-mobile browse-records__meta-cell--series">
        <a href="{{ url_for('main.browse_series', _id=record['series_id']) }}">{{ record["series"] }}</a>
    </td>
    <td class="govuk-table__cell browse-record-row__cell browse-records__meta-cell--status">
        {% set closure_type = record["closure_type"] %}
        <strong class="{% if closure_type == 'Open' %}govuk-tag govuk-tag--green{% else %}govuk-tag govuk-tag--red{% endif %}">
            {{ closure_type if closure_type else "Unknown" }}
        </strong>
    </td>
    <td class="govuk-table__cell browse-record-row__cell browse-record-row__cell--right govuk-table--right-align browse-records__meta-cell--opening-date">
        {% if record["opening_date"] %}
            <time datetime="{{ record["opening_date"] | format_date_iso }}">{{ record["opening_date"] }}</time>
        {% endif %}
    </td>
</tr>
<tr class="govuk-table__row browse-record-row__row browse-record-row__row--path">
    <td colspan="{{ number_of_columns }}"
        class="govuk-table__cell browse-record-row__cell browse-records__path-cell">
        <p class="govuk-body-s govuk-!-margin-0">
            {% if record["file_path"] %}
                {% set path_parts = record["file_path"].split('/')[:-1] %}
                {% if record["file_path"] | length > 60 and path_parts | length > 1 %}
                    {{ path_parts[0] }} <span class="browse-records__path-condensed-separator" aria-hidden="true"></span> {{ path_parts[-1] }}
                {% else %}
                    {% for path_part in path_parts %}
                        {{ path_part }}
                        {% if not loop.last %}
                            <span class="browse-records__path-separator" aria-hidden="true"></span>
                        {% endif %}
                    {% endfor %}
                {% endif %}
            {% endif %}
        </p>
    </td>
</tr>
<tr class="govuk-table__row browse-record-row__row browse-record-row__row--consignment">
    <td colspan="{{ number_of_columns }}"
        class="govuk-table__cell browse-record-row__cell browse-records__consignment-cell">
        <p class="govuk-body-s govuk-!-margin-0 browse-records__consignment-text">
            consignment {{ record["consignment_reference"] }} on {{ record["date_of_record"] or "-" }}
        </p>
    </td>
</tr>

```


I hope that makes sense? More than happy to chat through any concerns and why this can work even with the _overrides.scss file still existing.

donna-tna: @anthonyhashemi thank you updated @colinbowen can you review this

colinbowen (approved):

PR #2246, opened 2026-07-08
lock file

colinbowen:
## Changes in this PR
- renovate bumped scikit-image in e2e_tests and failed to update poetry lock due to python version 3.10. This adds a check to make sure the lock file is updated as part of the pr

### After


anthonyhashemi (commented):
Approving as I think it makes sense to move away from supporting 3.10. What's this fixing?

anthonyhashemi (dismissed):

.github/workflows/pre_commit.yml
```diff
@@ -11,5 +11,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      with:
+        persist-credentials: false
     - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
     - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd
+
+  poetry-lock-check:
```
anthonyhashemi, line 19: this is definitely good guard rail for now but I'm suprised that renovate if leaving us in this state. Ideally we we'd fix this at the source in the renovate config

anthonyhashemi (approved):
Interesting that renovate isnt updating the poetry lock when it updates the toml but the guard rail is good for now until we have time to investigate

.github/workflows/pre_commit.yml
colinbowen: Created https://national-archives.atlassian.net/browse/AYR-2103 for the renovate change

PR #2276, opened 2026-07-16
Simplify AYR dev setup

karwan-zaizi:
## Changes in this PR
Simplify AYR dev setup

## JIRA ticket
[AYR-2021](https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?selectedIssue=AYR-2021)

### After

[AYR-2021]: https://national-archives.atlassian.net/browse/AYR-2021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker-compose.yml
```diff
@@ -0,0 +1,295 @@
+services:
```
anthonyhashemi, line 1: why do we need a new compose file? My understanding is that the ci one worked ok locally before this? I was thinking that it would need minimal tweaking to get hot reloading working? As a minimum we should build up compose files like

local: `docker compose -f docker-compose.yml -f docker-compose.local.yml up --build --wait`
ci: ``docker compose -f docker-compose.yml -f docker-compose.ci.yml up --build --wait`

e.g. 
https://github.com/nationalarchives/ds-caselaw-public-ui/blob/6328c92174166cd5d9cc86fd1ac13276511bd18b/docker-compose.prod-test.yml#L12

Maybe worth chatting through

Makefile
```diff
@@ -0,0 +1,48 @@
...
+$(OPENSEARCH_CA):
+	cd local_services/opensearch_certs && sh ./generate_opensearch_certs.sh
+
+$(KEYCLOAK_CERT):
+	sh ./local_services/generate-keycloak-certs.sh
+
+setup: $(ENV_FILE) $(WEBAPP_POSTGRES_CA) $(OPENSEARCH_CA) $(KEYCLOAK_CERT)
+	$(COMPOSE) build webapp
+
+start:
```
anthonyhashemi, line 25: This is great!

docker-compose.yml
karwan-zaizi: Let's have a chat, in the ticket suggests to move the docker-compose.yml to the root level

karwan-zaizi: we agreed to use the root compose file

Makefile
karwan-zaizi: thank you

anthonyhashemi (approved):
Amazing stuff!

PR #2287, opened 2026-07-20
Ayr 2102/pdfjs client side render

colinbowen:
## Changes in this PR
- Use PDF.js for rendering of records

## JIRA ticket
https://national-archives.atlassian.net/browse/AYR-2102

### After


app/static/init.uv.js
karwan-zaizi, line 1: @colinbowen Everything looks good overall. The only point to flag is that some files appear overly zoomed in. Is this expected behaviour?

karwan-zaizi: Screenshot 2026-07-22 at 13 56 54

colinbowen: I've revised the calculation for zooming to check what the width is.

karwan-zaizi: @colinbowen, it is still zoomed in, isn't it?

Screenshot 2026-07-23 at 15 44 30

colinbowen: This has now been set to display at full width of the document @karwan-zaizi

karwan-zaizi (approved):

app/static/init.uv.js
anthonyhashemi: @colinbowen @karwan-zaizi Reviewing this after merge, but I think we will need to iterate on this again,  since there is no search functionality anymore for any file type rendered.

PR #2300, opened 2026-07-23
AYR-2070 - hierarchical filters

karwan-zaizi:
## Changes in this PR
Browse - Hierarchical filters (TB / Series / Consignment)
Browse - (Standard user) Orientation and record count

## JIRA ticket
[AYR-2070](https://national-archives.atlassian.net/browse/AYR-2070)
[AYR-2072](https://national-archives.atlassian.net/browse/AYR-2072)
### for single Transferring body users)
for-single-Transferring-body-users

### for all access users
all

[AYR-2070]: https://national-archives.atlassian.net/browse/AYR-2070?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AYR-2072]: https://national-archives.atlassian.net/browse/AYR-2072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/main/db/queries.py
```diff
@@ -376,11 +380,81 @@ def build_browse_records_query(accessible_transferring_body_names=None):
...
+            sub_query.c.file_id,
+        )
+
+    return query
+
+
+def _build_browse_records_filters(query, sub_query, filters):
+    transferring_body = filters.get("transferring_body")
+    if transferring_body:
+        filter_value = str(f"%{transferring_body}%").lower()
```
colinbowen: This cast to a string can be removed as f-strings return strings

app/main/util/base_schemas.py
```diff
@@ -133,12 +133,14 @@ class BrowseFilterSchema(DateFilterSchema):
...
-        validate=validate.OneOf(["date_last_modified", "opening_date", ""]),
+        validate=validate.OneOf(
+            ["date_last_modified", "opening_date", "transferred", ""]
+        ),
     )
     record_status = fields.String(
         allow_none=True,
         load_default="all",
-        validate=validate.OneOf(["all", "open", "closed"]),
+        validate=validate.OneOf(["all", "open", "closed", "close"]),
```
colinbowen: The schema now accepts four values:
validate.OneOf(["all", "open", "closed", "close"])
and the query builder maps "close" → "closed". But the actual UI radio button that submits this field only ever sends value="closed". Is this correct?

app/main/util/browse_records_utils.py
```diff
@@ -0,0 +1,222 @@
...
+    accessible_body_names,
+):
+    if selected_transferring_body:
+        return selected_transferring_body
+    if accessible_body_names and len(accessible_body_names) == 1:
+        return accessible_body_names[0]
+    return selected_transferring_body
+
+
+def get_options_rows(accessible_body_names):
```
colinbowen: This query could get intensive as every request will load this, not just when a user wants to look at the options. 

get_options_rows() joins Body → Series → Consignment → File and pulls back one row per distinct (body, series, consignment) selection on every page load, regardless of whether any filter is even being used. 
- The transferring-body list only needs distinct body names

The series/consignment autofill only needs to run when a series or consignment filter is actually typed in without a body filter, and could be a targeted WHERE filtered query

app/main/util/browse_records_utils.py
```diff
@@ -0,0 +1,222 @@
...
+    options_rows,
+):
+    if selected_series and not selected_transferring_body:
+        selected_series_lower = selected_series.lower()
+        matching_transferring_bodies = sorted(
+            {
+                row.transferring_body
+                for row in options_rows
+                if row.transferring_body
+                and (row.series or "").lower() == selected_series_lower
```
colinbowen:
```suggestion
                and selected_series_lower in (row.series or "").lower()
```

The transferring-body/series autofill logic previously required an exact match against the series name (== selected_series_lower), while the actual query filter that narrows the results uses a substring match (LIKE %value%). That meant typing a partial series name (e.g. "HO 4") filtered the record list correctly but left the Transferring Body box empty, since no series name matched it exactly.

app/main/util/browse_records_utils.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if selected_consignment and (
+        not selected_transferring_body or not selected_series
+    ):
+        selected_consignment_lower = selected_consignment.lower()
+        matching_rows = [
+            row
+            for row in options_rows
+            if row.consignment_reference
+            and row.consignment_reference.lower() == selected_consignment_lower
```
colinbowen:
```suggestion
            and selected_consignment_lower in row.consignment_reference.lower()
```
same here

app/main/db/queries.py
karwan-zaizi: Change for this in place

app/main/util/base_schemas.py
karwan-zaizi: reverted this to accept only three `validate.OneOf(["all", "open", "closed"])`

app/main/util/browse_records_utils.py
karwan-zaizi: Made the changes for this

app/main/util/browse_records_utils.py
```diff
@@ -0,0 +1,262 @@
...
+def get_accessible_body_names(ayr_user: AYRUser):
+    if ayr_user.is_standard_user:
+        body = ayr_user.transferring_body
+        return [body.Name] if body else []
+    if ayr_user.is_all_access_user:
+        return None
+    abort(403)
+
+
+def get_selected_filters(validated_data):
```
colinbowen, line 17: It looks like the transferring body / series filter values are being calculated twice for the new page, once in the shared build_filters() helper, then again with extra logic for autofill in browse_records_utils.py, with the second one silently overwriting the first. 

It may be better to change build_filters() so that it has a flag e.g. include_hierarchical_filters which defaults to True, so the other browse pages that use it aren't affected. The records page could set this to False, so browse_records_utils.py is now the only function doing this calculation.

colinbowen (approved):

PR #2326, opened 2026-07-30
AYR-2068 - sort component

karwan-zaizi:
## Changes in this PR
Add sort component

## JIRA ticket
[AYR-2068](https://national-archives.atlassian.net/browse/AYR-2068)

added this also [AYR-2110] https://national-archives.atlassian.net/browse/AYR-2110

## Screenshots of UI changes
JS enabled
Screenshot 2026-07-30 at 11 34 35

JS disabled
Screenshot 2026-07-30 at 11 35 16

### AYR-2110 -All access users

all-user

### Standard users  AYR-2110

standard

[AYR-2068]: https://national-archives.atlassian.net/browse/AYR-2068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AYR-2110]: https://national-archives.atlassian.net/browse/AYR-2110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/tests/test_browse_records.py
```diff
@@ -588,23 +596,62 @@ def test_browse_records_filters_and_sort_persist_in_per_page_controls(
...
-            hidden_sort = form.find(
-                "input",
-                {
-                    "type": "hidden",
-                    "name": "sort",
-                    "value": "file_name-desc",
-                },
+            sort_select = form.find("select", {"name": "sort"})
+            selected_sort_option = (
+                sort_select.find("option", selected=True)
```
anthonyhashemi: I highly reccomend avoiding conditional logic in tests so that they are easier to understand and so that they are testing one scenario - conditional logic suggests that there is at least 2 scenarios under test here, which makes debugging more difficult.

app/static/src/scss/includes/_sort-records.scss
```diff
@@ -69,63 +75,81 @@
...
-    width: 100%;
+    justify-content: space-between;
     margin-bottom: 16px;
 
-    .sort-list-records-no-js {
+    align-items: flex-end;
+    gap: 0.5rem;
+    margin-right: 35%;
+
+    .sort-list-records-no-js--sort-control {
```
anthonyhashemi, line 127: **-- used for BEM elements rather than __ (three places)**

```
.sort-list-records-no-js--sort-control { … }
.sort-list-records-no-js--sort-action  { … }
.sort-list-records-no-js--per-page     { … }
```

These are structural child parts of the .sort-list-records-no-js block, not state variants of it. [BEM](https://getbem.com/introduction/) uses __ for elements and -- for modifiers:

```
// wrong — reads as "the no-js form in its sort-control state"
.sort-list-records-no-js--sort-control { … }

// correct — reads as "the sort-control part of the no-js form"
.sort-list-records-no-js__sort-control { … }
.sort-list-records-no-js__sort-action  { … }
.sort-list-records-no-js__per-page     { … }
```

The HTML class attributes will need updating to match.

We should update our stylelint config to catch this

app/static/src/scss/includes/_sort-records.scss
```diff
@@ -69,63 +75,81 @@
...
       display: flex;
-      align-items: center;
-      margin-right: 35%;
+      flex-direction: column;
+      align-items: flex-start;
 
-      .browse-records__label--right {
-        font-size: 1.1rem;
+      .govuk-label {
```
anthonyhashemi: **.govuk-label overridden via descendant selector**

The [GDS guidance](https://design-system.service.gov.uk/get-started/extending-and-modifying-components/) says not to target `govuk-` classes directly. It will silently break on `govuk-frontend` updates. Options in order of preference:

1. Use the GDS spacing override class in the HTML: `class="govuk-label govuk-!-margin-bottom-1"`
2. If a custom class is needed: `class="govuk-label ayr-label--tight"` with `.ayr-label--tight { margin-bottom: 0.25rem; }`
  ayr- = our service's own prefix (not govuk-)
label = the component we're modifying
--tight = the modifier describing what it does (reduces margin)

app/templates/main/sort-records.html
```diff
@@ -3,65 +3,82 @@
...
       action="{{ url_for('main.browse_records', _anchor='records') }}"
       method="get"
       class="sort-list-records-form">
-    {% for key, value in preserved_params.items() %}
-        <input type="hidden" name="{{ key }}" value="{{ value }}">
-    {% endfor %}
+    {{ render_preserved_params(preserved_params) }}
     <div class="sort-list-records-left-wrapper">
         <div class="sort-list-records-js">
+            <label class="govuk-label govuk-visually-hidden" for="sort-records-js">Sort records</label>
```
anthonyhashemi, line 42: **sort-records.html: govuk-select__sort-container-select uses the reserved govuk- prefix**

`<select class="govuk-select govuk-select__sort-container-select js-records-sort-select" …>`

`govuk-select__sort-container-select` is not a govuk-frontend class , has no CSS rule anywhere in this codebase, and is not referenced in any JS. The actual JS hook is `js-records-sort-select`. The class does nothing.

It already exists in `search-transferring-body.html` and `sorting-list.html` as the same pre-existing dead code, worth a separate clean-up ticket for those. But this PR shouldn't add two more instances of it.

Either remove it entirely from both selects here, or if a custom styling hook is intended for future use, rename it with an `ayr-` prefix: `ayr-select--sort`.

app/templates/main/sort-records.html
```diff
@@ -3,65 +3,82 @@
...
+    "Modified date – newest": "date_of_record-desc",
+    "Modified date – oldest": "date_of_record-asc",
+    "File name – A to Z": "file_name-asc",
+    "File name – Z to A": "file_name-desc",
+    "Record opening date – sooner": "opening_date-asc",
+    "Record opening date – later": "opening_date-desc",
+    "Series – A to Z": "series-asc",
+    "Series – Z to A": "series-desc"
+} %}
+{% macro render_preserved_params(params) %}
```
anthonyhashemi, line 18: Nice macro refactors - it's now easier to comprehend the logic!


I noticed that `search-transferring-body.html` still has the same inline pattern which was extracted here into `render_preserved_params`:


```
{% for key, value in preserved_params.items() %}
    <input type="hidden" name="{{ key }}" value="{{ value }}">
{% endfor %}
```

Since the same logic now exists in two templates, `render_preserved_params` is a good candidate to move to a shared macros file (e.g. `templates/macros/forms.html`) and import in both. Not necessarily blocking this PR, but if not done here please make a follow-up ticket to avoid the drift widening.

app/templates/main/sort-records.html
```diff
@@ -3,65 +3,82 @@
 {% set _ = preserved_params.pop('per_page', None) %}
```
anthonyhashemi, line 3: `sort-records.html` is a candidate for moving into the existing `components/` directory

The codebase already has the right structure for this:

`sort-records.html` behaves like a component. It's included by `browse-records.html` and has its own SCSS partial, but it lives loose in `main/` rather than in `components/`. Not a blocker for this PR, but worth a follow-up to move it alongside the other components.

The longer-term direction, once in components/, would be to convert the {% include %} to a proper {% macro %} with an explicit parameter signature, making the component's data dependencies visible at the call site rather than implicit from view context. Looks like:


```
{# components/sort-records.html #}
{% macro sort_records(preserved_params, per_page_options, sort_options, selected_sort) %}
  …
{% endmacro %}

{# browse-records.html #}
{% from 'main/components/sort-records.html' import sort_records %}
{{ sort_records(preserved_params, dict, records_sort_options, selected_sort) }}
```

Not blocking but if not done here can you please make a followup ticket for it.

app/templates/main/sort-records.html
```diff
@@ -3,65 +3,82 @@
...
             </div>
-        </div>
-    </form>
+        </form>
+        <form action="{{ url_for('main.browse_records', _anchor='records') }}"
+              method="get"
+              class="sort-list-records-no-js--per-page">
+            {{ render_preserved_params(preserved_params) }}
+            <input type="hidden" name="sort" value="{{ selected_sort }}">
+            <label class="govuk-label browse-records__label--left" for="sort-js">Records per page</label>
```
anthonyhashemi: The no-JS form's `id` and `for` don't match, the label is not associated with the select.

The id="sort-js" in the no-JS form's label looks like a copy-paste from the JS form that was never updated when the id was changed to sort-no-js. Fix is a one-character change: `for="sort-no-js"`.

This suggests that this has not been tested - we should review what accessbility testing we have in the repo and whether any of the checks are running in CI. For example, we have some references to pa11y in the codebase. Could you create a ticket for us to investigate?

For completeness, the JS form (lines 46–47) is fine.

anthonyhashemi (changes requested):

app/tests/test_browse_records.py
karwan-zaizi: removed unnecessary test

donna-tna: @karwan-zaizi updated this @anthonyhashemi

app/static/src/scss/includes/_sort-records.scss
donna-tna: updated this file @anthonyhashemi

app/static/src/scss/includes/_sort-records.scss
donna-tna: updated this file @anthonyhashemi

app/templates/main/sort-records.html
donna-tna: updated this file @anthonyhashemi

app/templates/main/sort-records.html
donna-tna: okay new ticket will be made @anthonyhashemi

app/templates/main/sort-records.html
donna-tna: okay new ticket will be made @anthonyhashemi

app/templates/main/sort-records.html
donna-tna: okay new ticket will be made to investigate @anthonyhashemi

app/static/src/scss/includes/_sort-records.scss
```diff
@@ -25,12 +25,52 @@
...
+
+        &:focus {
+          outline: none;
+          box-shadow: none;
+        }
+
+        &:focus-visible {
+          outline: 3px solid #fd0;
+          outline-offset: 0;
+          box-shadow: inset 0 0 0 2px #0b0c0c;
```
anthonyhashemi: @donna-tna we should avoid hardcoding colours in component level scss, we have the colours.scss file where we define any colours if they don't come from GDS

app/templates/main/components/browse-record-row.html
```diff
@@ -26,6 +26,16 @@
         {% endif %}
     </td>
 </tr>
+{% if not is_standard_user %}
+    <tr class="govuk-table__row browse-record-row__row browse-record-row__row--transferring-body">
+        <td colspan="{{ number_of_columns }}"
+            class="govuk-table__cell browse-record-row__cell browse-records__transferring-body-cell">
+            <p class="govuk-body-s govuk-!-margin-0">
+                <a href="{{ url_for('main.browse_transferring_body', _id=record['transferring_body_id']) }}">{{ record["transferring_body"] }}</a>
```
anthonyhashemi: @donna-tna Did we intend to incorporate work for the transferring body line in the table in this Sort component PR? 

Either way, as clarified offline, the `href` here is not quite right. We plan to remove the other browse pages. Instead, it should link to a pre-filtered version of the browse-records page itself.

app/templates/main/sort-records.html
anthonyhashemi: Cheers, the accessbility testing investigation can be done in that followup ticket

However, the label issue described above is still present and should be fixed in this PR rather than introducing a bug we are aware of at PR time.

app/static/src/scss/includes/_sort-records.scss
anthonyhashemi: The issue described is still present in the updated file

for example on line 127:
`.sort-list-records-no-js--sort-control`
should be
`sort-list-records-no-js__sort-control`

app/static/src/scss/includes/_sort-records.scss
donna-tna: ahh yes thanks updated @anthonyhashemi

app/templates/main/sort-records.html
```diff
@@ -74,7 +74,7 @@
               class="sort-list-records-no-js--per-page">
             {{ render_preserved_params(preserved_params) }}
             <input type="hidden" name="sort" value="{{ selected_sort }}">
-            <label class="govuk-label browse-records__label--left" for="sort-js">Records per page</label>
+            <label class="browse-records__label--right" for="sort-no-js">Records per page</label>
```
donna-tna, line 77: @anthonyhashemi this was updated here in the last commit [13a05a8] on line 77

app/templates/main/components/browse-record-row.html
donna-tna: @anthonyhashemi thanks this I will update in the next commit after talking with @karwan-zaizi yesterday

app/templates/main/sort-records.html
donna-tna: replied below

app/static/src/scss/includes/_sort-records.scss
donna-tna: @anthonyhashemi updated this and also the test file which has `per-page`

anthonyhashemi (approved):

PR #2345, opened 2026-08-10
AYR-2126 - fix standard user series autofill issue

karwan-zaizi:
## Changes in this PR
When standard user filters by consignment, the series filter should autofill

## JIRA ticket
[AYR-2126](https://national-archives.atlassian.net/browse/AYR-2126)

### After

[AYR-2126]: https://national-archives.atlassian.net/browse/AYR-2126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/main/util/browse_records_utils.py
+++ b/app/main/util/browse_records_utils.py
@@ -39,8 +39,6 @@ def prefill_transferring_body_for_single_scope(
     selected_transferring_body,
     accessible_body_names,
 ):
-    if selected_transferring_body:
-        return selected_transferring_body
     if accessible_body_names and len(accessible_body_names) == 1:
         return accessible_body_names[0]
     return selected_transferring_body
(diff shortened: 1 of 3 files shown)


app/main/util/browse_records_utils.py
```diff
@@ -200,9 +198,14 @@ def build_browse_records_filter_data(
             accessible_body_names
         )
 
-    if not selected_transferring_body and (
-        selected_series or selected_consignment
-    ):
+    should_autofill_from_series = (
+        selected_series and not selected_transferring_body
+    )
+    should_autofill_from_consignment = (
```
colinbowen, line 204:
```suggestion
    should_autofill_from_consignment = (
    selected_consignment and not selected_series and not selected_transferring_body
)
```
This needs to be tested

colinbowen: I should have been clearer. I believe there should be a check to make sure the transferring body is not selected on the second part of this function. This new addition needs to be tested.

colinbowen: What happens if the transferring body is already selected? Will an autofill from consignment override it?

colinbowen (approved):

PR #2351, opened 2026-08-12
Refactor browse records query to improve performance with a two-stage…

anthonyhashemi:
… fetch approach

### After


app/main/db/queries.py
```diff
@@ -589,10 +592,13 @@ def _apply_base_query_sort(query, sorting_orders):
     return query
 
 
+_DEFAULT_SORTING_ORDERS = object()
```
anthonyhashemi: This feels unnecessarily complex, is there a reason we need this now compared to before where we just had None as the default. I assume it is to distinguish "argument omitted" from "argument explicitly set to None"?

Also, with this change, the new conditional block added `    if sorting_orders is None:` in `_apply_base_query_sort` is never hit.

app/main/routes.py
```diff
@@ -32,8 +32,9 @@
 from app.main.db.queries import (
     build_browse_consignment_query,
     build_browse_query,
-    build_browse_records_query,
```
colinbowen, line 35: Can build_browse_records_query be deleted? It's no longer being used

colinbowen: And _build_browse_records_filters

karwan-zaizi: Removed `build_browse_records_query`

anthonyhashemi: Can `_build_browse_records_filters` also be removed as Colin suggested?

karwan-zaizi: This is already removed

app/main/db/queries.py
karwan-zaizi: It's was one of Copilot suggestions I was looking at and suggested this.
I have reverted the changes

colinbowen (approved):

PR #2355, opened 2026-08-13
Improve stylelint config to include BEM  and add ESLint

donna-tna:
## Changes in this PR
Improve stylelint config to include BEM  and add ESLint
## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?filter=&groupBy=none&selectedIssue=AYR-2127
### After


app/templates/main/browse-all.html
```diff
@@ -42,7 +42,7 @@
                         <th scope="col"
                             class="govuk-table__header govuk-table--invisible-on-mobile">Series reference</th>
                         <th scope="col"
-                            class="govuk-table__header govuk-table--right-align govuk-table--on-mobile--no-padding-right">
+                            class="govuk-table__header govuk-table--right-align govuk-table__no-padding-right">
```
anthonyhashemi: This PR should not result in any end user visual changes.

Image

All these changes to the classes in the older browse pages have caused the visual snapshots for them to change - I assume because their styling is now broken since there is no corresponding style class in scss.

In particular you can see in the screenshot I posted above

I see you have ignored the scss with stylelint so styelint has not changed the browse-all.scss but one of the other linters must have changed these files I'm guessing?

donna-tna: Ah yes these are html pages I  started to change to due stylint BEM errors being flagged  and should have reverted back before pushing the branch as we will be deleting these pages shortly. We can refractor after the new browse page is in rather than refactoring pages due for deletion.  I can revert these changes and re push @anthonyhashemi

anthonyhashemi: Ah I see, that makes more sense, I was struggling to see which linter would have changed them!

.stylelintrc
```diff
@@ -1,5 +1,6 @@
 {
     "extends": [
+        "@nationalarchives/stylelint-config",
```
anthonyhashemi, line 3: now that we are extending `nationalarchives/stylelint-config` I imagine we can get rid of a lot of the rules that we had explicitly set. 

As we discussed yesterday, further customisation of our configs is fine but only if we have a good reason for it.

All the plugins like `scss/dollar-variable-empty-line-before` etc I assume would be covered by the tna one, as would the new BEM `selector-class-pattern` you added?

You can see the full tna config here: https://github.com/nationalarchives/stylelint-config/blob/main/index.mjs

eslint.config.mjs
```diff
@@ -0,0 +1,40 @@
...
+      "coverage/",
+      "storybook-static/",
+      "storybook/stories/",
+      "app/static/init.uv.test.js",
+      "configs/pa11y_ci_precommit.js",
+      "app/static/init.uv.js",
+    ],
+  },
+  ...tnaEslintConfig,
+  {
```
anthonyhashemi, line 8: Similar comment to the one I left for stylelintconfig. 

Since we inherit the TNA `tnaEslintConfig`, it would be good to discuss the need for any of these extra custom explicit rules that aren't covered by that config.

If you look at the definition of `tnaEslintConfig`: https://github.com/nationalarchives/tna-frontend/blob/main/eslint.config.mjs you will see that many of these are already covered.

donna-tna: @anthonyhashemi updated eslint removed the rules TNA share already

.stylelintrc
donna-tna: @anthonyhashemi I assumed another refactor of codebase ticket will be opened next sprint not done in this ticket once the browse pages are deleted and new browse page is in place. As some pages will be deleted along with their errors so just for now kept the rules to set them to null. Then after refactor will remove rules set to null and use TNA rules

eslint.config.mjs
anthonyhashemi: cheers

eslint.config.mjs
```diff
@@ -0,0 +1,38 @@
...
+    files: [".storybook/**/*.js", "accessibility_tests/**/*.js", "*.js"],
+    languageOptions: {
+      globals: {
+        module: "readonly",
+        require: "readonly",
+        process: "readonly",
+        __dirname: "readonly",
+      },
+    },
+    rules: {
```
anthonyhashemi: Thanks for removing the majority of the custom rules. Could you explain why we need these 2 remaining ones? Or can we also do without them?

.stylelintrc
anthonyhashemi: I think there might be a small misunderstanding here, so I'll clarify what I meant.

The rules I'm talking about removing aren't related to the browse pages deletion, they're about simplifying our config right now because we're in the process of extending from the TNA config.

Removing redundant rules won't affect validation at all since these rules are already set in the TNA config.

The redundant rules we can remove immediately are:

`declaration-empty-line-before`
`property-no-vendor-prefix`
`value-keyword-case`
`scss/double-slash-comment-empty-line-before`

Since the TNA config already sets these to null, we're just maintaining duplicate configuration.

The bem pattern rule you are adding in this PR may also not be needed since the tna config has 
```
extends: ["stylelint-config-standard-scss"],
plugins: ["stylelint-selector-bem-pattern", "stylelint-order"],
...
"plugin/selector-bem-pattern": {
  preset: "bem",
},
```

so worth trying without the extra custom pattern you put in and seeing if we get the same behaviour with what it has in built, because if that works fine then we don't need to maintain our own code for it.

.stylelintrc
```diff
@@ -1,36 +1,31 @@
...
-                ]
+                "message": "Class selector must follow BEM naming convention (block__element--modifier)",
+                "resolveNestedSelectors": true
             }
         ],
-        "scss/operator-no-newline-after": null,
+        "plugin/selector-bem-pattern": {
+            "preset": "bem"
+        },
         "declaration-property-value-keyword-no-deprecated": null,
```
anthonyhashemi: Further from the other comment, there are indeed some rules which were present before but are not defined in the tna config:

```
- declaration-property-value-keyword-no-deprecated: null
- block-no-redundant-nested-style-rules: null
- media-type-no-deprecated: null
- nesting-selector-no-missing-scoping-root: null
- no-invalid-position-declaration: null
- property-no-deprecated: null
```

These are all disabled which suggests our codebase currently violates them (or did at some point).

This part could feasibly be a separate ticket to avoid scope creep, though if it turns out removing these 6 rules and running stylelint --fix is truly just a quick, clean fix with no side effects, it might save time to just do it now while we're already in the config (either in this PR or just a followup PR against the same ticket). But that's probably a call to make based on how it actually plays out when you test that.

eslint.config.mjs
donna-tna: these should have been removed as the files in ignore were added `"lighthouserc.js",
      "app/static/init.uv*.js",`

.stylelintrc
donna-tna: @anthonyhashemi  I have been reading up and asked AI why we have it duplicated as I have not configured stylelint before so very confused why errors occur if taken out and we are using TNA .  It looks redundant, but please can you also test it because if we delete those it throws errors.
If TNA's config was carrying over cleanly, we wouldn't get errors when we delete those lines. The fact that the errors instantly come back proves the config inheritance is dropping them, and we have to declare them locally to keep them turned off.

anthonyhashemi: Okay that's interesting that it causes a change when removing duplicate ruless, I will check out the branch and see what happens on my machine 👌

eslint.config.mjs
```diff
@@ -0,0 +1,34 @@
+import tnaEslintConfig from "@nationalarchives/eslint-config";
+import { defineConfig } from "eslint/config";
+
+export default defineConfig([
+  {
+    ignores: [
```
anthonyhashemi: Why have we chosen to ignore all of these files? Does `eslint --fix` result in visual changes, I wouldn't imagine it should? Or does it not fix all the issues?

eslint.config.mjs
```diff
@@ -0,0 +1,34 @@
...
+    languageOptions: {
+      globals: {
+        module: "readonly",
+        require: "readonly",
+        process: "readonly",
+        __dirname: "readonly",
+      },
+    },
+  },
+]);
```
anthonyhashemi: Also why do we need to override the language options for these files?

.stylelintignore
```diff
@@ -0,0 +1,11 @@
+**/_browse-transferring-body.scss
+**/_browse-consignment.scss
+**/_search.scss
```
anthonyhashemi: Why are we ignoring search, search-transgerring-body, record, uv, breadcrumbs and table's .scss?

Does `stylelint --fix` not fix issues with it automatically like it has done for the other files?

Or does it result in visual change on those pages, again I wouldn't have thought it should?

I will test it out for breadcrumbs, for example, and see...

eslint.config.mjs
donna-tna: @anthonyhashemi I updated language options as TNA only has  (...globals.browser and parserOptions) in their config. However, for our Node configs and Jest test files, I added  language options so they pick up Node/Jest environments without erroring.  Based on this https://eslint.org/docs/latest/use/configure/migration-guide#configure-language-options as

.stylelintignore
donna-tna: @anthonyhashemi  removed search, search-transgerring-body, record, uv, breadcrumbs and table's .scss and applied stylelint --fix for 111 errors

anthonyhashemi (approved):

colinbowen (approved):

PR #2381, opened 2026-08-21
test accessibility noscript theory

donna-tna:
## Changes in this PR
test accessibility noscript theory
## JIRA ticket
https://national-archives.atlassian.net/jira/software/projects/AYR/boards/66?filter=&groupBy=none&selectedIssue=AYR-2138
### After


lighthouserc.js
```diff
@@ -184,6 +190,12 @@ module.exports = {
             minScore: 0.9,
           },
         ],
+        label: [
```
anthonyhashemi: So this is saying, mark it as  error if ANY category wasn't 100% fine. 

If the purpose of this PR is around flagging accessibility errors, that would make the block above

```
        "categories:accessibility": [
          "error",
          {
            minScore: 0.9,
          },
        ],
```


redundant.

Or was there a particular reason for this?

custom_hooks/pa11y_ci_precommit.sh
```diff
@@ -1,10 +1,15 @@
 #!/bin/bash
-set +e
+set -e
```
anthonyhashemi: Is this script even used anywhere? I'd assume this takes a long time to run and if so I doubt we'd want it as a pre-commit step? Thoughts?

app/tests/test_routes.py
```diff
@@ -128,6 +128,38 @@ def test_route_accessibility(self, client: FlaskClient):
         response = client.get("/accessibility")
         assert response.status_code == 200
 
+    def test_route_accessibility_has_required_sections_and_links(
```
anthonyhashemi: Out of interest, what prompted this test addition? Did code coverage show that these paths were not covered?

app/tests/test_browse_records.py
```diff
@@ -720,39 +720,54 @@ def test_browse_records_filters_and_sort_persist_in_per_page_controls(
         assert noscript_per_page_select is not None
         assert noscript_per_page_sort_select is None
 
-    def test_browse_records_no_results_for_nonsensical_filter(
+    def test_browse_records_labels_target_existing_controls(
```
anthonyhashemi: I do understand the underlying accessibility concern, as this stems from comments on the initial sort work where there was a mismatch between label and for attribute but I think that as long as we have proper accessbility testing with pally or otherwise which would check for accessibility violations, then the need for us to check the consistency of specific for and label would be redundant.

Worth verifying that the new accessibility ci run would flag an issue if a label/control this association were broken here - I'd hope it would, as otherwise we'd need to test for this sort of thing across all labels in our rendered html.


Also, this test at the moment is quite brittle due to its coupling with scss class names, and overcomplicated since it loops over different selectors and thus requires conditional logic to distinguish the different scenarios.

If we were to keep a form of this test, I'd suggest splitting it out into separate tests for readability and debugability. Hopefully, we don't need it though and hopefully the accessibility testing itself will cover the intention of this test.

accessibility_tests/lhci-auth-autorun.js
```diff
@@ -1,53 +1,20 @@
+const fs = require("node:fs");
```
anthonyhashemi: Can you explain why the authentication for lighthouse runs had to change?

.github/workflows/ci-tests.yml
```diff
@@ -47,6 +44,7 @@ jobs:
           accessibility:
             - 'app/**'
             - 'configs/**'
+            - 'accessibility_tests/**'
```
anthonyhashemi: what happened before we added this?

custom_hooks/pa11y_ci_precommit.sh
karwan-zaizi: You’re right. It is currently not wired in .pre-commit-config.yaml, so it does not run by default; it was intended as an optional local check.

I was testing some other options to run it locally and forgot to revert the changes, same as the label change

lighthouserc.js
karwan-zaizi: I added label because I saw label issues not always reflected by category score, now the other changes are in place this is not necessary, I have reverted this

.github/workflows/ci-tests.yml
karwan-zaizi: The tests still would run but this is just to double check if any files added in the future or changed would pick that as well

accessibility_tests/lhci-auth-autorun.js
karwan-zaizi: I added this, because the updated Lighthouse runner now reads the generated `.lighthouseci` JSON reports after `lhci autorun` completes, so we can validate that protected AYR routes were actually scanned and not redirected to Keycloak

app/tests/test_browse_records.py
karwan-zaizi: I removed that test because it is fragile. Accessibility checks should catch this already. I kept only the main fix in this PR: Lighthouse login/session and redirect safety check.

This extra tests were from earlier version before I focused this PR on accessibility pipeline fix.

.github/workflows/ci-tests.yml
anthonyhashemi: okay, the list of path filters is pretty big, maybe we need to rethink at some point

accessibility_tests/lhci-auth-autorun.js
anthonyhashemi: Just for clarity, have you not also changed the way we authenticate, moving from auth tokens to sessioncookies?

What I see happening in this function now is

1. authenticating into the app in browser with pupeeteer,
2. extracting session cookies from browser,
3. running lighthouse with the extracted session cookies
4. validating Lighthouse actually scanned protected routes.

compared to before where it was:

1. directly fetch access token from keycloak api with fetch,
2. run lighthouse with the access token as bearer token

karwan-zaizi: Yes, exactly. I changed from bearer token to session cookie on purpose. AYR protected routes are gated by app session after sign-in, so sending only a Keycloak access token was not enough and could still redirect Lighthouse to auth pages. The new flow logs in with Puppeteer, takes the real session cookies, runs Lighthouse with that cookie context, and then checks requested vs final URLs so we know protected pages were truly scanned.

app/tests/test_routes.py
karwan-zaizi: This came from an earlier iteration while I was adding extra safety checks around accessibility. As the PR scope was narrowed to Lighthouse auth/session and CI accessibility reliability, this didn't add any necessary checks

